Issue #368, opened 2012-08-23
Deal with duplicate blocks instead of giving an error


rjoomen:
Someone on StackOverflow gave an example of [very creative use of a child document](http://stackoverflow.com/questions/12095113/r-knitr-possible-to-programmatically-modify-chunk-labels/12097884). This did unfortunately not work, as it repeatedly processed the child, raising duplicate label errors. I made a fix that adds the chunk number to a duplicate label and warns the user, e.g. _Duplicate label mpg-histogram renamed to mpg-histogram-7_.

All checks still run with this change.

Do you think this is useful, or do you want to force users to use unique chunk names? An additional possibility could be to create a package option `allow.duplicates` defaulting to FALSE that controls this.

PR #369, opened 2012-08-23
This changes how knitr deals with duplicate chunk names, see issue #368

rjoomen:


gosianow:
The solution with knit_expand and {{}} looks really nice. However, as mentioned here https://gist.github.com/rmoff/a043676a2f084b81a434 it makes the code in child chunk not runnable. Is there any chance that you would provide a solution that can handle that?

PR #866, opened 2014-10-20
Add the RcppOctave engine

casallas:
- This engine allows to call octave/matlab code within R using RcppOctave


yihui:
Closing this, as discussed above — the RcppOctave engine is a better fit for the RcppOctave package itself, where an Octave/Matlab expert can maintain it. Thanks for the contribution and the interesting idea!

PR #928, opened 2015-01-07
chunks in markdown gain the inline option.

ThierryO:
inline = TRUE generates markdown code for inline plot. inline = FALSE or omitting inline generates the default behaviour.

Default behaviour after conversion to LaTeX

\begin{figure}
  \includegraphic{theplot}
\end{figure}

inline = TRUE after coversion to LaTeX

\includegraphic{theplot}

### DIFF
--- a/R/hooks-md.R
+++ b/R/hooks-md.R
@@ -30,7 +30,13 @@ hook_plot_md_base = function(x, options) {
 
   if (is.null(w <- options$out.width) & is.null(h <- options$out.height) &
     is.null(s <- options$out.extra) & options$fig.align == 'default') {
-    return(sprintf('![%s](%s%s) ', cap, base, .upload.url(x)))
+    markdown <- sprintf('![%s](%s%s) ', cap, base, .upload.url(x))
+    if(is.null(inl <- options$inline)){
+      markdown <- paste0(markdown, " ")
+    } else {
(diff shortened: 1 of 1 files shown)


yihui:
Thanks! You can just use any of the chunk options `fig.align`, `out.width`, `out.height`, and `out.extra` to trigger the LaTeX mode for graphics, e.g.

    ---
    graphics: yes
    output:
      pdf_document:
        fig_caption: yes
        keep_tex: yes
    ---
    
    ```{r out.extra=''}
    plot(1:10)
    ```

ThierryO:
The `inline` argument that I propose actually works on the markdown code.

`inline = FALSE` produces the default `caption` markdown code. When converted from markdown to LaTeX it becomes 
```
\begin{figure}
    \includegraphics{image.file}
    \caption{caption}
\end{figure}
```

`inline = TRUE` produces an inline image `caption\ ` Note the trailing '\ '. Converting this from markdown to LaTeX gives
```
\includegraphics{image.file}
```

I can't get the same behaviour `out.ext = ''`

yihui:
I did not say `out.ext`. I said `out.extra`. Did you try my example above?

ThierryO:
Oops. My mistake. `out.extra = ''` works indeed. So my request reduces to updating http://yihui.name/knitr/options/ ;-)

PR #937, opened 2015-01-16
Automatic test coverage reports with coveralls.

jimhester:
You will have to enable this repo on coveralls at https://coveralls.io/repos/new before merging this pull request for the report to work the first time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
   - sudo apt-get install -qq pandoc pandoc-citeproc texlive-latex-recommended > /dev/null
   - "wget -q -O - http://yihui.name/xran/r-config | bash"
   - Rscript -e 'devtools::install_deps(dep = TRUE)'
+  - Rscript -e 'devtools::install_github("jimhester/covr")'
 
 # run tests
 script:
@@ -37,3 +38,6 @@ script:
 after_success:
   - export R_PKG="$(basename $TRAVIS_REPO_SLUG)"
   - "[ $TARGET = integration ] && (wget -q -O - http://yihui.name/xran/r-xran | bash)"
+
+  # run test coverage only for the travis build
+  - "[ $TARGET = travis ] && Rscript -e 'library(covr);coveralls()'"
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # knitr
 
 [![Build Status](https://travis-ci.org/yihui/knitr.svg)](https://travis-ci.org/yihui/knitr)
+[![Coverage Status](https://img.shields.io/coveralls/yihui/knitr.svg)](https://coveralls.io/r/yihui/knitr?branch=master)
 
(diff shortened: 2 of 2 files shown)


yihui:
Thanks! This is very interesting and useful. One question, though: since I do not use testthat (I use my own tiny package [testit](https://github.com/yihui/testit) instead), does covr still work in that case? Or actually, does covr really have to "endorse" a specific testing framework? I have noticed covr a while ago, and I have been waiting for a better documentation (e.g. a package vignette) about how it actually works. Otherwise it is completely magic to me, and I'm not sure if it really works in my case. I tried to read the source code, but it does not seem to be something that I can understand in a few minutes :)

yihui:
BTW, while I'm reading the help pages, I have a humble suggestion that you export as fewer functions as possible, e.g. I do not quite see the value of exporting `clear_counters`, `count`, `key`, and so on. Two reasons: 1. avoid names clashing with other packages; 2. exported objects are kind of like promises to users that these are mature objects to use and won't change much in the future.

.travis.yml
```diff
@@ -19,7 +19,10 @@ before_install:
     fi
 
 # install dependencies
+after_success:
+  - Rscript -e 'library(covr);coveralls()'
 install:
+  - ./travis-tool.sh github_package jimhester/covr
```
yihui: Perhaps surprising to you again, I do not use travis-tool.sh, either, since I think it is trivially easy to set it up by myself, and I'm more familiar and comfortable with `apt-get` :)

jimhester:
`covr` just sources the files in your `tests`, `inst/tests` directories, so it should work with any testing framework.  I verified it does work with `testit` and the `knitr` tests before submitting this pull request.

re: exporting functions that is a fair criticism, I un-exported them at jimhester/covr@e77b2a5147435a0dfa985087a9ece77eeb0b598 , thank you for the suggestion and I agree with your reasoning.

I had previously seen that you did not use `r-travis`, but forgot about that fact when I was submitting this pull request, my apologies, I modified the yaml with https://github.com/hadley/devtools/pull/681 and clearly did not examine the file before submitting, my apologies.

yihui:
No no I was not criticizing you exporting functions. It was just a suggestion :)

Now it seems everything is good. Could you squash your three commits into one? e.g. `git reset --soft HEAD~3` (assuming you are on the covr branch) then commit and force push (https://github.com/yihui/knitr/blob/master/CONTRIBUTING.md)

BTW, I'd still love to see a vignette describing how covr works under the hood. I understand you may not want to spend time on it, but I'm very interested in your magic :+1:

jimhester:
Ok I squashed everything into one commit!  Let me know if there is anything else you need.

I think your vignette suggestion is a good one, a couple other people have wanted to know how it works as well.  I will put one together soon.

yihui:
Excellent. Thanks!

PR #938, opened 2015-01-17
try to support clojure

kindlychung:
I made some changes so the following appears to work:

    > eng_interpreted(list(engine="lein", engine.opts="exec -e", eval=TRUE, code="(print \"This is a long string.\")", eval=TRUE, error=FALSE, echo=TRUE, results="show"))
    running: lein exec -e  '(print "This is a long string.")'
    [1] "(print \"This is a long string.\")\nThis is a long string.\n"

But when I try to actually compile the knitr-clojure.Rmd, it complains:

    Quitting from lines 17-19 (knitr-clojure.Rmd) 
    Error in in_dir(opts_knit$get("root.dir") %n% input_dir(), (knit_engines$get(options$engine))(options)) : 
    attempt to apply non-function
    Calls: <Anonymous> ... process_group.block -> call_block -> block_exec -> in_dir
    Execution halted

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
-Version: 1.8.14
+Version: 1.8.15
 Date: 2014-11-11
 Authors@R: as.person(c(
     "Adam Vogt [ctb]",


kindlychung:
This requires the lein-exec package to be installed, btw.

R/engine.R
```diff
@@ -85,8 +84,9 @@ eng_interpreted = function(options) {
     }, haskell = paste('-e', shQuote(paste(':script', f))), f)
   } else paste(switch(
     engine, bash = '-c', coffee = '-e', groovy = '-e', node = '-e', perl = '-e',
-    python = '-c', ruby = '-e', scala = '-e', sh = '-c', zsh = '-c', NULL
+    python = '-c', ruby = '-e', scala = '-e', sh = '-c', zsh = '-c', lein="exec -e", NULL
```
yihui: Style: spaces around `=` please.

R/engine.R
```diff
@@ -66,7 +66,6 @@ engine_output = function(options, code, out, extra = NULL) {
 
 ## TODO: how to emulate the console?? e.g. for Python
 #  see some experiments at https://github.com/yihui/runr
-
```
yihui: If there is no good reason to delete blank lines, please do not do so.

inst/examples/knitr-clojure.Rmd
```diff
@@ -0,0 +1,19 @@
+---
```
yihui: Thanks for adding an example, but please add it to https://github.com/yihui/knitr-examples instead.

R/engine.R
```diff
@@ -85,8 +84,9 @@ eng_interpreted = function(options) {
     }, haskell = paste('-e', shQuote(paste(':script', f))), f)
   } else paste(switch(
     engine, bash = '-c', coffee = '-e', groovy = '-e', node = '-e', perl = '-e',
-    python = '-c', ruby = '-e', scala = '-e', sh = '-c', zsh = '-c', NULL
+    python = '-c', ruby = '-e', scala = '-e', sh = '-c', zsh = '-c', lein="exec -e", NULL
   ), shQuote(paste(options$code, collapse = '\n')))
+  print(code)
```
yihui: I do not think I need this.

yihui:
Okay, please make the changes, clean the GIT history, and force push: https://github.com/yihui/knitr/blob/master/CONTRIBUTING.md Thanks!

kindlychung:
Hi, Yihui. The at the moment example document cannot really be successfully compiled, as I said in the pull request message, there is this error:

    Error in in_dir(opts_knit$get("root.dir") %n% input_dir(), (knit_engines$get(options$engine))(options)) : 
    attempt to apply non-function
    Calls: <Anonymous> ... process_group.block -> call_block -> block_exec -> in_dir
    Execution halted
 
What is perplexing is that when I try the following in the R console, it seems to work:

    > options = list(engine="lein", eval=TRUE, code="(print \"This is a long string.\")", eval=TRUE, error=FALSE, echo=TRUE, results="show")
    > (knit_engines$get(options$engine))(options)
    [1] "(print \"This is a long string.\")\nThis is a long string.\n"


Any ideas? (I think it makes more sense to fix this error before pushing again. )

yihui:
Did you install the package with your changes? Try to print `names(knit_engines$get())` in the example.

kindlychung:
Ok, I see the problem now. I though the modified package was already installed, but it was not because pandoc-citeproc was not in PATH, now everything seems to be working. 

There is still a minor problem though, clojure code block does not have a colored background (while python and R do have), here is a screenshot from the pdf output:

screen shot 2015-01-20 at 14 42 55

Any ideas how to fix this?

PR #956, opened 2015-02-18
Added 'http:' prefix in links

mg14:
Hi Yihui
Adding the http prefix allows to load the css and js files also when opening the html reports from a file. 
Cheers Moritz

PS Now in a separate branch.


yihui:
Thanks, but that was intentional. Please see http://stackoverflow.com/q/4831741/559676

If you really want to play with DataTables, I'd recommend you to try the DT package instead: https://github.com/rstudio/DT

PR #968, opened 2015-03-10
Add newline to fix multiple plots in a chunk for asciidoc

dnusinow:
Multiple plots in a chunk (as in a loop) would concatenate the title
of the next plot on to the code to specify the previous plot, which
confused the asciidoc program. A simple newline fixes the problem.


yihui:
Thanks!

PR #972, opened 2015-03-12
Start header in a new line

hofnerb:
Currently, the insertions in `LaTeX` are directly after the `\documentclass{}` command. This works but is not very pretty in the output. Pehaps a newline (`\n\n`) would be benefitial. 

Furthermore, one could add some comments to the definitions and also wrap the whole block in a comment which I didn't do at the moment but am happy to contribute.

### DIFF
--- a/R/header.R
+++ b/R/header.R
@@ -15,7 +15,7 @@ insert_header = function(doc) {
 # Makes latex header with macros required for highlighting, tikz and framed
 make_header_latex = function() {
   h = paste(c(
-    sprintf('\\usepackage[%s]{graphicx}\\usepackage[%s]{color}',
+    sprintf('\n\n\\usepackage[%s]{graphicx}\\usepackage[%s]{color}',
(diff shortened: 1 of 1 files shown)


yihui:
Thanks, but I do not care much about the "prettiness" here... Normally I do not read the .tex output at all unless I run into trouble and need to debug the issue.

hofnerb:
Perhaps others (such as myself) do (at least to some extend). You are right that one usually doesn't read the tex source but sometimes one will. Formating the code and adding some comments would help others to use and tweak your LaTeX header. 

With the same argument one could state that pretty code is not necessary (as one uses the function and doesn't look into it). But I am sure that you also will object here.

yihui:
Pretty code is necessary in general, but the importance of prettiness can vary in different cases. To me, being pretty in R and source code is important, but being so in the auto-generated LaTeX code is not, since 99% of the time I read the final PDF output instead of the intermediate .tex output. Sorry.

PR #975, opened 2015-03-15
Add support for tikzscale package, remove extra whitespace

ilia-kats:
This patch adds support for the tikzscale package, making it possible to generate tikz plots with external=FALSE while still being able to scale them to the desired size. In addition, several linebreaks in the generated LaTeX code are commented out, as they led to extra whitespace being inserted before the plot, which is very annoying when trying to have multiple plots in one figure using the subfig package.

### DIFF
--- a/R/header.R
+++ b/R/header.R
@@ -20,7 +20,8 @@ make_header_latex = function() {
             opts_knit$get('latex.options.color') %n% ''),
     .header.maxwidth, opts_knit$get('header'),
     if (getOption('OutDec') != '.') '\\usepackage{amsmath}',
-    if (out_format('latex')) '\\usepackage{alltt}'
+    if (out_format('latex')) '\\usepackage{alltt}',
+    if (test_latex_pkg('tikzscale', '', FALSE)) '\\usepackage{tikzscale}'
   ), collapse = '\n')
   if (opts_knit$get('self.contained')) h else {
     writeLines(h, 'knitr.sty')
--- a/R/hooks-latex.R
+++ b/R/hooks-latex.R
@@ -64,6 +64,7 @@ hook_plot_tex = function(x, options) {
   }
 
   tikz = is_tikz_dev(options)
+  tikzscale = test_latex_pkg('tikzscale', '', FALSE) & xor(!is.null(options$out.width), !is.null(options$out.height))
 
   a = options$fig.align
   fig.cur = options$fig.cur %n% 1L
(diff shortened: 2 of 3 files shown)


yihui:
Thanks, but why not just insert `\usepackage{tikzscale}` in your LaTeX preamble? I do not quite see the point of adding this automatically to all LaTeX documents.

ilia-kats:
You need to use \includegraphics in order for tikzscale to do its thing, 
and the knitr behaviour up until now was to use \input when 
external=FALSE. And if we now use \includegraphics for plain tikz files 
we also need to automatically include tikzscale in the preamble.

yihui:
I see. I think the `resize.width`/`resize.height` options were designed for rescaling `\input{foo.tikz}`, although this is the first time I have seen they could be useful to a real user. Here is a minimal example:

```
\documentclass{article}
\usepackage{tikz}
\begin{document}

<<foo, dev='tikz', external=FALSE, resize.width='.7\\textwidth'>>=
plot(1:10)
@

\end{document}
```

Is there a reason why you prefer `external=FALSE`? It makes use of your LaTeX preamble, and makes it faster to compile the document when `cache=TRUE`.

ilia-kats:
The difference between resize.width and tikzscale is that resizewidth 
also scales the fonts, while tikzscale keeps the fontsize constant and 
just scales everything else. So if I get my graphics device size close 
to what I want as final size, I can use tikzscale for finetuning using 
''.7\\textwidth'' or somesuch. This is also the reason why I prefer 
external=FALSE, this way the fonts get rendered with the settings of the 
environment the figure is in, not just the global settings.

ilia-kats:
Oh, and as far as speed, goes, I found a workaround to make tikzexternalize work, so that's not an issue. Briefly:
```
\usetikzlibrary{external}
\ifx\iftikzinternalcomputation\undefined\tikzexternalize\fi

<<setup, include=FALSE, cache=FALSE>>=
    library(knitr)
    library(tikzDevice)
    options(tikzXelatexPackages = c(
    getOption( "tikzXelatexPackages" ),
    "\\newif\\iftikzinternalcomputation",
    "\\tikzinternalcomputationtrue"
    ))
@
```

yihui:
Closing as stale. The tikzscale support remained inconclusive, and you'd found a workaround via tikz externalization. Happy to revisit in a new PR if there's renewed demand. Thanks!

PR #984, opened 2015-04-01
added switch for stata engine

muschellij2:
This should work on Unix-based systems, but it doesn't seem fully operational still for what is needed.


yihui:
Thanks! I'm just going to trust you guys since I have no idea about the command-line usage of stata.

PR #985, opened 2015-04-06
Add adjustments for East Asian wide characters

### LINKED ISSUE #980: kable() cannot handle East Asian wide characters when format="pandoc"
Thanks for your great work!

I got problem when I use `kable()` with some data containing East Asian wide characters. `kable()` seems to assume that their width is 1 unit, while [pandoc thinks their width is 2 units](https://github.com/jgm/pandoc/blob/b983adf0d0cbc98d2da1e2751f46ae1f93352be6/src/Text/Pandoc/Pretty.hs#L489-L490). So, table is not knitted correctly when some row contains more-than-three “East Asian wide characters” because each column is only two spaces away.

The code bellow is a minimal example; third and fourth row are knitted incorrectly. Knitted version of example is here: https://rpubs.com/yutannihilation/66064

```r
a <- read.csv(text = '
    "A",  "B", "C",  "D",  "E"
    "a",  "a", "a", 1.23, 1.23
    "囧", "a", "a", 1.23, 1.23
    "囧","囧","囧", 1.23, 1.23
    "囧囧囧","a","a", 1.23, 1.23', stringsAsFactors = FALSE)
kable(a, format="pandoc")
#> 
#> 
#> A     B     C        D      E
#> ----  ----  ---  -----  -----
#> a       a    a    1.23   1.23
#> 囧      a     a    1.23   1.23
#> 囧     囧     囧     1.23   1.23
#> 囧囧囧   a     a     1.23   1.23
```

I know this is avoidable if I use `format="markdown"`, though. I'm not sure if this should be treated as a known issue or a bug...

yutannihilation:
fixes #980

### DIFF
--- a/R/table.R
+++ b/R/table.R
@@ -225,13 +225,13 @@ kable_mark = function(x, sep.row = c('=', '=', '='), sep.col = '  ', padding = 0
   if (sep.col == '|') for (j in seq_len(ncol(x))) {
     x[, j] = gsub('\\|', '&#124;', x[, j])
   }
-  l = if (prod(dim(x)) > 0) apply(x, 2, function(z) max(nchar(z), na.rm = TRUE))
+  l = if (prod(dim(x)) > 0) apply(x, 2, function(z) max(nchar(z, type = 'width'), na.rm = TRUE))
   cn = colnames(x)
   if (length(cn) > 0) {
     cn[is.na(cn)] = "NA"
     if (sep.col == '|') cn = gsub('\\|', '&#124;', cn)
     if (grepl('^\\s*$', cn[1L])) cn[1L] = rownames.name  # no empty cells for reST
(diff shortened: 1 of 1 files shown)


yihui:
Thanks a lot! It looks like a nice fix. It requires the development version of stringr, though, which I don't know when will be released to CRAN. Perhaps you can write your own version of `str_pad()` at the moment. We can replace it when the next version of stringr is on CRAN.

yutannihilation:
> It requires the development version of stringr, though

Ah... Sorry, I wasn't aware of that. Then, I will try to write my own `str_pad()`.

yutannihilation:
I use `sapply()` to handle vector of widths. How about this?

yihui:
That works, but I guess it is slow. I will improve it by myself later. Thanks!

yutannihilation:
Wow, thanks! :+1:

PR #989, opened 2015-04-08
fix Stata linux command flag

Hemken:
And I will also confirm that all the changes we have made to SAS and Stata engines and engine output work as expected on linux.


yihui:
Thanks!

PR #997, opened 2015-04-17
knit_params function for extracting params from document yaml

jjallaire:
Not ready for merge yet, just putting up for discussion.

@yihui I am not quite sure how to weave this into `purl`, perhaps you can help me there? (I added you as  a committer to jjallaire/knitr in case you want to push changes directly to that branch).

@jcheng5 I think this reflects what we discussed yesterday. The schema is open-ended in that we don't validate fields (other than to make sure there is always a `value` field). 

The supported types don't use arbitrary R code (not sure exactly how we could do this anyway as the yaml_load function needs a static list of handers, perhaps we could do it if we did two passes over the yaml?). Anyway, for the first version I think we can get way with not worrying about arbitrary extensibility.

cc: @jmcphers @aronatkins


jjallaire:
See also rmarkdown PR which uses knitr::knit_params to implement parameterized reports: https://github.com/rstudio/rmarkdown/pull/413

yihui:
I have completed the purl part, and I'm ready to merge this PR unless other people have more thoughts.

R/params.R
```diff
@@ -0,0 +1,264 @@
...
+  # special 'type' designator as a hint to front-ends
+  # (e.g. 'file' to indicate a file could be uploaded)
+  type_handler <- function(type) {
+    function(value) {
+      attr(value, "type") <- type
+      value
+    }
+  }
+
+  list(
```
jjallaire: One open question is what base types we need at the outset. We already support all the base R types that yaml handles (character, logical, numeric, etc.) 'file' is just a way of tagging a character vector as a file. 'date' uses the as.Date function to convert character to date. I'm wondering if we need a 'datetime' type based on e.g. as.Posixct?

yihui: A `datetime` type sounds good to me.

jcheng5:
LGTM

R/params.R
jjallaire: Okay, I'll add that later today.

jjallaire: Just added support for `datetime` as well as some additional documentation on custom types so I think we should be good to merge.

PR #999, opened 2015-04-19
add stringr:: prefix to relevant functions

robertzk:
This kind of style has been favored heavily in the R community recently as explicitly referencing the package makes it clearer that this depends on an external function.

Let me know if you like it and I'll find any remaining stragglers.


yihui:
I do not have a strong preference here, but I will accept it anyway. Thanks!

robertzk:
Thanks!

PR #1000, opened 2015-04-19
explicit inherits = FALSE for .Random.seed

robertzk:
If an attached package exports a function `.Random.seed` it will heavily confuse knitr. Better to be safe than sorry.


yihui:
Agreed.

PR #1001, opened 2015-04-19
use [:xdigit:] for hexadecimal regex

robertzk:
What do you think of this version? [Source](https://stat.ethz.ch/R-manual/R-devel/library/base/html/regex.html).

### DIFF
--- a/R/cache.R
+++ b/R/cache.R
@@ -104,7 +104,7 @@ known_globals = c(
 cache = new_cache()
 
 # a regex for cache files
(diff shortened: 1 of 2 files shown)


yihui:
I think I still prefer the explicit list `abcdef0123456789` here. It is not too long, and `[[:xdigit:]]` can be locale/char set dependent (I do not really understand the consequence).

yihui:
Note Travis CI failed in this case.

robertzk:
Thanks! Makes sense.

PR #1002, opened 2015-04-19
reuse in_dir helper

robertzk:
You already have a helper for it. What do you think of reusing the helper?


robertzk:
I promise I didn't change anything except wrapping with `in_dir`!

yihui:
You don't need to promise. Let Travis CI defend your PR :)

In this case, I don't quite like it. I use `in_dir()` for shorter expressions on the scale of three or five lines. When the expression is very long, I'd rather do `on.exit()` at the top, and the main reason is to reduce the level of indentation. Normally I try to avoid indentation if possible.

robertzk:
Makes sense, thank you for the comments.

PR #1003, opened 2015-04-19
better .Random.seed fetch

robertzk:
Maybe I'm missing something, but are these not equivalent in all circumstances? (except when `.Random.seed` is not in the global namespace, but *is* exported by an attached package, which I have already covered in a separate pull request)

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -285,11 +285,12 @@ eng_asis = function(options) {
 }
 
 # set engines for interpreted languages
-for (i in c(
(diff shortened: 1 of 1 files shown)


yihui:
Looks good to me. But please resolve conflicts before I'm able to merge.

robertzk:
Should be okay now.

yihui:
Can you squash the two commits into one? I guess you must be a GIT expert, but here are some commands that I use very often: https://github.com/yihui/knitr/blob/master/CONTRIBUTING.md#pull-requests

robertzk:
Thanks! I will take a look at the git commands. I do have a good collection of aliases.

PR #1005, opened 2015-04-19
use href in citation docs

robertzk:
What do you think of this format?


yihui:
The downside is that the plain text help will not display the URL. Only HTML help will.

robertzk:
I thought that may have been the reason!

PR #1008, opened 2015-04-19
remove stray rm

robertzk:
The `rm(i)` looks awkward and alone here.


yihui:
I have no strong preference here.

PR #1009, opened 2015-04-19
slightly cleaner if statement block

robertzk:
What do you think of this version?

### DIFF
--- a/R/header.R
+++ b/R/header.R
@@ -3,13 +3,11 @@
 
 # doc is the output of processed document
 insert_header = function(doc) {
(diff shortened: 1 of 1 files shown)


yihui:
It looks better, but personally I do not like `else` to be on a new line. The reason is that if I remove the `else` clause some day, I may fail to notice that the `if` clause returns an implicit `NULL` to the function.

I'm not the original author of these lines of code. If I were to write them, I'd put `return` directly after `if` on the same line since the `return` statements are reasonably short.

robertzk:
Thanks, that makes a lot of sense!

PR #1016, opened 2015-04-19
more namespacing of packages

robertzk:
Note that `evaluate::is.error` is never actually used, so I removed it from the `#' @imports` section and the NAMESPACE file. If it *is* used metaprogramatically and I failed to grep for it as such, I will restore it.

Otherwise, this pull request follows in the spirit of https://github.com/yihui/knitr/pull/999.


R/package.R
```diff
@@ -17,7 +17,7 @@
 #'   + \code{weave}).
 #' @references Full documentation and demos: \url{http://yihui.name/knitr/};
 #'   FAQ's: \url{http://bit.ly/knitr-faq}
-#' @importFrom evaluate evaluate is.recordedplot is.source is.warning is.error
```
robertzk: Removed due to lack of usage. I may be wrong if metaprogramming is at play.

yihui:
Again, I hope you can get rid of the second commit. If you want to sync up with the master branch, you need to pull _with rebase_:

    git checkout more_namespacing
    git pull --rebase https://github.com/yihui/knitr.git more_namespacing

`git rebase more_namespacing~2 -i` may help you clean up the history (use `fix` on the second commit)

R/package.R
yihui: Seems good to me.

robertzk:
Thank you, I know what went wrong on my end. Does it look better now?

PR #1018, opened 2015-04-19
use lapply for hooks

robertzk:
I believe these are equivalent.

### DIFF
--- a/R/hooks-html.R
+++ b/R/hooks-html.R
@@ -168,10 +168,7 @@ render_html = function() {
   }
   h = opts_knit$get('header')
   if (!nzchar(h['highlight'])) set_header(highlight = .header.hi.html)
-  z = list()
-  for (i in c('source', 'warning', 'message', 'error'))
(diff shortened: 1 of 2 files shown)


R/hooks-html.R
```diff
@@ -168,10 +168,7 @@ render_html = function() {
   }
   h = opts_knit$get('header')
   if (!nzchar(h['highlight'])) set_header(highlight = .header.hi.html)
-  z = list()
-  for (i in c('source', 'warning', 'message', 'error'))
-    z[[i]] = html.hook(i)
-  knit_hooks$set(z)
+  knit_hooks$set(lapply(c('source', 'warning', 'message', 'error'), html.hook))
```
yihui: No, they are not equivalent. `z` is a named list, and `lapply()` returns an unnamed list.

robertzk: Yikes... You are certainly correct. We could use a `setNames` trick but then it becomes more trouble than it is worth.

PR #1021, opened 2015-04-19
just grab .Random.seed directly

robertzk:
Too many master changes, I had to make a new PR to squish. Does this look better?


yihui:
I don't think you need to sync up with the master branch before you can squash two commits. `git reset --soft HEAD~2` should do it, if you are on that PR branch.

PR #1023, opened 2015-04-23
Add informative error if ffmpeg is not found

cpsievert:
Note that `brew install ffmpeg` didn't quite work, but `brew install ffmpeg --with-libvpx` did


R/hooks-html.R
```diff
@@ -78,6 +78,10 @@ hook_ffmpeg = function(x, options, format = '.webm') {
 
   ffmpeg.cmd = paste('ffmpeg', '-y', '-r', 1 / options$interval,
                      '-i', fig.fname, mov.fname)
+  if (nchar(Sys.which("ffmpeg")) == 0)
```
yihui: Normally I just use `Sys.which("ffmpeg") == ''`, and I'm not 100% sure if it is equivalent to `nchar() == 0`.

PR #1034, opened 2015-05-13
closes #1031: add option to suppress output from cropping utility

### LINKED ISSUE #1031: quiet option for plot_crop()
`plot_crop()` should gain a `quiet` argument with `opts_knit$get("progress")` as its default value such that `knit(, quiet = TRUE)` is as quiet as it should be. Specifically, in case `quiet = TRUE`:
* omit `message("cropping ", x)`
* set `ignore.stdout = TRUE` in the system call.

This fixes the question at http://stackoverflow.com/questions/29077093/how-to-suppress-pdfcrop-printing-when-using-r-rmarkdown-render-function

Agreed? If yes, should I submit a pull request?

WastlM:
The new `quiet` argument defaults to `!opts_knit$get("progress")`
such that standard output is suppressed for `knit(, quiet = TRUE)`.
The more portable `system2()` is now used instead of `system()`
to call the command line cropping utility.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -281,21 +281,28 @@ fig_process = function(FUN, path) {
 #' \command{convert} is a command in ImageMagick (Windows users may have to put
 #' the bin path of ImageMagick into the \var{PATH} variable).
 #' @param x the plot filename
+#' @param quiet whether to suppress standard output from the command line utility
 #' @export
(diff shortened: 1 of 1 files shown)


yihui:
I have seen multiple reports in which `system2()`/`system()` failed to work on Windows. Anyway, I merged this one and will see if `system2()` is still problematic for Windows users. Thanks!

PR #1037, opened 2015-05-15
example for tweak=FALSE

WastlM:
use `digest` to illustrate `tweak=FALSE`, since `rpart` and `survival` are tidy


yihui:
Thanks!

PR #1044, opened 2015-05-21
No warning if chunk with child has only whitespace

scottkosty:
This warning is useful to make sure authors are aware that code in
the chunk will not be run. But because pure whitespace is not code,
there is no concern.

This commit has the extra benefit of increasing compatibility with
LyX's chunk inset. An empty chunk inset in LyX consists of a blank
line, and would thus trigger the warning without this commit.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -31,7 +31,7 @@ call_block = function(block) {
   if (opts_knit$get('progress')) print(block)
 
   if (!is.null(params$child)) {
-    if (length(params$code)) warning(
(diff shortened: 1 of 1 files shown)


yihui:
That makes sense. Please use `!is_blank(params$code)`.

scottkosty:
Thanks for the quick reply, Yihui. I made your suggested change.

yihui:
Perfect. Thanks!

Issue #1053, opened 2015-06-08
Inline numerics not rounded correctly


numatt:
There seem to be two problems with how knitr is rounding numerics in inline blocks:
1. Rounding to getOption("digits") decimal places, instead of significant digits
2. Limiting display to 15 significant digits, instead of the 22 generally supported in R

The below Rmd demonstrates both problems. I'm planning to submit a pull request with a proposed fix. 

---
    title: "knitr_rounding"
    output: html_document:
    keep_md: true
    ---
    
    # Knitr uses decimal places instead of signficant digits
    
    ## Examples where inline value differs from R standard
    ```{r}
    options(digits=1)
    pi
    ```
    knitr inline: `r pi`  
    **pi to 1 significant digit should be 3, not 3.1**
    
    ```{r}
    options(digits=2)
    pi
    ```
    knitr inline: `r pi`  
    **pi to 2 significant digits should be 3.1, not 3.14**
    
    ## Technical details
    To render numerics, knitr uses `round(x, digits=getOption("digits"))`. This is
    problematic because `options("digits")` defines the number of significant digits
    desired (see `?options` and `?print.default`). However, the digits parameter for
    `round()` sets the number of decimal places, not significant digits. There is a
    separate rounding function, `signif()`, which rounds to the number of signficant
    digits. Also, the `format()` method by default converts a numeric to a string,
    rounding to `getOption("digits")` significant digits.
    
    # Knitr limited to 15 significant digits inline (R limit is 22)
    
    ## Example where inline value is differs from R standard
    ```{r}
    options(digits=22)
    pi
    ```
    knitr inline: `r pi`  
    **pi to 22 significant digits should be 3.141592653589793115998**
    
    ## Technical details
    knitr uses `as.character()` and `sprintf()` to convert numerics to strings.
    However, `as.character()` only supports up to 15 significant digits (see
    `?as.character`), while R generally supports 22 significant digits. The
    `format()` method is an alternative that supports the full 22 digits, and also
    provides rounding for numerics.

PR #1054, opened 2015-06-08
handling of numeric rounding

numatt:
Proposed fix to match R numeric rounding standards. Addresses issue #1053.

Also includes updated and new tests for numeric rounding cases.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,6 +67,14 @@ To make sure you did not break anything, you need to run tests, which are
 done through the [**testit**](http://cran.rstudio.com/package=testit)
 package. If you added any features, add your own tests in `tests/testit/`.
 
+You can build the package and run the tests from a shell. Ensure that the root
+"knitr" project directory is in your working directory, then:
+
+```bash
+R CMD build knitr
+R CMD check knitr_<version>.tar.gz
+```
+
 ## Travis CI
 
 If you are lazy or do not understand what I said above, just push commits to
--- a/R/hooks.R
+++ b/R/hooks.R
@@ -1,6 +1,5 @@
 # format a single inline object
 .inline.hook = function(x) {
-  if (is.numeric(x)) x = round(x, getOption('digits'))
   paste(as.character(x), collapse = ', ')
 }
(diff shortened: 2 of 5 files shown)


numatt:
The rounding changes impact the precision of the results in the example files. Unfortunately, I'm unable to update the examples because the knit-all.R script crashes R on on my machine. Please feel free to review the proposed fix, and adopt something similar if you feel it's worthwhile. Thanks.

yihui:
Thanks! I think this is a correct fix in theory. I cannot speak for other users, but personally I rarely find significant digits useful in practice. I often need to control the number of decimal places instead. I was abusing `getOption('digits')`.

numatt:
Makes sense.

What bothered me was seeing different results in code chunks and inline chunks. I'm currently taking the Data Science series of courses on Coursera, and for new users at least, it's confusing and a bit frustrating when different packages inexplicably print out numerics in different formats. It makes it feel like you can't rely on the conventions or documentation.

I do tend to agree with your view that specifying decimal places may be more useful in practice. Would be nice to enable this in a way that doesn't abuse getOptions('digits'). :)

Regardless, thanks for the great package. I've been using it extensively.

yihui:
I agree and I can see the frustration. Perhaps we can just replace `getOption('digits')` with something like `getOption('knitr.inline.digits', getOption('digits'))` so it is possible to configure the number of decimal places (fall back to the `digits` option if `knitr.inline.digits` is not specified).

numatt:
Thanks. That provides a way of setting the decimal digits without abusing getOption('digits'). But, I think it doesn't provide an option of using the standard R behavior of rounding by significant digits. It would be nice to have a way to configure knitr so that printing a numeric in a code block gives the same results as printing it inline.

So, maybe something like `getOption('knitr.inline.rounddecimal', TRUE)`. The default would use the current rounding logic (rounding based on decimal places), and the alternative would be to round based on significant digits.

yihui:
The `knitr.inline.rounddecimal` idea sounds good to me. If you are still interested in it, please feel free to update this PR. Thanks!

PR #1069, opened 2015-07-09
support for !expr in knit_params

jjallaire:
- correct 'type' field for Date and POSIXct objects generated via !eval
- add 'expr' field to flag when a parameter's value was the result of an R expression)

@trestletech @aronatkins @jcheng5

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - added an argument `quiet` to `plot_crop()` (thanks, @WastlM, #1034)
 
+- support for !expr in knit_params (correct 'type' field for Date and POSIXct objects generated via !eval; add 'expr' field to flag when a parameter's value was the result of an R expression)
+
 ## MAJOR CHANGES
 
 - the two hook functions `hook_rgl()` and `hook_webgl()` have been moved from **knitr** to the **rgl** package (v0.95.1247) (thanks, @dmurdoch)
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 - added an argument `quiet` to `plot_crop()` (thanks, @WastlM, #1034)
 
-- support for !expr in knit_params (correct 'type' field for Date and POSIXct objects generated via !eval; add 'expr' field to flag when a parameter's value was the result of an R expression)
+- support for !expr in knit_params (correct 'type' field for Date and POSIXct objects generated via !expr; add 'expr' field to flag when a parameter's value was the result of an R expression)
 
 ## MAJOR CHANGES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,8 @@
 
 - added an argument `quiet` to `plot_crop()` (thanks, @WastlM, #1034)
 
-- support for !expr in knit_params (correct 'type' field for Date and POSIXct objects generated via !expr; add 'expr' field to flag when a parameter's value was the result of an R expression)
+- support for !expr and !r in knit_params.
+
 
 ## MAJOR CHANGES
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
-Version: 1.10.13
+Version: 1.10.14
 Date: 2015-05-06
 Authors@R: c(
     person("Adam", "Vogt", role = "ctb"),
--- a/R/params.R
+++ b/R/params.R
@@ -16,7 +16,7 @@
 #'     \item{\code{name}}{The parameter name.}
 #'     \item{\code{value}}{The default value for the parameter.}
 #'     \item{\code{class}}{The R class names of the parameter's default value.}
-#'     \item{\code{expr}}{The R expression that yielded the default value (if any).}
+#'     \item{\code{expr}}{The R expression (if any) that yielded the default value.}
 #'   }
 #'
 #'   In addition, other fields included in the YAML may also be present
@@ -55,7 +55,7 @@
 #' This second form is useful when you need to provide additional details
 #' about the parameter (e.g. a \code{label} field as describe above).
 #'
-#' You can use R code to yield the value of a parameter by prefacing the value
+#' You can also use R code to yield the value of a parameter by prefacing the value
 #' with \code{!r}, for example:
 #'
 #' \preformatted{
--- a/R/params.R
+++ b/R/params.R
@@ -14,12 +14,9 @@
 #'
 #'   \describe{
 #'     \item{\code{name}}{The parameter name.}
-#'     \item{\code{type}}{The parameter type. This can be a standard R object
-#'     type such as \code{character}, \code{integer}, \code{numeric}, or
-#'     \code{logical} as well as the special \code{date}, \code{datetime}, and
-#'     \code{file} types. See the \emph{Types} section below for additional
-#'     details.}
 #'     \item{\code{value}}{The default value for the parameter.}
+#'     \item{\code{class}}{The R class names of the parameter's default value.}
+#'     \item{\code{expr}}{The R expression that yielded the default value (if any).}
 #'   }
 #'
 #'   In addition, other fields included in the YAML may also be present
@@ -58,58 +55,16 @@
 #' This second form is useful when you need to provide additional details
 #' about the parameter (e.g. a \code{label} field as describe above).
 #'
-#' Parameter types are deduced implicitly based on the value provided. However
-#' in some cases additional type information is required (for example when
-#' a character vector needs to be interpreted as a date or as a file path).
-#' In these cases a special type designater precedes the value. For example:
+#' You can use R code to yield the value of a parameter by prefacing the value
+#' with \code{!r}, for example:
 #'
 #' \preformatted{
 #' ---
 #' title: My Document
 #' output: html_document
 #' params:
-#'   start: !date 2015-01-01
+#'   start: !r Sys.Date()
 #' ---
-#' }
-#'
-#' @section Types:
-#'
-#' All of the standard R types that can be parsed using
-#' \code{\link[yaml]{yaml.load}} are supported. These types are used
-#' implicitly based on the \code{value} provided so no special type
-#' designater is required. Built-in types include \code{character},
-#' \code{integer}, \code{numeric}, and \code{logical}.
-#'
-#' In addition there are a number of custom types used to represent
-#' dates and times as well as to note that character values have
-#' special semantics (e.g. are the name of a file). These types are
-#' specified by prefacing the YAML \code{value} with !\emph{typename},
-#' for example:
-#'
-#' \preformatted{
-#' ---
-#' title: My Document
-#' output: html_document
-#' params:
-#'   start: !date 2015-01-01
-#'   end: !datetime 2015-01-01 12:30:00
-#'   data: !file data.csv
-#' ---
-#' }
-#'
-#' Supported custom types include:
-#'
-#' \describe{
-#'   \item{\code{date}}{A character value representing a date.
-#'   The underlying date value is parsed from the character
-#'   value using the \code{\link[base]{as.Date}} function.}
-#'   \item{\code{datetime}}{A character value representing a
-#'   date and time. The underlying datetime value is parsed from
-#'   the character value using the \code{\link[base]{as.POSIXct}}
-#'   function. Note that these values should always speicifed using
-#'   UTC (Universal Time, Coordinated).}
-#'   \item{\code{file}}{A character value representing the name
-#'   of a file.}
 #' }
 #'
 #' @export
@@ -196,48 +151,39 @@ yaml_front_matter = function(lines) {
 # define custom handlers for knitr_params
 knit_params_handlers = function() {
 
-  # generic handler for intrinsic types that need a special 'type' designator as
-  # a hint to front-ends (e.g. 'file' to indicate a file could be uploaded)
-  type_handler = function(type) {
-    force(type)
+  # generic handler for r expressions where we want to preserve both the original
+  # code and the fact that it was an expression.
+  expr_handler = function() {
     function(value) {
-      attr(value, "type") = type
-      value
+      evaulated_value = eval(parse(text = value))
+      attr(evaulated_value, "expr") = value
+      evaulated_value
     }
   }
 
   list(
 
+    # r expressions where we want to preserve both the original code
+    # and the fact that it was an expression.
+    r = expr_handler(),
+    expr = expr_handler(),
+
+    # date and datetime (for backward compatibility with previous syntax)
+    date = function(value) {
+      value = as.Date(value)
+      value
+    },
+    datetime = function(value) {
+      value = as.POSIXct(value, tz = "GMT")
+      value
+    },
+
+    # workaround default yaml parsing behavior to allow keys named 'y' and 'n'
     `bool#yes` = function(value) {
       if (tolower(value) == "y") value else TRUE
     },
-
     `bool#no` = function(value) {
       if (tolower(value) == "n") value else FALSE
-    },
-
-    # date
-    date = function(value) {
-      value = as.Date(value)
-      attr(value, "type") = "date"
-      value
-    },
-
-    # datetime
-    datetime = function(value) {
-      value = as.POSIXct(value, tz = "GMT")
-      attr(value, "type") = "datetime"
-      value
-    },
-
-    # file
-    file = type_handler("file"),
-
-    # expr
-    expr = function(value) {
-      value = eval(parse(text = value))
-      attr(value, "expr") = TRUE
-      value
     }
   )
 }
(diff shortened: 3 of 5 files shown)


jcheng5:
Are `!expr` types different than the others (e.g. `datetime`) in that they are *always* fulfilled by the given expression, or can they also be overridden by user input (or arguments passed at runtime)? The latter meaning that basically these are just dynamic *default values* rather than a distinct parameter type?

aronatkins:
@jcheng5 I think `!expr` types are just types that need to be evaluated if not otherwise overridden. Think a report with a `start` and `end` date. Those might default to (expressions) `end=today` and `start=one-month-ago`. A user might want to manually override those to specific dates `end=2014-12-31` and `start=2014-01-01`.

Parameters defined by expression values can be overridden. They are not "read-only".

The fact that they are expressions might be revealed in a report-building UI - one could say that "2015-07-09" is the current value for `end` but that the value might change when the report is run. Or maybe the expression-backed parameter could be re-evaluated periodically, updating the UI.

The richness of the expression type also made us realize that one cannot simply compute a static schema describing both the parameters and their default values. That information can be extracted, yes, but it should probably be evaluated close to render-time.

jcheng5:
Then does that mean that a single parameter can be both `!expr` and, say, `!date`?

aronatkins:
Yes. The value produced by an expression has a type. Note that this PR is attempting to differentiate between the type of the value (`param$type`) and if it was produced by an expression (`param$expr`).

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 - added an argument `quiet` to `plot_crop()` (thanks, @WastlM, #1034)
 
+- support for !expr in knit_params (correct 'type' field for Date and POSIXct objects generated via !eval; add 'expr' field to flag when a parameter's value was the result of an R expression)
```
yihui: Looks like `!eval` is a typo? Should be `!expr` instead?

jjallaire: Indeed, just fixed!

jjallaire:
To reinforce what Aron said, !expr isn't a type the same way that !date and !datetime are types, rather it will resolve to a type (dynamically based on the expression result) after it's evaluated. It's essentially a dynamic default value and it can indeed be overridden.

The reason we have added the `expr` flag to knit_params is so that a front-end (e.g. Connect) knows to give the user a hint that the value will resolve to something dynamic when run (rather than being statically bound). For example, a report will always resolve to the last 30 days of data rather than the last 30 days from a static date. Granted that we don't yet have rich enough context to provide a really seamless UI (i.e. "Range: Last 30 Days") and perhaps we need to add another type primitive for specifying relative date ranges (e.g. the syntax nginx uses for cache expiration: https://serversforhackers.com/nginx-caching).

Something like this:

    ---
    params:
       range: !datewindow 30d
    ---

That would allow the UI to be much nicer than just knowing that an R expression yielded the parameter. 

Thoughts? Would this be a sound improvement?

aronatkins:
I think with your `datewindow` example, you want two pieces of information: Start date and window size.

    ---
    params:
       range: !datewindow 2014-09-05 30d
    ---

The disadvantage of this approach over separate `start` and `end` is that the report author is fixing the interval. Maybe that's OK, but removes flexibility in how someone might use the report in the future. I suppose that if someone really wants a fixed 30-day window, they could just ask for a `start` and then add 30d in R code..

My example has a fixed start-point, but I could also see wanting a window with a dynamic start date (30d ending with today; 30d ending with the a week ago). I think you were proposing an implicit start date. Very quickly, you want something like a `datewindow` with an expression defining the start point.

Maybe an alternate approach:

    ---
    params:
        range: !daterange
            start: !date 2015-07-04
            duration: !duration 1mo
            end: !date 2015-08-04
    ---

I've shown all three of `start`, `duration`, and `end`, but in truth, only two are needed. Type modifiers could be assumed (`start` is `!date`), but this syntax also allows alternate value types (`start` could be an expression!)

How would one override a range when rendering? Would they pass a tuple of dates? A start date and a duration?

Nothing feels quite right.

jjallaire:
I'm feeling like the creation of our own date and datetime "types" might be an anti-pattern when we have a type system in R that we can rely on (and be extensible based on). Perhaps we should do this:

    ---
    params:
       start: !Date 2015-01-01
       snapshot: !POSIXct 2015-01-01 12:30:00
    ---

Both of these can be converted from string using `as.Date` and `as.POSIXct` respectively. 

Note that we still need to register type handlers explicitly to do the conversion but at least we'd be using the same nomenclature as R. 

If we did this then the "type" field of knit_params would only have one purpose: to communicate extended type information to front-ends (e.g. a character vector is actually a filename so should be edited using a file uploader).  The "expr" field would also have only the purpose of communicating to a front end that a value was derived from an R expression (so the front end might want to provide clues that hard-coding it might be a bad idea).

I don't feel like we're converging on the right answers yet and I feel like we need to before getting much further.

aronatkins:
I'm OK with R type tags. We talked about language agnostic type tags because it might mean that a client didn't need to know R types to do the type mapping, but we're in an R world, so that's probably an acceptable burden.

The tags should really only be used when the natural type conversion doesn't work.

Here are the YAML type tags that already exist. There is a `!!timestamp` http://yaml.org/type/ and many timestamps should be naturally detected: http://www.yaml.org/spec/1.2/spec.html#id2761292 (but this is YAML-parser dependent; `r-yaml` doesn't appear to handle them).

jjallaire:
I think my datewindow example was misguided as we can easily capture these semantics with an integer, e.g.

    ---
    params:
        days: 45
    ---

Or:

    ---
    params:
        start: !expr Sys.Date()
        days: 45
    ---

I have an alternate formulation that might simplify things quite a bit. One insight is that front-ends will need to know both *that* an R expression was used as well as *what* the R expression was. An example would be a "front-end" that automatically generates an R function for running the report with parameters -- in that case you'd want to use the R expression as the default value of the parameter.

Another insight is that users will want to use arbitrary types. For example, you can imagine wanting to pass a matrix to a report (where the the default matrix is e.g. a randomly generated one). As a consequence of this we may want to dispense with the idea of a "type" field altogether (since we have the value as an R object which can be interrogated for it's class). In this world the only special type info would be front-end hinting for e.g. the "file" case, but that could be handled with the extended attribute syntax:

    ---
    params:
        dataset: 
            value: sample.csv
            file: true
    ---
 
`file: true` could also be written as `type: 'file'`. The whole point of the extended attribute syntax is to provide hints for front-ends so we should use it for that (rather than creating a "special" !file syntax). A bit more verbose but much more consistent with the overall model.

The logical consequence of the above is that `!type` goes away entirely, except for the case of R expressions which we evaluate and then record the original expression for:

    ---
    params:
       start: !expr as.Date("2015-01-01")
       snapshot: !expr as.POSIXct("2015-01-01 12:30:00")
    ---

To make this a bit more transparent we could introduce an alias for `!expr` called `!R`

    ---
    params:
       start: !R as.Date("2015-01-01")
       snapshot: !R as.POSIXct("2015-01-01 12:30:00")
    ---

This syntax is close to as readable as the `!date` and `!datetime` syntax but is infinitely extensible.

Front ends could then decide which R types they know how to provide editors for (the interesting types are relatively few and numerable I think).

jjallaire:
And coming back to our date window example, we could now write it this way:

    ---
    params:
       start: !R Sys.Date()
       months: 3
    ---

jcheng5:
@jjallaire has hit on what I was getting at, which is that type hinting is still needed in some cases, even when you have an R object in hand. I'm fine with the explicit `type: file` but please, please, not `file: true`.

I don't see what's wrong with keeping the !type shorthand though. It's something we can easily provide autocompletion for, and is far faster/easier to parse at runtime (in the case where there are no R expressions, we could generate the form UI without firing up an R process at all).

jjallaire:
Okay, we'll do "type: file" as well as support the shorthand "!file".

So with the current proposal the *only* time the type field is populated
would be for !file (and any other similar constructs we create). Native R
types like dates would be deduced by just inspecting the class of the
parameter value. This will allow us to support all R types and not have our
own shadow type system (e.g. "date", "datetime").

jcheng5:
It's not really all R types, right? We have to specifically dispatch on each R class and decide how to present it in the UI, don't we?

jjallaire:
I mean that the system is *open* to all R types without us having to
explicitly create new mappings. We can create UI handlers for the common
types.

I don't know if this is the best answer, just advancing it as a candidate.
If anyone thinks we should whitelist types in via mappings I'm not opposed
to that either.

aronatkins:
I'm ok with supporting any R data via parameters. UI support is a separate
issue. One could think of a report generator passing arbitrary data via
parameters in the render call. Let's not assume this is only UI driven.

jcheng5:
Gotcha, if the UI case is not the primary scenario then that sounds fine to me. OTOH, for the non-UI case, defaulting values (and converting default values from strings to more specific types) is pretty trivial to do inside of a code chunk anyway.

```
---
params:
  start: "2015-01-01"
  end: null
---

```{r}
start <- as.POSIXct(params$start)
end <- if (is.null(params$end)) Sys.time() else as.POSIXct(params$end)
```

jjallaire:
We do have a decision to make here. If the user uses the form you just
illustrated then there is no type info for us at the UI level. If they
rather use R code then there will be type-info. We definitely want/need the
type info at the UI level, the question is what the most straightforward
and long-term extensible mechanism is?

Allowing *any* R type has it's appeal as we invent no new data types or
rules (any expression that yields a date is valid vs. our having to define
what a valid date string is). The syntax for this is a bit more opaque
though.

What is your leaning here? We could do both easily, but if we don't want to
introduce and maintain a shadow type system now is the time to draw the
line in the sand.

aronatkins:
I'm leaning toward R types. The YAML-to-R already has some type conversion happening. Any UI/driver doesn't really care about parameter types in a language agnostic situation -- we are always mapping back to R -- so adding aliases for what really are R types hides what needs to be obvious.

Report authors are already R developers - and they (presumably) understand R types. 

What I like about using R types is that a report-developer could start using a new (not UI-driven) type in their report. If we're using R types, we don't need to update the entire toolchain to start supporting that in the UI. We only need update the UI builder and bit of code that calls `rmarkdown::render`.

If we have shadow types, each new type probably needs to flow through `knitr` and `rmarkdown` before UI support can be added.

jjallaire:
I'd agree that the R types are a bit more flexible and will require less foreknowledge of what users want to do in the toolchain. If you compare the syntax I don't think R types are much different and they benefit from not requiring special rules/documentation/validation:

```yaml
---
params:
   start: !r as.Date("2015-01-01")
   snapshot: !r as.POSIXct("2015-01-01 12:30:00")
---
```

vs.

```yaml
---
params:
   start: !date "2015-01-01"
   snapshot: !datetime "2015-01-01 12:30:00"
---
```

But of course the R dates are much more flexible as they support:

```yaml
---
params:
   start: !r Sys.Date()
   snapshot: !r Sys.time()
---
```

Then an auto-generated function for these params would be:

```r
runReport <- function(start = Sys.Date(), snapshot = Sys.time()) {
    rmarkdown::render("myreport.Rmd", params = list(
       start = start,
       snapshot = snapshot 
   ))
}
```

jcheng5:
You're always going to need the shadow types too, right? For constraints/types that don't have a corresponding R type?

jcheng5:
I feel like I'm missing a lot of context. Maybe one of you can catch me up in person on Monday.

jcheng5:
What's the R type equivalent of:

```r
params:
  birthdate: !datetime null
  maxrows: !int null
```

(i.e. either no sensible default is conceivable, or, `null`/`missing()` is the sensible default value)

jjallaire:
For the cases where users are going to do `if (is.null(params$foo)) {}` and then provide a default value within their R code I'd argue it's more transparent and self-documenting to provide it in the YAML so requiring that idiom is fine.

In the case where `NULL` is a legit value (`maxrows` being an excellent example) I think we're arguing that using R type names is better than our creating a parallel set of type names which must be learned and mapped to actual R types. For this case I'd argue that this form is more desirable:

```yaml
params:
  birthdate: !POSIXct null
  maxrows: !integer null
```

If the value is non-null we can always deduce the type from the object returned from yaml.load. 

Does this make sense and hold up to scrutiny?

This does suggest that the "type" field should be a vector so could therefore be:

```r
c("character", "file")
```

The `type` field then becomes `class(param$value)` combined with whatever other type hint the user provided (e.g. `type: file`).

jjallaire:
Looking at implementing this it would be much more flexible for the case of handling `NULL` to specify type as an object field (because then we wouldn't need a custom YAML `!` handler for every type we wan t to allow). For example:

```yaml
params:
  maxrows: 
     value: null
     type: integer
```

We could also encourage the use of sentinel values wherever that makes sense:

```yaml
params:
  maxrows: !r integer()
```

Or

```yaml
params:
  maxrows: -1
```

My main point is that having to meticulously maintain a set of static type mappings (and require changes to knitr every time we want a new type supported) is quite painful and inflexible. I wouldn't do that just to handle `NULL` more cleanly when we have reasonable workarounds (all the of the above would work) available.

jjallaire:
To further maintain the point about type extensibility, imagine that we wanted to support `data.frame` as an input type. The user could use this YAML:

```yaml
params:
  data: !r datasets::mtcars
```

The author of the report can now build it with the sample dataset and it's easy for us to automatically generate an R function that takes a `data` argument to provide other inputs. In this case RSConnect can look for the `data.frame` type and provide a file upload control as the user input.

This example is obvious as we'd naturally support `data.frame`. But imagine someone in Bioconductor wants to use one of the Bioconductor S4 objects as an input. With the shadow type system we'd need to make a change to knitr to support this type (as far as a I can see there is no "fallthrough" for `!` type handlers they all need to be explicit). You can imagine this playing out time and time again for time series objects, spatial objects, etc.

Granted that RSConnect is still going to need knowledge of how to provide user-input for e.g. a spatial type (I don't see a way around this save for some sort of S3 method that indicates how to provide input in RSConnect). I still think that's far better than requiring a change to knitr for every input type the user can dream up.

aronatkins:
I wonder if the fact that certain data came from a file might be better expressed as a separate property, not as a positional item in the `type` property.

(using yaml just for showing this metadata that could be produced):
```
render_param:
    name: data
    value: 1234
    type: integer
    source: file
```

Obviously, `r` (expressions) are another source which produce typed values. But we could eventually see other sources in play (database and URI come to mind) and might want ways of transforming those sources to the value format (think: data.frame/matrix/tabular).

At some point, this breaks down; I don't think we want to declare a complete prototyping system for R data structures. Any aggregate structure becomes difficult if different elements have different types.

jjallaire:
What's really fantastic about the idea of eliminating the `!file` construct is that it collapses the "type" construct into a pure R "class" (i.e. `param$class <- class(param$value)`) and removes all special knowledge of file-inputs from the knitr layer. This is as it should be as this is front-end specific hinting along the same lines as "label", "min", "max", etc.

I might argue that rather than `source: file` you could do `input: file` which would then be consistent with constructs like:

```yaml
params:
   beta:
      value: 0.5
      input:
         type: slider
         min: 0.0
         max: 1.0
```

aronatkins:
Oh, that's good. And for file inputs, you could embed a lot of other information, too, such as expected file type (for auto-ingestion).

```
params:
  data:
    value: <some-tabular-data>
    input:
        type: file
        content-type: csv
```

jjallaire:
I updated the PR to reflect this approach. The diff is big enough that I think it's not a useful way of reviewing the work at this point. Here's the new params.R in it's entirety (notice there is no longer any mention of "file" anywhere, it's now an orthogonal concern to knit_params):

https://github.com/jjallaire/knitr/blob/feature/knit-params-expr/R/params.R

jjallaire:
Joe, note that this approach is entirely compatible with our introducing type-hinting (note that the "date" and "datetime" fields are still supported for backwards compatibility) it just doesn't require type-hinting to get off first base.

jcheng5:
This approach is totally fine with me--as long as there's some way to be explicit about the (UI) type I'm fine with it.

For non-UI scenarios, will `param$class` be used for anything in particular?

jjallaire:
It's possible you could use the param$class for validation if you
automatically generated a function wrapper for calling the report with
parameters (although it's not clear whether you wouldn't need extra
guidance as the actual *required* class might be a subset of the classes
that param$value actually carries).

jjallaire:
Actually, I think param$class can be eliminated entirely (perhaps that's
what you were getting at) since it is now always == class(param$value).
I'll update the PR to reflect this.

jjallaire:
@yihui After much discussion I think we've settled on an approach here so we are good to merge assuming you think the code looks okay.

jjallaire:
@kevinushey Once this is merged we'll want packrat to call knit_params in order to grab the R expressions for dependency parsing.

jjallaire:
@yihui Thanks for merging this. Aron and I will be beginning work on
parameterized Rmd this week. Hopefully we can tease out any additional
issues/requirements for knit_params and then wrap this up and go to CRAN.
The mechanism we've come up with is very general and extensible both using
R and custom YAML fields so hopefully we are done mucking with knit_params!

PR #1073, opened 2015-07-13
Use tikzDevice for measuring plots when requested

### LINKED ISSUE #1066: Latex macros produce incorrect spacing in tikz device output
When I try to compile figures with latex math symbols (e.g. \times), it looks like knitr creates space for the word "\times" instead of for the symbol "x" which leaves a lot of ugly whitespace. Here are a few simple pictures illustrating what I see.

When I use "1 x 2"
selection_001

When I use "1 $\\times$ 2"
selection_002

When I use "1 $\\times \\Sigma \\vartheta$ 2"
selection_003

Here is a minimal example reproducing the results in the pictures above. https://drive.google.com/open?id=0BxYcpdRpKOV2X2xGbUtnMTVqa0E

krlmlr:
Fixes #1066.

Tested with the example given in #1066.

I think it would be great if @jremmons could submit a pull request to https://github.com/yihui/knitr-examples that shows the behavior we saw in this issue.

@yihui: Could you please double-check that my pull request doesn't break existing examples?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 - added a chunk option `ffmpeg.format` (default: `webm`) to control the video format of FFmpeg; this option specifies the filename extension of the video (#712)
 
+- plots drawn by `tikzDevice` are also measured and recorded with that device (thanks, @krlmlr, #1066)
+
 # CHANGES IN knitr VERSION 1.11
 
 ## NEW FEATURES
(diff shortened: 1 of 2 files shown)


yihui:
Thanks! We need a way to suppress the message "Creating new TikZ metrics dictionary at...". Travis CI failed because of that.

krlmlr:
Those are plain `message`s that are fired unconditionally. Suppressing them by knitr should work, but perhaps a `verbose = interactive()` argument to `tikz()` makes more sense.

yihui:
I'll merge this PR after it becomes possible to mute messages in tikzDevice.

krlmlr:
I've triggered a rebuild by merging your current `master`.

yihui:
Could you squash the three commits into one and force push (for the sake of a clean git history)?

krlmlr:
Done, also edited comments and added a NEWS item.

PR #1078, opened 2015-07-22
Split yaml_params.yaml out of knit_params

aronatkins:
Some contexts have already read and processed the YAML from the source document. `knit_params.yaml` supports those scenarios.

In addition, we do not always care about expression evaluation (such as when we are performing dependency analysis). The optional `evaluate` parameter allows this lighter-weight option.

This PR also bumps the `knitr` version.


NAMESPACE
```diff
@@ -59,6 +59,7 @@ export(knit_global)
 export(knit_hooks)
 export(knit_meta)
 export(knit_params)
+export(knit_params.yaml)
```
jjallaire: The `.` suffix for functions is usually reserved for S3 methods. I'd suggest just `knit_params_yaml`.

yihui:
This looks good to me. Let me know when you feel it is ready to merge. Thanks!

aronatkins:
I think we're going to make a bunch more changes to how we process the YAML, so let's hold on merging.

PR #1082, opened 2015-07-29
Allow YAML header to end with ...

noamross:
See https://github.com/rstudio/rmarkdown/pull/485

I ran `make check`, and got the following error, which appears unrelated as I get it when I run it without these changes:

```
* checking examples ... ERROR
Running examples in ‘knitr-Ex.R’ failed
The error most likely occurred in:

> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: hook_pdfcrop
> ### Title: Built-in chunk hooks to extend knitr
> ### Aliases: hook_optipng hook_pdfcrop hook_plot_custom hook_purl
>
> ### ** Examples
>
> if (require("rgl")) knit_hooks$set(rgl = hook_rgl)
Loading required package: rgl
Error in knit_hooks$set(rgl = hook_rgl) : object 'hook_rgl' not found
Execution halted
```

### DIFF
--- a/R/params.R
+++ b/R/params.R
@@ -10,7 +10,7 @@
 #' @param evaluate If TRUE, expression values embedded within the YAML will be
 #' evaluated. This is the default. When FALSE, parameters defined by an
 #' expression will have the parsed expression in its \code{value} field.
-#'
+#' 
 #' @return List of objects of class \code{knit_param} that correspond to the
 #'   parameters declared in the \code{params} section of the YAML front matter.
 #'   These objects have the following fields:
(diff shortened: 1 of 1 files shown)


R/params.R
```diff
@@ -10,7 +10,7 @@
 #' @param evaluate If TRUE, expression values embedded within the YAML will be
 #' evaluated. This is the default. When FALSE, parameters defined by an
 #' expression will have the parsed expression in its \code{value} field.
-#' 
+#'
```
yihui: Could you get rid of these changes of white spaces? Thanks!

PR #1090, opened 2015-08-13
Add a bitrate option to control the quality of webm videos.

JasonPunyon:
Greetings and Salutations. I hope you're having a great day.

First off, thank you very much for this awesome package. I started working with it 5 days ago and it's making my life a lot better.

I was making some animations of ggplot2 plots yesterday (super neat feature), but the quality of the resulting webm video was low (there was blockiness and blurriness in spots). 

As you call out in the documentation and from the R Markdown output the command line call for ffmpeg is something like this...

```
ffmpeg -y -r {fps} -i {working-dir}/figure-html/{chunk-name}-%d.png {working-dir}/figure-html/{chunk-name}-.webm
```

From the [ffmpeg documentation on VP8 encoding](https://trac.ffmpeg.org/wiki/Encode/VP8)..

> **Important**: If neither -b:v nor -crf are set, the encoder will use a low default bitrate and your result will probably look very bad. Always supply one of these options—ideally both.

To that end, this pull request changes the ffmpeg hook to read `bitrate` from the chunk options and pass it through to ffmpeg, or pass a default of 1Mbps through. It also adds the recommended default `-crf` value of 10.

Thanks again!

### DIFF
--- a/R/hooks-html.R
+++ b/R/hooks-html.R
@@ -75,9 +75,16 @@ hook_ffmpeg = function(x, options, format = '.webm') {
   # set up the ffmpeg run
   fig.fname = paste0(sub(paste0(fig.num, '$'), '%d', x[1]), '.', x[2])
   mov.fname = paste0(sub(paste(fig.num, '$', sep = ''), '', x[1]), format)
-
-  ffmpeg.cmd = paste('ffmpeg', '-y', '-r', 1 / options$interval,
+  
+  if (format == '.webm') {
+    mov.bitrate = ifelse(is.null(options$bitrate), "1M", options$bitrate)
+    ffmpeg.cmd = paste('ffmpeg', '-y', '-r', 1 / options$interval,
+                     '-i', fig.fname, '-b:v', mov.bitrate, '-crf 10', mov.fname)  
+  } else {
+    ffmpeg.cmd = paste('ffmpeg', '-y', '-r', 1 / options$interval,
                      '-i', fig.fname, mov.fname)
+  }
+  
   message('executing: ', ffmpeg.cmd)
   system(ffmpeg.cmd, ignore.stdout = TRUE)
 


JasonPunyon:
@Yihui: I'm not sure that code was specifically being tested before...lemme know if you'd like me to fix that.

yihui:
I'm okay not to write a test for this function, so please feel free to ignore the coverage failure.

This PR looks good, but I guess it can take a more general approach: just let users specify whatever command-line arguments they want. For example, we may introduce an option `options$ffmpeg.args`, and users may pass additional arguments like `ffmpeg.args = "-b:v 1M -crf 10"`. That said, I think it makes perfect sense to use reasonable default arguments to produce videos of good quality by default, e.g. we may check if `-b:v` or `-crf` has been specified in `ffmpeg.args` (if not, append good defaults to the command).

JasonPunyon:
I think there's a few reasons why this might be a bit nicer than a general `ffmpeg.args`. 

1. Simpler for knitr users: I think the interface to the parameters is nicer as something like `options$bitrate` as it keeps knitr users at the knitr level of abstraction without having to dive down to reading the ffmpeg manual.

2. Simpler for knitr the project: A general `options$ffmpeg.args` is a much larger promise to make because the arguments on the `ffmpeg` command line are positional. Output args go before the output, input args go before the input, then there's another spot for choosing your video codec, etc, etc. That's just what I know from working with ffmpeg for a couple days. Keeping up with how all those change over time, providing `options$ffmpeg.args` properties for each position in the command line, and funneling them through to the right place reliably seems like a big ask of knitr.

yihui:
Thanks! Good points. Okay, I tend to agree with you. Let me think more about it. My current little concern is the arbitrary expansion of chunk options: I was hoping I could "namespace" them (in particular, the animation/video options). I regret a little that I used the option names `aniopts` and `interval`. Given that we are about to introduce another option `bitrate`, I tend to give it a prefix, e.g. `ffmpeg.bitrate`.

JasonPunyon:
I like the idea of consolidating all these under one roof, it might be good to get the file extension under that too instead of requiring people to set the ffmpeg hook with the extension they want...

yihui:
Absolutely.

yihui:
I merged it, but I think I will rename the option to something like `ffmpeg.bitrate` instead. Thanks!

Issue #1085, opened 2015-08-06
asis chunk in math mode adds an extra newline


jpritikin:
knitr 1.10.5

Here is the input Rnw:

```
\begin{align}
<<results="asis">>=
drift <- matrix(1:4,2,2)
print(xtable(drift, align=rep('',1+ncol(drift))), floating=FALSE, tabular.environment="bmatrix", 
  hline.after=NULL, include.rownames=TRUE, include.colnames=TRUE,
  sanitize.rownames.function=function(x)x, sanitize.colnames.function=function(x)x)
@
\end{align}
```

Here is the output latex:

```
\begin{align}
% latex table generated in R 3.2.1 by xtable 1.7-4 package
% Thu Aug  6 09:40:59 2015
\begin{bmatrix}{}
  & 1 & 2 \\ 
 1 &   1 &   3 \\ 
  2 &   2 &   4 \\ 
  \end{bmatrix}

\end{align}
```

Where does the newline before \end{align} come from? This newline confuses latex because it thinks that it is a new paragraph. I've look at the output from print.xtable and there is only 1 newline at the end. Where does the 2nd newline come from?

yihui:
Sorry I don't have time to track down this newline at the moment. One thing you could do is to post-process the output from xtable, e.g.

```latex
\documentclass{article}
\usepackage{amsmath}
\begin{document}

\begin{align}
<<results="asis",echo=FALSE>>=
library(xtable)
drift <- matrix(1:4,2,2)

print.xtable <- function(x, ...) {
  out <- capture.output(xtable::print.xtable(x, ...))
  cat(gsub('\\s+$', '', paste(out, collapse = '\n')))
}

print(xtable(drift, align=rep('',1+ncol(drift))), floating=FALSE, tabular.environment="bmatrix",
  hline.after=NULL, include.rownames=TRUE, include.colnames=TRUE,
  sanitize.rownames.function=function(x)x, sanitize.colnames.function=function(x)x)
@
\end{align}

\end{document}
```

jpritikin:
Good idea. Thanks.

jpritikin:
Wait, you don't think this should eventually be fixed?

yihui:
It will be nice to fix this at some point, but given my limited time and the fact that this is the first report through the four years, all I can say at the moment is pull requests are welcome :)

PR #1092, opened 2015-08-17
Do not add extra newline at chunk boundaries

jpritikin:
This will fix issue #1085.


yihui:
Thanks, but it is not that simple :) See https://travis-ci.org/yihui/knitr/jobs/75886766 Travis CI failed because the output of the examples in the repo https://github.com/yihui/knitr-examples has changed.

jpritikin:
How do I run the integration test? I get:

````
knitr-examples: No such file or directory.
````

Do I need to clone knitr-examples into knitr like a submodule?

yihui:
Basically clone https://github.com/yihui/knitr-examples and run `make deps knit diff` in it. It does not have to be a submodule.

PR #1104, opened 2015-09-16
Better Stata support for OS X

Hemken:
This adds slightly better functioning for Stata under OS X:  I've been working with a users from Duke University who is a Mac user.  The improvement is that Stata shuts down automatically, rather than waiting for the user to exit manually.


yihui:
Thanks!

PR #1140, opened 2015-12-10
Local chunk options supersede opts_template() options

aaronwolen:
Love the `opts_template()` feature but occasionally want to override a specific template parameter. This PR just alters the order in which params are processed so local chunk options override any redundant options specified within an options template.

For example:

    ```{r setup}
	knitr::opts_template$set(quiet = list(echo=FALSE))
	```

	```{r template, opts.label='quiet'}
	print("This is hidden")
	```

	```{r override, echo=TRUE, opts.label='quiet'}
	print("This is visible")
	```

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -15,15 +15,16 @@ call_block = function(block) {
   # now try eval all options except those in eval.after and their aliases
   af = opts_knit$get('eval.after'); al = opts_knit$get('aliases')
   if (!is.null(al) && !is.null(af)) af = c(af, names(al[af %in% al]))
+
+  # expand parameters defined via template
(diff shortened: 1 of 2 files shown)


aaronwolen:
It looks like the existing installation of knitr is used to run the tests? If so, it failed the test I added to make sure local options supersede template options.

yihui:
No, the tests always use your latest revision of the package. Did you run the tests locally?

aaronwolen:
Yeah, tests passed locally and in a clean docker image. I'll dig into it a little more if you're interested in the proposed change.

yihui:
I think I know what the problems are. I'll fix them myself. Thanks!

PR #1169, opened 2016-02-19
Added ... to kable in knit_print.data.frame

dmenne:
Otherwise ... makes no sense


yihui:
This example is to show how you can change the default printing behavior when a top-level expression returns a visible value that is a data frame (e.g. `mtcars`), so there is actually no way to pass additional arguments to kable(). I mean

    ```{r}
    mtcars
    ```

instead of explicitly calling the `knit_print()` function.

dmenne:
True, but knit_print is also rather smooth here. Less thinking...

PR #1174, opened 2016-03-05
knit2wp environment fix

carloscinelli:
Hi, Yihui,

The function `knit2wp()` did not pass the argument `envir = parent.frame()` to the `knit()` function (like the other `utils-conversion.R` functions do). That was causing problems because the `.rmd` file was being evaluated in the function environment.

So, for instance, if one had a `rm(list = ls())` in the `.rmd` file, it would remove all the function arguments and cause the function to break.

See this minimal example. 

Create a `bug.rmd` file with:

> <div> --- </div>
> title: "minimal"
> author: "Carlos Cinelli"
> date: "5 de março de 2016"
> output: html_document
> <div>---</div>
> 
> <div>```{r rm}</div>
> rm(list = ls())
> <div>```</div>
> 

Then try to `knit2wp()` the `bug.rmd` file. The function does not find the `encoding` argument.

I would write a test for this, but since it requires to provide a WordPress user and password, I could not think of a simple way to do it. 

Hope this helps, thanks!

### DIFF
--- a/R/utils-conversion.R
+++ b/R/utils-conversion.R
@@ -140,11 +140,11 @@ knit2html = function(input, output = NULL, ..., envir = parent.frame(), text = N
 #'   (especially when using Windows).
 #' @examples # see the reference
 knit2wp = function(
-  input, title = 'A post from knitr', ..., shortcode = FALSE,
+  input, title = 'A post from knitr', ..., envir = parent.frame(), shortcode = FALSE,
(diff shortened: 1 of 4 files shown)


yihui:
That is straightforward enough, and I don't need a test. Thanks!

carloscinelli:
Glad it helped!

PR #1179, opened 2016-03-08
Fix syntax highlighting of "NULL".

ClaudiusL:
Minor fix (and a first test of how to file pull requests).


yihui:
Good catch. My brain was poisoned by the R documentation syntax.

PR #1193, opened 2016-04-10
add js and css engines

jjallaire:
The engines are quite straightforward, the just surround their content with `<script>` and `<style>` tags respectively, and print no output when not in an HTML document.

This is of course already possible by direct embedding of `<script>` and `<style>` tags, the main motivation for making them engines is to enable use of the RStudio js and css modes for editing, to better demark them as executable code within documents, and enable use of both of the above with chunk options like eval = FALSE, etc.

@jcheng5


jjallaire:
See related PR in RStudio IDE: https://github.com/rstudio/rstudio/pull/681

PR #1239, opened 2016-07-11
Enable caching of Rcpp engine chunks

jjallaire:
This PR aims to enable support for caching of Rcpp engine chunks. Once completed this will resolve https://github.com/rstudio/bookdown/issues/113.

This also requires changes in Rcpp which can be found here: https://github.com/RcppCore/Rcpp/pull/504. To install this branch:

```r
devtools::install_github("RcppCore/Rcpp", ref = "feature/source-cpp-cache-dir")
```

The way Rcpp does caching of compilations isn't entirely compatible with the default codepath for knitr caching so we'll need some additional customized behavior for the Rcpp engine to get this fully working. 

Cross-session caching for the `sourceCpp` function can now be enabled by passing a `cacheDir` argument to `sourceCpp` (it normally defaults to `tempdir()`). This will cache the expensive part of the compilation (the R CMD SHLIB) however the code still needs to run in order to bind the R function that wraps the C++ function to the location of the (cached) shared library on disk.

So what we want is that for the Rcpp engine the `cache` option is passed through to the engine (so it can invoke `sourceCpp` with the `cacheDir` option) however knitr essentially ignores this option and always runs the chunk.

The more general notion here is that engines should be able to have their own caching strategies. I trawled around the knitr code for a while attempting to make this work but couldn't find a path. @yihui let me know your thoughts.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -67,7 +67,9 @@ call_block = function(block) {
     }
     hash = paste(valid_path(params$cache.path, label), digest::digest(content), sep = '_')
     params$hash = hash
-    if (cache$exists(hash, params$cache.lazy) && isFALSE(params$cache.rebuild)) {
+    if (cache$exists(hash, params$cache.lazy) &&
+        isFALSE(params$cache.rebuild) &&
+        params$engine != 'Rcpp') {
       if (opts_knit$get('verbose')) message('  loading cache from ', hash)
       cache$load(hash, lazy = params$cache.lazy)
       if (!params$include) return('')
--- a/R/engine.R
+++ b/R/engine.R
@@ -157,30 +157,17 @@ eng_Rcpp = function(options) {
   opts = options$engine.opts
 
   # use custom cacheDir for sourceCpp if it's supported
-  cache <- options$cache && (packageVersion("Rcpp") >= "0.12.5.6")
+  cache <- options$cache && (packageVersion("Rcpp") >= "0.12.5.7")
   if (cache) {
     opts$cacheDir <- file.path(opts_chunk$get('cache.path'),
                                paste(options$label, "sourceCpp", sep = "_"))
-    if(!dir.exists(opts$cacheDir))
-      dir.create(opts$cacheDir, recursive = TRUE)
+    opts$cleanupCacheDir = TRUE
   }
 
   if (!is.environment(opts$env)) opts$env = knit_global() # default env is knit_global()
   if (options$eval) {
-    # build the library
     message('Building shared library for Rcpp code chunk...')
-    result <- do.call(getFromNamespace('sourceCpp', 'Rcpp'), c(list(code = code), opts))
-    # if we are using a cache then cleanup old entries
-    if (cache) {
-      cpp_source_path <- normalizePath(result$cppSourcePath)
-      build_directory <- normalizePath(result$buildDirectory)
-      cache_dir <- dirname(build_directory)
-      cache_files <- list.files(cache_dir, pattern = glob2rx("*.cpp"), recursive = FALSE, full.names = TRUE)
-      cache_files <- c(cache_files, list.dirs(cache_dir, recursive = FALSE, full.names = TRUE))
-      cache_files <- normalizePath(cache_files)
-      old_cache_files <- cache_files[!cache_files %in% c(cpp_source_path, build_directory)]
-      unlink(old_cache_files, recursive = TRUE)
-    }
+    do.call(getFromNamespace('sourceCpp', 'Rcpp'), c(list(code = code), opts))
   }
 
   options$engine = 'cpp' # wrap up source code in cpp syntax instead of Rcpp
--- a/R/engine.R
+++ b/R/engine.R
@@ -157,13 +157,26 @@ eng_Rcpp = function(options) {
   opts = options$engine.opts
 
   # use custom cacheDir for sourceCpp if it's supported
-  if (options$cache && (packageVersion("Rcpp") >= "0.12.5.5"))
-    opts$cacheDir <- file.path(opts_chunk$get('cache.path'))
+  cache <- options$cache && (packageVersion("Rcpp") >= "0.12.5.5")
+  if (cache) {
+    opts$cacheDir <- file.path(opts_chunk$get('cache.path'),
+                               paste(options$label, "sourceCpp", sep = "_"))
+    if(!dir.exists(opts$cacheDir))
+      dir.create(opts$cacheDir, recursive = TRUE)
+  }
 
   if (!is.environment(opts$env)) opts$env = knit_global() # default env is knit_global()
   if (options$eval) {
     message('Building shared library for Rcpp code chunk...')
-    do.call(getFromNamespace('sourceCpp', 'Rcpp'), c(list(code = code), opts))
+    result <- do.call(getFromNamespace('sourceCpp', 'Rcpp'), c(list(code = code), opts))
+    print(result)
+    # remove old build directories from cache
+    if (cache) {
+      buildParent <- dirname(result$buildDirectory)
+      buildDirs <- list.dirs(path = buildParent, full.names = TRUE, recursive = FALSE)
+      oldBuildDirs <- buildDirs[buildDirs != result$buildDirectory]
+      unlink(oldBuildDirs, recursive = TRUE)
+    }
   }
 
   options$engine = 'cpp' # wrap up source code in cpp syntax instead of Rcpp
--- a/R/engine.R
+++ b/R/engine.R
@@ -155,6 +155,11 @@ eng_Rcpp = function(options) {
   # engine.opts is a list of arguments to be passed to Rcpp function, e.g.
   # engine.opts=list(plugin='RcppArmadillo')
   opts = options$engine.opts
+
+  # use custom cacheDir for sourceCpp if it's supported
+  if (options$cache && (packageVersion("Rcpp") >= "0.12.5.5"))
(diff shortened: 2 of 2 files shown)


jjallaire:
There's one additional change that I think we should consider alongside this one (and in fact may simplify the implementation of special caching behavior for the Rcpp engine). 

Currently, Rcpp chunks are all compiled as standalone entities (meaning that unlike the way R code is evaluated they can't refer to functions/types/etc. defined in previous chunks). This makes writing about Rcpp more difficult because you have to include all related code in a single chunk (rather than breaking up your code/implementation into multiple chunks that are explained in turn). 

Within the RcppGallery we have a workaround for this: if you use a cpp file + doxygen to author an article then we concatenate all of the C++ chunks into a single chunk that's executed at the beginning. I believe that this is how the Rcpp engine in knitr should work as well.

So here's everything that's being proposed:

1. Concatenate all Rcpp engine chunks into a single chunk that's executed at the beginning of the document. Allow this to be controlled via a knit option (e.g. `rcpp.combine`) for those that want to rely on the previous behavior.

2. Allow Rcpp to handle it's caching internally (by inspecting `options$cache` and providing the knitr cache path as the `cacheDir` when invoking `sourceCpp`).

@yihui I'll probably need to rely on your to sort out the best way to do this cleanly.

@kevinushey @eddelbuettel @thirdwing

yihui:
1. This is simple. We actually talked about it once before, although I don't remember if you were in the conversation. There are two ways to achieve it:
    1. Extract all chunk labels of Rcpp chunks and combine these chunks using the `ref.label` option, e.g. ```` ```{Rcpp, ref.label=knitr::all_labels(engine = 'Rcpp')[-1]} ````;
    1. Or selectively combine certain chunks using either `ref.label = c('foo', 'bar', ...)`, or inline chunk references `<<>>`: http://yihui.name/knitr/demo/reference/
1. I see what you mean, and I need to think more about it.

BTW, I'm not sure if it is the expected behavior, but it seems a new set of files/dirs will be created under `cacheDir` every time I run `sourceCpp()` with a fixed `cacheDir`. I was testing on OS X.

jjallaire:
Okay, sounds like implementing the collection of Rcpp chunks will be relatively straightforward. I personally think that we should make this a default behavior for the Rcpp engine, but if you think that knitr shouldn't be doing something that intrusive / opinionated (I can definitely see that viewpoint) then perhaps it could be an rmarkdown feature. Let me know what you think.

In terms of the new files being created, I probably need to scope the cacheDir by the chunk label. Is the chunk label always available to the engine via options and is it always unique within the cache directory?

jjallaire:
One more thought: I do actually think that having knitr collect all the Rcpp chunks is the right default behavior, as this will make Rcpp behave more like R (you can build up a set of types/functions then use them below). It will also dramatically improve performance for documents that have several Rcpp chunks. I think if we provide a way to turn the behavior off then we'd be in the clear.

jjallaire:
I've made a change to eliminate the problem of old build directories -- we now create a per-chunk label cache directory and remove old build directories whenever we get a new build. Note that you'll need to re-install from the Rcpp dev branch in order to see this working:

```r
devtools::install_github("RcppCore/Rcpp", ref = "feature/source-cpp-cache-dir")
```

yihui:
Actually I didn't mean old build directories on the knitr side. I was testing this:

```r
Rcpp::sourceCpp(code='
  #include <Rcpp.h>

  // [[Rcpp::export]]
  int fibonacci(const int x) {
    if (x == 0) return(0);
    if (x == 1) return(1);
    return (fibonacci(x - 1)) + fibonacci(x - 2);
  }', cacheDir = '~/tmp'
)
```

If I run it three times, I get three sets of cache files/dirs:

```bash
$ ls ~/tmp/sourceCpp-0.12.5.5-x86_64-apple-darwin15.5.0-70800/
cache.rds		file2d4e74bdf7e5.cpp	sourcecpp_2d4e7f4c89be
file2d4e329c23ee.cpp	sourcecpp_2d4e28faf837	token.rds
file2d4e6e707b93.cpp	sourcecpp_2d4e63de1743
```

jjallaire:
Ah, okay! I've already taken care of the .cpp files in the Rcpp PR, however
the directories still pile up. Perhaps I can move the logic I have in knitr
over to the Rcpp side.

Note that these have historically been allowed to pile up b/c they were in
the session temp directory so ultimately always cleaned up. A stable cache
directory changes the game though.

jjallaire:
Okay, I found and fixed the problem (it was a bug in my cache lookup
logic). Now re-executing sourceCpp over and over again with the same code=
parameter won't create any extra files.

If the code changes we still get a bit of a pilup (as there's no way for us
to determine which old "code" is being replaced by new "code" without some
sort of context identifier (like we have w/ knitr chunk labels).

jjallaire:
I've now enhanced the Rcpp engine to cleanup old cache directories within a chunk label's cache directory.

Here is my current solution to allowing the Rcpp engine to handle caching internally: https://github.com/jjallaire/knitr/commit/a5f49a46234051629544efa4376e9cb69bd460d2

Note that this is correct for both newer versions of Rcpp that use `sourceCpp::cacheDir` as well as older versions of Rcpp for which the cached objects are invalid across sessions. Does this approach seem reasonable to you?

R/engine.R
```diff
@@ -155,10 +155,32 @@ eng_Rcpp = function(options) {
...
+    if (cache) {
+      cpp_source_path <- normalizePath(result$cppSourcePath)
+      build_directory <- normalizePath(result$buildDirectory)
+      cache_dir <- dirname(build_directory)
+      cache_files <- list.files(cache_dir, pattern = glob2rx("*.cpp"), recursive = FALSE, full.names = TRUE)
+      cache_files <- c(cache_files, list.dirs(cache_dir, recursive = FALSE, full.names = TRUE))
+      cache_files <- normalizePath(cache_files)
+      old_cache_files <- cache_files[!cache_files %in% c(cpp_source_path, build_directory)]
+      unlink(old_cache_files, recursive = TRUE)
+    }
```
yihui: I think it makes more sense to move this cleaning logic to the Rcpp package, and knitr can just call the cleaning function here, since Rcpp knows more about the internals of the cache.

R/block.R
```diff
@@ -67,7 +67,9 @@ call_block = function(block) {
     }
     hash = paste(valid_path(params$cache.path, label), digest::digest(content), sep = '_')
     params$hash = hash
-    if (cache$exists(hash, params$cache.lazy) && isFALSE(params$cache.rebuild)) {
+    if (cache$exists(hash, params$cache.lazy) &&
+        isFALSE(params$cache.rebuild) &&
+        params$engine != 'Rcpp') {
```
yihui: I was thinking of some sort of callback functions to be passed to `cache$load()`, but I could not come up with a clean solution, so let's just use your approach here to make Rcpp an exception in knitr's caching system.

R/engine.R
jjallaire: Okay, I'll move this to Rcpp now.

jjallaire: Okay, done with these two commits:

https://github.com/RcppCore/Rcpp/pull/504/commits/a825e48102d311b6ecd78e35829428585c949a5b
https://github.com/yihui/knitr/pull/1239/commits/3af0974124dc129704e1c67804bc5eb65a7d158f

jjallaire:
Now that I've taken a closer look at the mechanics of rolling up Rcpp chunks into a single compilation I agree that we should leave this as a manual mode thing (as the user really needs to understand what they are doing, how it interacts with the cache, etc.). Once this is merged I'll write up an article on authoring long-form Rcpp articles/chapters that walks people through options and best practices.

R/block.R
jjallaire: I think this is okay, and given how sourceCpp works we should have already been exempting the Rcpp from the caching system (since the cached R objects can't survive across sessions w/o the dedicated cache we've now added).

yihui:
Okay, merged. Do you want me to provide a wrapper function for `knitr::all_labels(engine = 'Rcpp')`?

jjallaire:
Yeah, a wrapper function would be great!

jjallaire:
Here's what I've sorted out so far as to what the usage idiom should be and how the wrapper function would be integrated:

    ```{Rcpp, cache=TRUE, ref.label=knitr::all_rcpp_chunks()}
    ```

    ```{Rcpp, eval=FALSE}
    #include <Rcpp.h>
    ```

    ```{Rcpp, eval=FALSE}
    // [[Rcpp::export]]
    double timesTwo(double x) {
       return x * 2;
    }
    ```

Does that look right to you?

Were you thinking of something like this or something even higher level, e.g.

    ```{r}
    knitr::evaluate_rcpp_chunks(cache = TRUE)
    ```

    ```{Rcpp}
    #include <Rcpp.h>
    ```

    ```{Rcpp}
    // [[Rcpp::export]]
    double timesTwo(double x) {
       return x * 2;
    }
    ```

Where a side effect of `knitr::evaluate_rcpp_chunks` would be setting `eval = FALSE` on all of the Rcpp chunks in the document?

The second form is higher level but actually obscures more of what is actually going on. It also has the subtle problem of needing to include a `cache = FALSE` on the chunk that calls `knitr::evaluate_rcpp_chunks` to ensure that the Rcpp caching logic always gets called.

For those reasons I think I'd prefer the first form even though it's a bit more syntax (you may have already concluded the same thing!)

yihui:
Yes I prefer the first form. I'll add the wrapper function to extract Rcpp chunk labels.

PR #1240, opened 2016-07-12
added string-align

r2evans:
Concept: use a single multi-character string to align columns as an alternative to a vector of single chars. These two uses are identical:

```r
kable(head(iris), align = c("c","l","r","c","r"))
kable(head(iris), align = "clrcr")
```


R/table.R
```diff
@@ -88,6 +91,9 @@ kable = function(
   } else 'pandoc'
   if (is.function(format)) format = format()
 
+  # expand align if applicable
+  if (length(align) == 1L && nchar(align) > 1L) align <- strsplit(align, '')[[1]]
```
yihui: Everything looks good. A few things to do before I merge:

1. `<-` to `=` here.
2. You can either leave out the changes in .Rd, or install **Rd2roxygen** and use the Build & Reload button in RStudio to generate the Rd.
3. A news item in NEWS.md.

Thanks!

PR #1241, opened 2016-07-12
integrate jcheng5/knitsql as knitr engine

javierluraschi:
This PR adds a SQL engine for knitr to allow chunks based on DBI that look like:

```
``{r}
library(DBI)
conn <- dbConnect(RSQLite::SQLite(), dbname = "sql.sqlite")
number <- 42
``

``{sql output, conn = conn}
SELECT ?number
``

``{r}
output
``
##   42
## 1 42
```

This change comes from integrating https://github.com/jcheng5/knitsql adding a sample `.Rmd` file and removing the `yaml` configuration feature.

+ @jjallaire @jcheng5


R/engine.R
```diff
@@ -366,6 +366,55 @@ eng_js = eng_html_asset('<script type="text/javascript">', '</script>')
 # include css in a style tag (ignore if not html output)
 eng_css = eng_html_asset('<style type="text/css">', '</style>')
 
+# sql engine
+eng_sql = function(options) {
+  # Return char vector of sql interpolation param names
+  varnames_from_sql <- function(conn, sql) {
+    varPos <- DBI::sqlParseVariables(conn, sql)
```
jjallaire: I think we are going to want the **DBI** package to be a Suggests dependencies, which implies that we should preface our calls to DBI with a `loadable` check (see https://github.com/yihui/knitr/blob/07b79a9ff1a3126a0ecf5405946463c1f4b1ceee/R/utils.R#L795).

demo/sql.Rmd
```diff
@@ -0,0 +1,43 @@
...
+  html_document: default
+  html_notebook: default
+---
+
+```{r}
+library(DBI)
+conn <- dbConnect(RSQLite::SQLite(), dbname = "sql.sqlite")
+```
+
+```{sql, conn=conn}
```
jjallaire: I might favor having this attribute be `connection` rather than `conn` (just a bit less obscure).

demo/sql.Rmd
```diff
@@ -0,0 +1,43 @@
...
+CREATE TABLE packages (id INTEGER, name TEXT)
+```
+
+```{sql, conn=conn}
+INSERT INTO packages VALUES (1, 'readr'), (2, 'readxl'), (3, 'haven')
+```
+
+```{sql allPackages, conn=conn}
```
jjallaire: I'm not sure about the convention of assigning a variable based on the chunk label. This might be surprising to users and there are other legit uses for chunk labels (e.g. dependency references, including chunks in other parts of the document, navigation in the IDE, etc.) which might make their natural name different than the R variable you want to assign.

The `stan` engine uses `x` for this, but that might be a bit obscure as well. Some other ideas:

- `output`
- `var`
- `output.var`
- `x`

@yihui Let us know what you think.

yihui: I agree that using the chunk label as the variable name does not sound like a good idea. A chunk option `output.var` sounds better:

    ```{sql conn=conn, output.var='foo'}

R/engine.R
yihui: Putting **DBI** in Suggests is enough. If it is not installed or loadable, R will just signal an error here. I use `loadable()` when the package is not absolutely needed, which is not true here.

R/engine.R
```diff
@@ -366,6 +366,55 @@ eng_js = eng_html_asset('<script type="text/javascript">', '</script>')
...
+    do.call(DBI::sqlInterpolate, c(list(conn, sql), args))
+  }
+
+  conn <- options$conn
+  varname <- options$label
+  sql <- options$code
+
+  query <- interpolate_from_env(conn, sql)
+  result <- DBI::dbGetQuery(conn, query)
+  output <- capture.output(print(result))
```
yihui: `output` is not used -- should be passed to `engine_output()`

R/engine.R
javierluraschi: Updated to only use `suggests`

demo/sql.Rmd
javierluraschi: changed.

demo/sql.Rmd
javierluraschi: using `output.var`.

R/engine.R
javierluraschi: thanks, passed now when `result` is not null.

javierluraschi:
Here is the output with the pr feedback changes:

screen shot 2016-07-12 at 11 55 53 pm

jjallaire:
Another thought aimed at preventing performance problems when exploring large datasets: if there is no `ouptut.var` provided then I think we should automatically append a `LIMIT` clause to `SELECT` statements. This is also the default behavior of e.g. dplyr which will fully realize the dataset only for `collect()` but only show 10 records for `print()`. We could add a `max.display` or `max.print` chunk option to control the number of records printed.

R/engine.R
```diff
@@ -366,6 +366,55 @@ eng_js = eng_html_asset('<script type="text/javascript">', '</script>')
...
+    names <- unique(varnames_from_sql(conn, sql))
+    names_missing <- names[!mexists(names, env, inherits)]
+    if (length(names_missing) > 0) {
+      stop("Object(s) not found: ",
+           paste('"', names_missing, '"', collapse = ", "))
+    }
+
+    args <- if (length(names) > 0) {
+      setNames(
+        mget(names, inherits = inherits),
```
yihui: I guess we also need `envir = env` for `mget()`.

jjallaire:
A related suggestion: @yihui What would you think about importing the **tibble** (https://cran.r-project.org/web/packages/tibble/index.html) package for printing data frames within knitr? I guess you could also just check whether it is installed and use it in that case. Alternatively we could do the same thing in rmarkdown.

yihui:
I think the `LIMIT` suggestion makes sense, although I'm not sure it can be done reliably. 

I definitely prefer tibble's printing behavior over R's default, so yes, `if (loadable('tibble'))  print(tibble::as_tibble(result)) else print(result)` will be nice.

jjallaire:
Another more robust way to implement the LIMIT suggestions is to use
dbSendQuery/dbFetch(n=10)/dbClearResult rather than dbGetQuery in cases
where no variable is being assigned.

Glad you like the tibble idea, Javier let's do it as Yihui suggests (and
add a Suggests dependency for tibble)

yihui:
Oh yeah, passing the limit as a function argument should be more robust than manipulating the SQL code directly.

jjallaire:
@javierluraschi Could you merge master into this PR (I want to provide instructions for installing from the PR branch and I want the branch to include the recent Rcpp work we did)

javierluraschi:
@jjallaire merged to master + added tibble dependency.

DESCRIPTION
```diff
@@ -101,7 +101,8 @@ Suggests:
     jpeg,
     XML,
     RCurl,
-    DBI
+    DBI,
+    tibble
```
jjallaire: You also need the bit that Yihui provided to use tibble to do the printing when it's available (i.e. `loadable`).

javierluraschi: Thanks updated.

jjallaire:
A couple more changes I think we still need here:

1. Add the `envir = env` to the mget call as suggested by @yihui 

2. Add the behavior of throttling the number of records retrieved when we are only displaying the data (I'd say 10 records is okay but I could live with 15 or 20). This should be done via dbSendQuery/dbFetch(n=10)/dbClearResult.

3. Add a `max.display` option which lets the user override the default number of displayed records.

yihui:
Yeah these are simple enough, so I'm adding them now.

yihui:
All done now. For 2, the chunk option is `limit`, which defaults to 10. Users can set it to -1 to fetch all records.

jjallaire:
I have one more idea that I may want to explore for data display (will send a PR tomorrow or Monday but just wanted to put it out there for discussion). If we know we are in rmarkdown we could display the data using `kable`. Furthermore, the html_document format in rmarkdown could attach stickytableheaders to the kable output (as I do in flexdashboard: https://github.com/rstudio/flexdashboard/blob/master/inst/www/stickytableheaders/jquery.stickytableheaders.js). These two things together would make it pretty easy to display more like 1000 records and not have it overwhelm the document.

yihui:
Coincidently I also thought for a while about using `kable()`! 😄

PR #1242, opened 2016-07-14
various improvements to sql engine

jjallaire:
A number of related improvements here:

1. Never display records when capturing to a variable (this is equivalent to an R assignment to a data frame which is invisible)

2. Use single `max.print` option (replaces `limit` and `max.display`) whose default is drawn from a knitr option (`sql.max.print`) that in turn defaults to 10.

3. Check `DBI::dbHasCompleted` before calling `DBI::dbFetch` to prevent warnings for queries that don't return records.

4. Print table using `kable` when emiting markdown output. For HTML output also add a div around the table with class "knitsql-table" to make it easier for page renderers to provide custom styling/behaviors (e.g. sticky table headers for larger numbers of records).

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -410,27 +410,70 @@ eng_sql = function(options) {
     do.call(DBI::sqlInterpolate, c(list(conn, sql), args))
   }
 
+  # extract options
   conn = options$connection
   varname = options$output.var
-  limit = options$limit %n% 10  # fetch 10 records by default if varname not provided
-  if (is.na(limit)) limit = -1
-  max.display = options$max.display
+  max.print <- options$max.print %n% (opts_knit$get('sql.max.print') %n% 10)
+  if (is.na(max.print) || is.null(max.print))
+    max.print = -1
   sql = options$code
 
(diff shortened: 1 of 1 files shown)


R/engine.R
```diff
@@ -410,27 +410,70 @@ eng_sql = function(options) {
...
   }
 
+  # extract options
   conn = options$connection
   varname = options$output.var
-  limit = options$limit %n% 10  # fetch 10 records by default if varname not provided
-  if (is.na(limit)) limit = -1
-  max.display = options$max.display
+  max.print <- options$max.print %n% (opts_knit$get('sql.max.print') %n% 10)
+  if (is.na(max.print) || is.null(max.print))
```
yihui: It seems `max.print` can never be `NULL` here.

yihui:
Okay, just forget about my previous comment about `limit` in 20688190c1420819347851db08fa5edc7fbcd541. I think what you did here makes perfect sense. I can merge it if you don't have anything else to add.

jjallaire:
Okay, great. No other work pending here so you can merge as is. Thanks!

PR #1243, opened 2016-07-14
eng_sql: add caption to kable html output

jjallaire:
Not sure if this is the right way to go about adding a caption to a kable table but here's my attempt.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -412,6 +412,9 @@ eng_sql = function(options) {
 
   # extract options
   conn = options$connection
+  if (is.null(conn))
+    stop("The 'connection' option (DBI connection) is required for sql chunks.",
+         call. = FALSE)
   varname = options$output.var
   max.print = options$max.print %n% (opts_knit$get('sql.max.print') %n% 10)
   if (is.na(max.print) || is.null(max.print))
--- a/R/engine.R
+++ b/R/engine.R
@@ -442,23 +442,28 @@ eng_sql = function(options) {
       options$results = 'asis'
 
       # wrap html output in a div so special styling can be applied
-      if (is_html_output()) cat('<div class="knitsql-table">\n')
+      if (is_html_output())
+        cat('<div class="knitsql-table">\n')
 
-      # print using kable
-      print(kable(data))
-
-      # terminate div and include caption
-      if (is_html_output()) {
-        # determine records caption
+      # determine records caption
+      caption = options$fig.cap
+      if (is.null(caption) || is.na(caption))
+        caption = NULL
+      if (is.null(caption)) {
         rows = nrow(data)
         rows_formatted <- formatC(rows, format="d", big.mark=',')
-        records <- if (max.print == -1 || rows < max.print)
+        caption <- if (max.print == -1 || rows < max.print)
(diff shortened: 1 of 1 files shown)


yihui:
We can just pass a caption to `kable(data, caption = ...)`. By default, use the caption you composed (n records), but users should be able to either suppress the caption (`NA`), or provide their own captions.

jjallaire:
Okay, updated to pass `caption` directly to kable.

R/engine.R
```diff
@@ -442,13 +442,28 @@ eng_sql = function(options) {
...
+        rows_formatted <- formatC(rows, format="d", big.mark=',')
+        caption <- if (max.print == -1 || rows < max.print)
+          paste(rows_formatted, "records")
+        else
+          paste("Displaying records 1 -", rows_formatted)
+      }
 
       # print using kable
-      print(kable(data))
+      print(kable(data, caption = caption, align = "l"))
```
yihui: Is there a reason for left alignment?

jjallaire:
What I really want to do is ensure that the first column is never right aligned (I don't care so much about the other columns). The issue with the first column is that for database queries it's very often the id field (numeric) which then ends up with right alignment by default. This in turn looks very strange in a table as the contents of the first column don't align with the left edge of the column.

So I guess really what I want is to use the default alignment behavior but then fixup the first column. I can easily do this via JS or CSS on the rmarkdown side but I wanted to do something that gave the same benefit to other renderings (e.g. LaTex, Word, etc.).

Try this demo with and without the align option to see what I am talking about:

    ```{r message=FALSE}
    library(DBI)
    db <- dbConnect(RSQLite::SQLite(), dbname = ":memory:")
    ```

    ```{sql, connection=db}
    CREATE TABLE packages (id INTEGER, name TEXT)
    ```

    ```{sql, connection=db}
    INSERT INTO packages VALUES (1, 'readr'), (2, 'readxl'), (3, 'haven'), (4, 'readr'), (5, 'readxl')
    ```

    ```{sql connection=db, max.print = 5}
    SELECT * FROM packages
    ```

jjallaire:
I guess I could also just do an as.character on the first column before calling kable

jjallaire:
Okay, I dropped the align="l" and instead am just converting the first column to character to force left alignment. Let me know if you think this is a reasonable approach.

yihui:
It makes sense to left align the first column is it is the ID column, but I wonder if users doing data analysis will select the ID column from the database at all (I think the numeric ID column is rarely useful unless they want to join the data frame with other data frames later). The small 2-column example is kind of an extreme case: typically I don't think users only want to select 2 columns (and one of them is ID). I think the weird appearance of this table is mostly due to the 100% width showing too few columns.

yihui:
It will also be strange if the first column is not incremental IDs but a numeric data column. I think a more robust approach is to actually check if it is an incremental column (e.g. `all(diff() == 1)`). That should significantly reduce the false positive rate.

jjallaire:
Okay, I think you are correct on all counts (my example is a bit
contrived). I just updated the PR to only do as.character for all(diff() ==
1

PR #1257, opened 2016-07-28
add support for custom sql printing using sql.print option

javierluraschi:
Allow `eng_sql` to support custom print methods through `sql.print` options. Related to https://github.com/rstudio/rmarkdown/pull/751

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -439,8 +439,16 @@ eng_sql = function(options) {
     # apply max.print to data
     display_data = if (max.print == -1) data else head(data, n = max.print)
 
+    # get custom sql print function
+    sql.print = opts_knit$get('sql.print')
+
     # use kable for markdown
-    if (out_format('markdown')) {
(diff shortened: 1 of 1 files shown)


R/engine.R
```diff
@@ -439,8 +439,16 @@ eng_sql = function(options) {
     # apply max.print to data
     display_data = if (max.print == -1) data else head(data, n = max.print)
 
+    # get custom sql print function
+    sql.print = opts_knit$get('sql.print')
+
     # use kable for markdown
-    if (out_format('markdown')) {
+    if (out_format('markdown') && !is.null(sql.print)) {
```
jjallaire: We probably don't need to check for `out_format('markdown')` in this conditional since whomever set `sql.print` is responsible for knowing the final output format and ensuring that `sql.print` does the right thing.

javierluraschi:
Updated.

PR #1261, opened 2016-08-03
For wrapped code, set `tidy=TRUE`

JoshOBrien:
This now needs to be done explicitly, either where I've made the edit or above that in `opt_chunk$set()`  (which will result in slightly different output for the pdf's first listing).


yihui:
Thanks!

Issue #1265, opened 2016-08-08
Allow numeric values for `fig.keep` to index plots


ClaudiusL:
This is a feature request based on [this question on SO](http://stackoverflow.com/q/38798078/2706569).

So far, `fig.keep` accepts only `'high'`, `'none'`, `'first'`, `'last'` (and, implicitly, `'all'`) . Would it be possible to extend the functionality in a way similar to `echo`, such that `fig.keep = c(1, 3)` keeps the first and the third plot?

One potential drawback is that `fig.keep = c(1, 3)` cannot be combined with `fig.kepp = 'high'`. Therefore, the indices would implicitly always refer to low-level plots. (The same already applies to `fig.keep = 'first'`).

I think the implementation could be quite simple: In [block.R](https://github.com/yihui/knitr/blob/5235ec08ed9f4728eb3239432f1804e9b25d7c02/R/block.R#L222), line 222, an additional `if` clause could be added:

```
if (keep %in% c('first', 'last')) {
  res = res[-(if (keep == 'last') head else tail)(which(figs), -1L)]
} else {
   if (is.numeric(keep)) { #new 
     res = res[which(figs)[keep]]
    }
  # merge low-level plotting changes
  if (keep == 'high') res = merge_low_plot(res, figs)
}
```

However, I am not entirely sure about the contents of `res` and `figs`, so my suggested code might be wrong.

yihui:
That looks okay to me. Could you submit a pull request? Thanks!

ClaudiusL:
I also proposed an update for http://yihui.name/knitr/options/. Should this be a separate pull request? (Not sure about the correct workflow for this.)

ClaudiusL:
Ok. Slowly I'm getting used to how this works … ;)

PR #1266, opened 2016-08-09
Accept numeric indices for fig.keep (#1265).

ClaudiusL:
As discussed in #1265. However, I had to create a new variable for the indices and manipulate `keep` to remain of length 1. Otherwise, passing a vector to `fig.keep` would crash the comparisons accessing `keep`.


yihui:
Three years later, I discovered that this was buggy, and just fixed it via 53b968c5e5b0547c17508bc5d7c5de0e5ac52046. Just FYI and don't worry :)

PR #1282, opened 2016-08-26
Added engine for Fortran95

stefanedwards:
The default behaviour for the Fortran engine is to produce a temporary xxx.f file which causes SHLIB to interpret it as Fortran 77 (https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-F95-code).  Fortran 77 is different from Fortran 90/95.
I will not divulge into the problems of using Fortran 95 (as indicated in the link), I just want to compile Fortran 95 code; the SHLIB mechanism apparently switches to a Fortran 95 compiler (if available) when the file extension is .f95.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -134,7 +134,7 @@ eng_interpreted = function(options) {
 
 ## C and Fortran (via R CMD SHLIB)
 eng_shlib = function(options) {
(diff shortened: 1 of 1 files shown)


yihui:
Thanks!

Issue #1305, opened 2016-10-21
For results='asis' the knitrout environment and the fontsize settings are missing.


kornl:
Is it intentional that for results='asis' the knitrout environment and the fontsize settings are missing?

For example see the following document, were the code of the second chunk is included with option size="tiny", but for the third chunk this option is not applied for the included R code.

```
\documentclass{report}
\begin{document}

\huge

<<echo=FALSE>>=
if (exists("opts_chunk")) {
  opts_chunk$set(size="tiny")
}
@

<<>>=
cat("\\textbf{Hello World!}")
@

<<results='asis'>>=
cat("\\textbf{Hello World!}")
@

\end{document}
```

Thanks and best regards, Kornel

yihui:
It was intentional but I don't quite remember why: https://github.com/yihui/knitr/blob/74670a18d/R/hooks-latex.R#L173-L174 Perhaps it makes sense not to lose `size` when `results='asis'`. Do you want to submit a pull request?

kornl:
Thanks.

I'm still wondering why the knitrout environment should not be included (should this be added to the documentation?), but the pull request for just respecting the size option is now submitted.

And btw. since the function `.rm.empty.envir` was involved and I was looking at it: Are `\begin` and `\end` mixed up for verbatim environments? (Not that the function is ever called for verbatim environments as far as I can see - so it doesn't matter that much in the moment):

```r
# rm empty kframe and verbatim environments
.rm.empty.envir = function(x) {
  x = gsub('\\\\begin\\{(kframe)\\}\\s*\\\\end\\{\\1\\}', '', x)
  gsub('\\\\end\\{(verbatim|alltt)\\}\\s*\\\\begin\\{\\1\\}[\n]?', '', x)
}
```

yihui:
I thought a while about it, I think a safer solution is to use a custom chunk hook: http://yihui.name/knitr/hooks/ Here is a quick example (not tested, just to show the idea):

```r
knitr::knit_hooks$set(size = function(before, options) {
  custom = options$results == 'asis' || options$size == 'normalsize'
  if (!custom) return()  # no need to customize
  if (before) {
    paste0('{\\', options$size)
  } else {
    '}'
  }
})
```

All issues in #1307 and #1308 can be solved in this way.

Re: `.rm.empty.envir()`, `\begin` and `\end` are not mixed up. What it does is to merge `verbatim` and `alltt` environments, e.g. from

```latex
\begin{verbatim}
1
[...]

PR #1307, opened 2016-10-24
Respect option `size` for `asis=TRUE`

kornl:
Pull request for issue #1305


kornl:
I guess I have to change https://github.com/yihui/knitr-examples/blob/master/044-size.Rnw too?

kornl:
Okay, perhaps I understand why it was chosen to ignore the size for asis=TRUE. In the current state of the pull request also the result output is affected by the size argument. Perhaps a more intuitive behavior would be to give the code the size from the size option, but ignore it for the result output?

PR #1313, opened 2016-10-28
add note option to insert simple table note for latex format

liuyadong:
add a note option for kable function to insert simple table note using threeparttable package


yihui:
The `note` argument only needs to be added to `kable_latex()`. All other instances can be removed, because of `kable`'s `...` argument.

yihui:
[Thanks](https://yihui.name/en/2017/11/thanks-sergio-oller/) @liuyadong!

PR #1316, opened 2016-11-06
Support eval = FALSE and results = 'hide' in sql engine.

dmurdoch:
See discussion on <http://stackoverflow.com/questions/40444521/can-knitr-suppress-execution-or-output-in-an-sql-chunk>.


yihui:
Thanks! 👍

PR #1330, opened 2016-12-04
Add go engine

hodgesds:
This engine allows for adding go code within a `{go}` block. It will attempt to run `go fmt` first on any code so that it is rendered in the standard go format. Also supports `eval` as long as the block contains an executable block of code.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -539,6 +539,42 @@ eng_sql = function(options) {
   engine_output(options, options$code, output)
 }
 
+# go engine
+eng_go = function(options) {
+  f = tempfile('code', '.', fileext=".go")
+  writeLines(code <- options$code, f)
+  on.exit(unlink(f), add = TRUE)
+
+  fmt_args = sprintf('fmt %s', f)
+  
+  tryCatch(
+    system2("go", fmt_args, stdout = TRUE, stderr = TRUE, env = options$engine.env),
+    error = function(e) {
+      if (!options$error) stop(e)
+      paste('Error in running command', command_string)
+    }
+  )
+  
+  code = readLines(f)
+
+  run_args = sprintf("run %s", f)
+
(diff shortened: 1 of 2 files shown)


yihui:
Thanks! This looks interesting. Could you show a minimal example? I'm not really familiar with `go`.

hodgesds:
Here are a few examples:

First example is just a code block for a simple struct that is unformatted and will be properly formatted using `{go eval=FALSE}`. You can also copy and paste this code into a file and run `go fmt` on it directly (assuming `go` is in your path) to get similar rendered formatting.
```{go eval=FALSE} 
package example

type Person struct{
Name string `json:"name,"`
age int `json:"age,"` // unexported
}
```

Here is a example that actually executes code using `{go eval=TRUE}`:

```{go eval=TRUE} 
package main

func main(){
  println("Hello World")
}

File: R/engine.R [hunk 1]
```diff
@@ -539,6 +539,42 @@ eng_sql = function(options) {
...
+# go engine
+eng_go = function(options) {
+  f = tempfile('code', '.', fileext=".go")
+  writeLines(code <- options$code, f)
+  on.exit(unlink(f), add = TRUE)
+
+  fmt_args = sprintf('fmt %s', f)
+  
+  tryCatch(
+    system2("go", fmt_args),
```
hodgesds: Good catch, will make it consistent with the rest of the changes.

R/engine.R
```diff
@@ -539,6 +539,42 @@ eng_sql = function(options) {
...
+    error = function(e) {
+      if (!options$error) stop(e)
+      paste('Error in running command', command_string)
+    }
+  )
+  
+  code = readLines(f, -1)
```
hodgesds: I think I left this in while going through the `readLines` documentation (haven't used R in awhile so I'm a bit rusty).

Issue #1333, opened 2016-12-08
Feature request: change chunk background color in RMarkdown


AmeliaMN:
As [discussed on StackOverflow](http://stackoverflow.com/questions/41030477/changing-chunk-background-color-in-rmarkdown), I would like to have a certain code chunk highlighted in a different color (e.g. red) to indicate that it is bad practice. If I was using `.Rnw`, I could add the chunk option `background = 'red'` and get what I want, but this does not seem to work in `.Rmd`. 

@jennybc seconded this request, and @ijlyttle has a solution that might turn into a PR!

ijlyttle:
As a first thought, I am thinking to propose a new "standard" chunk option called `class.html`. It would be the end-programmer's responsibility to provide a bit of css to tell the browser what to do with the class.

Q: I am thinking `class.html` to indicate that this may be useful only for HTML, and to try to fit in with the existing naming scheme. Is this the "right" name? My feelings will not be hurt if it is not.

Q: Does this strike the right balance between being useful in the general situation, but not-too-cumbersome for the specific situation that @AmeliaMN wants to solve?

yihui:
Yes, users have to provide the CSS themselves. This can actually be done via the `css` engine in knitr, e.g.

    ```{css}
    .red { background-color: red; }
    ```

For the option name, does `html.class` sound better? I don't know how this could be useful to non-HTML output (otherwise `class.html` will be a better name).

For the second Q, I think this can be useful in general. The only thing that I'm a little concerned about is whether `html.class` should only be applied to the source code chunk, or also to the text output. Maybe there should be another option that controls the latter, e.g. `html.class.output`. I tend to make `html.class.output = html.class` by default, and users can choose to turn it off explicitly by `html.class.output = NA`.

ijlyttle:
I did not know about being able to put CSS directly into knitr! 

`html.class` works for me - I have no idea if this could be useful to non-HTML - you have a much better view of this than I do.

Following your source/output reasoning, how about `html.class.source` and `html.class.output` that could be set independently? Then `html.class` could be used to set both.

For the @AmeliaMN case, you could then set only `html.class.source`.

yihui:
Let me list the pros and cons in my eyes:

- `html.class` + `html.class.output`: if the major use case is to apply the class to both source and output, this combination wins, because only `html.class` is likely to be used, and this name is short. In other words, users opt-out by `html.class.output` (style output unless I tell you not to).

[...]

PR #1335, opened 2016-12-11
adds options class.output and class.source for html documents

ijlyttle:
Addresses #1333

This work is just beginning - I am opening the PR now so that @yihui and @AmeliaMN can pull me back towards the right path should I go the wrong way.

- [x] code
  -  [x] adapt `render_markdown()` in `hooks-md.R`
     - [x] `class.source`
     - [x] `class.output`
     - [x] clean out extraneous comments in `hooks.t.output`
- [x] documentation
  - [x] make PR to [knitr-examples](https://github.com/yihui/knitr-examples)
  - [x] docs (docs/options/index.md)
- [x] add a name to DESCRIPTION/Contributors

If you prefer to keep all the feedback until the whole PR is "done", I'll send a signal when I think I am there.

FWIW, here's a [gist](https://gist.github.com/ijlyttle/000efb922103ff7bd55625443187e774) with the Rmd file I've been using to test things.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -27,6 +27,7 @@ Authors@R: c(
     person("Heewon", "Jeon", role = "ctb"),
     person("Henrik", "Bengtsson", role = "ctb"),
     person("Hiroaki", "Yutani", role = "ctb"),
+    person("Ian", "Lyttle", role = "ctb"),
     person("Jake", "Burkhead", role = "ctb"),
     person("James", "Manton", role = "ctb"),
     person("Jared", "Lander", role = "ctb"),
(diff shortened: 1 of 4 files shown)


yihui:
- Don't worry about the duplicated code. I'll take care of it since it is simple enough.
- The documentation can be added here (as you have found out): https://github.com/yihui/knitr/blob/master/docs/options/index.md No need to add vignettes or Rd.
- No need to add tests under `tests/`. An example in the [knitr-examples](https://github.com/yihui/knitr-examples) repo will be great, though, and that can serve as a test.
- Then add your name to DESCRIPTION and done.

ijlyttle:
Will do - should not be much longer - thanks!

ijlyttle:
Done, methinks (mehopes). Thanks for the navigation through the process, @yihui!

PR #1338, opened 2016-12-17
Added clarifying note about fig.path behavior

khughitt:
Added a note which should help clarify the behavior of the `fig.path` chunk option and should help make it easier to see how to change the figure output directory to a relative location. In particular, it is useful to indicate that 'output/dir' and 'output/dir/' will behave very differently when passed to `fig.path`. (e.g. see http://stackoverflow.com/questions/21582402/knitr-how-to-set-a-figure-path-in-knit2html-without-using-setwd)

### DIFF
--- a/options/index.md
+++ b/options/index.md
@@ -186,7 +186,7 @@ leading white spaces have special meanings in markdown.
 
 ### Plots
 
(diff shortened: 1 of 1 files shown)


yihui:
👍

PR #1357, opened 2017-02-08
Add expand Option to SQL Engine

saurfang:
SQL chunk is an awesome way to showcase SQL queries in syntax highlighted fashion. The support of interpolation is also great to allow further flexibility. While most the time I like showing uninterpolated SQL which is simple and easy to reason, sometimes I wish I can show interpolated SQL instead. This allows me to construct more complex SQL programmatically but show readers the end results.

This PR adds a new option `expand` for SQL engine, which would show interpolated SQL query in the code chunk.

I am open to any suggestion for a better name. Let me know if you can think of a better idea. Thank you!

Putting a space between ` so code block works correctly.

```
` ``{r}
x <- 1
` ``

` ``{sql, connection=db}
SELECT ?x
` ``

` ``{sql, connection=db, expand=TRUE}
SELECT ?x
` ``
```

image

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -428,7 +428,10 @@ is_sql_update_query = function(query) {
 
 # sql engine
 eng_sql = function(options) {
-  if (isFALSE(options$eval)) return(engine_output(options, options$code, ''))
+  # return chunk before interpolation eagerly to avoid connection option check
+  if (isFALSE(options$eval) && !isTRUE(options$sql.show_interpolated)) {
+    return(engine_output(options, options$code, ''))
+  }
 
   # Return char vector of sql interpolation param names
   varnames_from_sql = function(conn, sql) {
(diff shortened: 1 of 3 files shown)


R/engine.R
```diff
@@ -538,8 +538,11 @@ eng_sql = function(options) {
   # assign varname if requested
   if (!is.null(varname)) assign(varname, data, envir = knit_global())
 
+  # reset query to pre-interpolated if not expanding
+  if (!isTRUE(options$expand)) query <- options$code
```
yihui: This will change the default behavior, i.e. previously the default is to expand the query, and now the default is to show raw query (because `options$expand` is `NULL` by default). I'd use

```
if (isFALSE(options$sql.expand))
```

so that users have to set `sql.expand = FALSE` explicitly to present the query that is not interpolated.

saurfang: Correct me if I'm wrong but I think the previous default behavior is to show query **un**interpolated because it is returning `engine_output(options, options$code, output)` and `options$code` is before interpolation.

I am indifferent what should be the default behavior. I have found both equally common in my writing.

saurfang:
I renamed the option to `sql.show_interpolated` for clarity because there is already ` knit_expand` concept and I fear it can cause confusion. In additional I added support to show interpolated SQL even when `eval=FALSE`. This is particularly useful for chunk references as in:

```r

` ``{sql x, connection=db}
SELECT ?x
` ``

` ``{sql x, connection=db, eval=FALSE, sql.show_interpolated=TRUE}
` ``

```

So I can show interpolated SQL at the appendix section of my writing.

R/engine.R
yihui: Ah you are correct. I was wrong. Sorry.

yihui:
Okay, this makes perfect sense to me now. Please add a NEWS item to NEWS.md, your name to DESCRIPTION as a contributor (`ctb`), and I'll merge it. Thanks!

saurfang:
Thanks! Done and commits squashed.

yihui:
👍

PR #1362, opened 2017-02-20
Support sql engine character connection

javierluraschi:
The following document renders a SQL chunk by setting the connection based on a character string, as in, `knitr::opts_chunk$set(connection = "db")`. This is supported in RStudio notebooks but not in `knitr`, this PR matches functionality.

    ```{r}
    ---
    title: "R Notebook"
    output:
      html_document: default
      html_notebook: default
    ---

    ```{r setup}
    library(DBI)
    db <- dbConnect(RSQLite::SQLite(), dbname = "sql.sqlite")
    knitr::opts_chunk$set(connection = "db")
    ```

    ```{sql}
    SELECT 1
    ```

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -464,6 +464,8 @@ eng_sql = function(options) {
 
   # extract options
   conn = options$connection
+  if (is.character(conn))
+    conn <- get(conn, envir = globalenv())
   if (is.null(conn)) stop2(
     "The 'connection' option (DBI connection) is required for sql chunks."
   )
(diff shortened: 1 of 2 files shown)


R/engine.R
```diff
@@ -464,6 +464,8 @@ eng_sql = function(options) {
 
   # extract options
   conn = options$connection
+  if (is.character(conn))
+    conn <- get(conn, envir = globalenv())
```
yihui: The environment should be `knit_global()`. It is not necessarily the global environment, although in most cases it is.

javierluraschi: Makes sense, thank you Yihui.

Issue #1341, opened 2017-01-06
Modify the or add a new vignette engine to build rst vignettes with pandoc instead of rst2pdf?


trevorld:
If desired I can probably write a patch but I'd like to know your implementation preferences (see below), CRAN/hadley has had issues with getting rst2pdf on some of their testing machines and I wonder if they are widely supporting pandoc now (are they?) then perhaps this would be more convenient all around.

1) Should this be done as the default rst vignette engine instead of rst2pdf and hence should I add a new rst2pdf engine for those who want it or keep the existing behavior and add this as a new engine?   Although the former would break reverse compatibility it might be the better default for CRAN and I'm not sure how many people besides me would be affected and I don't care as long as a nice document is produced in the end.

2) Would it be better to use knitr's pandoc utility conversion function or rmarkdown's pandoc utility conversion function to do the rst to html or pdf conversion?  Doing the latter would make the package author's need to suggest rmarkdown for a non-markdown vignette but would reinforce the pandoc dependency.

3) Would it make sense to generalize the knit2html function to handle non markdown files or easier to add a new function just for rst files that would be called when it detects an rst file (like it does now)?

yihui:
I don't like rst2pdf, and I'd be very happy to get rid of it and use Pandoc if possible (someone else introduced it in knitr long time ago). So for 1, I'd say yes; for 2, let's use rmarkdown instead of `knitr::pandoc()` (the latter is also something I don't like); for 3, I don't like `knit2html`, either. I think rmarkdown has done an excellent job, and it should be very rare that users would need `knit2html` at all.

trevorld:
I don't think ``rmarkdown`` has a function to directly support conversion of `rrst` to `html/pdf`.  I looked at ``rmarkdown::render`` and it has a bunch of markdown specific code.  So we'd need to update knitr's existing functions or write new functions that first converts `rrst` to `rst` with ``knitr::knit`` and then calls ``rmarkdown::pandoc_convert`` to convert the ``rst`` to either an ``html`` or ``pdf`` file.

yihui:
Good point!

I probably won't have time for this in the next couple of weeks, but if your pull request looks simple enough, I may be able to merge it soon. Thanks!

PR #1386, opened 2017-04-03
New function to build rst vignettes with pandoc

trevorld:
Should close (#1341). Added one new function ``knit2pandoc`` which is fairly general so a user could theoretically use it to convert from many formats to other formats (i.e. Rhtml to pdf) but selected a default output html so one could also directly use it as an rst vignette builder (theoretically could use it as a vignette builder for other formats as well like asciidoc?).

Didn't seem to break your unit tests and my rst vignette in optparse package seems to build fine with this modified although doesn't seem any of pandoc's pretty code coloring options were enabled.

This update will cause any packages with an Rrst vignette to switch vignette builidng from rst2pdf to pandoc.  Seems a little uglier but pandoc seems a less painful dependency to install for CRAN etc.  Not sure if any users besides me would be affected by such a change.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -114,7 +114,7 @@ URL: http://yihui.name/knitr/
 BugReports: https://github.com/yihui/knitr/issues
 VignetteBuilder: knitr
 SystemRequirements: Package vignettes based on R Markdown v2 or reStructuredText require Pandoc (http://pandoc.org). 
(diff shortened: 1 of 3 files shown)


yihui:
Yeah, to my knowledge, you are the only one who actually uses rst vignettes. During my next release to CRAN, I'll explain that you are aware of this change (from rst2pdf to pandoc, and PDF to HTML). Thanks!

PR #1391, opened 2017-04-12
Gobble longest common whitespace

thisirs:
Hi,

I often write
    % blah <- 1
instead of
    %blah <- 1
in my Rtex files as I find it more readable. The problem is that it inserts a spurious space in the output or in the tangled file. This pull request fixes this by removing the longest common whitespace.


thisirs:
Done! Thanks!

PR #1417, opened 2017-07-20
add knitr::opts_chunk$append()

yonicd:
i extended the defaults idea in my [pkg](https://github.com/metrumresearchgroup/sinew/blob/master/R/defaults.R), to include the option of appending to an object to the the default values if they are a vector, so user can grow options from a minimal default set.

thought it could also be useful here too


DESCRIPTION
```diff
@@ -80,7 +80,8 @@ Authors@R: c(
     person("Viktoras", "Veitas", role = "ctb"),
     person("Weicheng", "Zhu", role = "ctb"),
     person("Wush", "Wu", role = "ctb"),
-    person("Zachary", "Foster", role = "ctb")
+    person("Zachary", "Foster", role = "ctb"),
+    person("Jonathan", "Sidi", role = "ctb")
```
yihui: The names are listed alphabetically.

R/defaults.R
```diff
@@ -22,7 +22,17 @@ new_defaults = function(value = list()) {
   merge = function(values) merge_list(defaults, values)
   restore = function(target = value) defaults <<- target
 
-  list(get = get, set = set, merge = merge, restore = restore)
+append = function(...) {
```
yihui: Indent by 2 spaces.

R/defaults.R
```diff
@@ -22,7 +22,17 @@ new_defaults = function(value = list()) {
...
+    dots = list(...)
+    if (length(dots) == 0) return()
+    if (is.null(names(dots)) && length(dots) == 1 && is.list(dots[[1]]))
+      if (length(dots <- dots[[1]]) == 0) return()
+    dots<-sapply(names(dots),function(x) dots[[x]]<-c(defaults[[x]],dots[[x]]),simplify = FALSE)
```
yihui: I'd just use a for loop:

```r
for (i in names(dots)) dots[[i]] <- c(defaults[[i]], dots[[i]])
```

tests/testit/test-chunkopts.R
```diff
@@ -0,0 +1,8 @@
+library(testit)
+
+knitr::opts_chunk$append(list(fig.path=c('mypath/')))
+
+assert(
+  'knitr::opts_chunk$append() returns appended chunk option',
+  identical(knitr::opts_chunk$get('fig.path'),c('figure/','mypath/'))
+)
```
yihui: `opts_chunk$restore()` afterwards so that the side effect in this test won't affect other tests.

tests/testit/test-chunkopts.R
```diff
@@ -0,0 +1,8 @@
+library(testit)
+
+knitr::opts_chunk$append(list(fig.path=c('mypath/')))
```
yihui: do not need the qualifier `knitr::`, but that is fine

justmytwospence:
How is this feature used? Like this?

```r

opts_chunk$set(cache.extra=Sys.Date())

```{r, cache=TRUE, cache.extra=opts_chunk$append(tools::md5("foo.sql"))}
```

Could warrant a bullet point in the chunk options documentation page.

yonicd:
this feature is meant for fields that have the option for growing like `opts_chunk$get('dev')`.

in your example:

```
opts_chunk$set(cache.extra=Sys.Date())
opts_chunk$get('cache.extra')
# [1] "2018-01-23"
opts_chunk$append(cache.extra=tools::md5sum("foo.sql"))
opts_chunk$get('cache.extra')
#                 foo.sql 
# "2018-01-23"           NA 
```

justmytwospence:
Ah I see. My problem is that local chunk options merge with global chunk options, but this merging doesn't extend to options that are lists. List options are just entirely overridden with no way to add to the list for a particular chunk as in the example with cache.extra. Unless there's a way to do this that I haven't found, maybe I'll open a separate issue.

yihui:
@justmytwospence I think you meant something like this:

```r
opt_append = function(name, ...) {
  list(knitr::opts_chunk$get(name), ...)  # or c(...), or append(...)
}
```

Then

````
```{r, cache=TRUE, cache.extra=opt_append('cache.extra', tools::md5sum("foo.sql"))}
````

There is no existing syntax for extending a chunk option at the moment.

PR #1418, opened 2017-07-20
Sanitise latex captions of kable() and figures

ThierryO:
Using `%` and `_` in captions with kable() or fig.cap were problematic when rendered to pdf. These are now sanitised with `escape_latex()`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -74,6 +74,7 @@ Authors@R: c(
     person("Taiyun", "Wei", role = "ctb"),
     person("Thibaut", "Assus", role = "ctb"),
     person("Thibaut", "Lamadon", role = "ctb"),
+    person("Thierry", "Onkelinx", role = "ctb"),
     person("Thomas", "Leeper", role = "ctb"),
     person("Tom", "Torsney-Weir", role = "ctb"),
     person("Trevor", "Davis", role = "ctb"),
--- a/R/hooks-latex.R
+++ b/R/hooks-latex.R
@@ -90,8 +90,8 @@ hook_plot_tex = function(x, options) {
     switch(a, left = '\\hfill{}\n\n', center = '\n\n}\n\n', right = '\n\n', '')
 
   # figure environment: caption, short caption, label
-  cap = options$fig.cap
-  scap = options$fig.scap
+  cap = escape_latex(options$fig.cap)
+  scap = escape_latex(options$fig.scap)
   fig1 = fig2 = ''
   mcap = fig.num > 1L && options$fig.show == 'asis' && !length(subcap)
   # initialize subfloat strings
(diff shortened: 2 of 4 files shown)


yihui:
The problem is that if I forcibly sanitize them, they may be problematic in other types of output if the source document format is R Markdown, because a character that is special to LaTeX is not necessarily special to another output format.

If your source document format is R Markdown, you may consider the feature of text references in bookdown for maximum portability: https://bookdown.org/yihui/bookdown/markdown-extensions-by-bookdown.html#text-references

ThierryO:
The format of our source files is R Markdown/bookdown. We have been bitten by the examples below when rendering to pdf

    ```{r my-plot, fig.cap = "Estimate and 95% confidence interval}
    my.plot()
    ```

    ```{r my-table}
    kable(mtcars, caption = "Caption with some _ underscore")
    ```

The format is taken into account when sanitising the `kable` caption. That might not be the case for the `fig.cap`.

Could you explain why the Travis build fails? Because `R CMD Check --as-cran` passes on my machine.

ThierryO:
Here is a minimal bookdown example that works with this PR. The Travis errors seem unrelated to me.

    ---
    title: "Caption test"
    author: "Thierry Onkelinx"
    date: "24 juli 2017"
    output: 
      bookdown::pdf_book: default
      bookdown::epub_book: default
      bookdown::html_document2: default
    ---
    
    ```{r load, include=FALSE}
    library(knitr)
    opts_chunk$set(echo = TRUE)
    ```
    
    ```{r pressure, echo=FALSE, fig.cap = "Pressure with percent % and underscore_"}
    plot(pressure)
    ```
    
    ```{r cars}
    kable(mtcars, caption = "mtcars with percent % and underscore_")
    ```

yihui:
Thanks for this, and apologies for the very long delay.

Revisiting this in 2026, the two halves of the PR have diverged:

- **`kable()` captions** — this is now handled. `kable()` escapes LaTeX/HTML special characters in captions by default (via `kable_caption()`), respecting the format and the `escape` argument. So the `table.R` portion of this PR is no longer needed.

- **`fig.cap` / `fig.scap`** — these are still passed through unescaped in `hook_plot_tex()`, so the figure half of this PR does address a real remaining gap for Sweave/`.Rnw` output.

That said, I don't think I can merge the figure change as-is:

1. It has no opt-out. `.Rnw` users can write raw LaTeX in `fig.cap` (e.g. `fig.cap = "$\\alpha$ level"`), and force-escaping would break that with no way to disable it — unlike `kable()`, which has the `escape` argument for exactly this reason.
2. The PR conflicts with the current codebase.

Since the table half is superseded and the figure half needs a different (opt-out-aware) design, I'll close this. If the `fig.cap` escaping is still worth pursuing, it's better done as a fresh, smaller change with an escape toggle. Thanks again for surfacing the problem.

PR #1433, opened 2017-09-17
New imgur_upload with xml2 and httr

cderv:
Hi @yihui, 

Following a _PS_ call in your recent [blog post](https://yihui.name/en/2017/09/knitr-imgur-upload/), I modified `imgur_upload` to use `httr` and `xml2` instead of `RCurl` and `XML`

Note that `RCurl` is also used in `knitr::image_uri2` for `RCurl::base64Encode`. `httr` has a `base64url` but not exported. What should we do to remove `RCurl` suggest dependency to RCurl ? 

`XML` package is used is called `inst/examples/knitr-themes.Rnw`. I think we can remove the _suggests_ dependency but not sure.

One thing: There is a [imguR](https://github.com/cloudyr/imguR) package with a `upload_imgur` function. I do not know if it is better to have a `imgur_upload` function specific to `knitr` or if it may be a good idea to use a dependency toward `imguR` package and import the `imgur_upload`. 

What do you think of all this ? 
Does it help ?

---
This change is [Reviewable](https://reviewable.io/reviews/yihui/knitr/1433)

### DIFF
--- a/R/utils-upload.R
+++ b/R/utils-upload.R
@@ -1,7 +1,7 @@
 #' Upload an image to imgur.com
 #'
-#' This function uses the \pkg{RCurl} package to upload a image to
-#' \url{imgur.com}, and parses the XML response to a list with \pkg{XML} which
+#' This function uses the \pkg{httr} package to upload a image to
+#' \url{imgur.com}, and parses the XML response to a list with \pkg{xml2} which
 #' contains information about the image in the Imgur website.
 #'
 #' When the output format from \code{\link{knit}()} is HTML or Markdown, this
@@ -36,12 +36,20 @@
 #' }
 imgur_upload = function(file, key = '9f3460e67f308f6') {
   if (!is.character(key)) stop('The Imgur API Key must be a character string!')
-  res = RCurl::postForm(
-    'https://api.imgur.com/3/image.xml', image = RCurl::fileUpload(file),
-    .opts = RCurl::curlOptions(httpheader = c(Authorization = paste('Client-ID', key)),
-                               cainfo = system.file("CurlSSL", "cacert.pem", package = "RCurl"))
+  resp <- httr::POST(url = "https://api.imgur.com/3/image.xml",
+                     config = httr::add_headers(Authorization = paste("Client-ID", key)),
+                     body = list(image = httr::upload_file(file)))
+  httr::stop_for_status(resp, task = "Fail to upload")
+  info_media <- httr::parse_media(resp$headers[["Content-Type"]])
+  if (info_media$complete != "text/xml") stop("Fail to upload; response is not XML")
+  if (is.null(encoding <- info_media$params$charset)) {
+    message("No encoding in response : defaulting to UTF-8")
+    encoding <- "utf-8"
+  }
+  res <- xml2::as_list(
+    xml2::read_xml(x = httr::content(resp, as = "raw"),
+                   encoding = encoding)
   )
-  res = XML::xmlToList(res)
-  if (is.null(res$link)) stop('failed to upload ', file)
-  structure(res$link, XML = res)
+  if (is.null(res$link[[1]])) stop('failed to upload ', file)
+  structure(res$link[[1]], XML = res)
 }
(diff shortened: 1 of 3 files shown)


R/utils-upload.R
```diff
@@ -35,13 +35,27 @@
 #' opts_knit$set(upload.fun = function(file) imgur_upload(file, key = 'your imgur key'))
 #' }
 imgur_upload = function(file, key = '9f3460e67f308f6') {
+
+  if (!has_package("xml2") || !has_package("httr")) {
+    stop("Pkg xml2 and httr needed for uploading file to imgur. Please install them or do not upload files.",
+         call. = FALSE)
+  }
```
yihui: I'd remove this if statement, and let `httr::` / `xml2::` fail directly later.

R/utils-upload.R
```diff
@@ -35,13 +35,27 @@
...
+  resp <- httr::POST(url = "https://api.imgur.com/3/image.xml",
+                     config = httr::add_headers(Authorization = paste("Client-ID", key)),
+                     body = list(image = httr::upload_file(file)))
+  httr::stop_for_status(resp, task = "Fail to upload")
+  info_media <- httr::parse_media(resp$headers[["Content-Type"]])
+  if (info_media$complete != "text/xml") stop("Fail to upload; response is not XML")
+  if (is.null(encoding <- info_media$params$charset)) {
+    message("No encoding in response : defaulting to UTF-8")
+    encoding <- "utf-8"
+  }
```
yihui: Is it safe to just assume UTF-8? Six lines of code (49-54) to find the encoding is probably not worth it in my eyes. This function does not need to be "bulletproof" :)

man/imgur_upload.Rd
```diff
@@ -18,8 +18,8 @@ A character string of the link to the image; this string carries an
   file; see Imgur API in the references.
 }
 \description{
-This function uses the \pkg{RCurl} package to upload a image to
```
yihui: Could you leave out the changes in Rd? https://github.com/yihui/knitr/blob/master/.github/CONTRIBUTING.md#roxygen2-documentation

cderv: Sorry for that. I should have read more closely your documentation. I reverted the changes.

R/utils-upload.R
cderv: I think we can assume that it is UTF-8. I did not found the info on imgurAPI but should be safe. You're right - no need to be too "bulletproof" 😄

R/utils-upload.R
cderv: done.

cderv:
I took care of your review comments. Thanks. 

There is no test for `imgur_upload`. I tested it with this code. 

```r
library(ggplot2)
ggplot(mtcars, aes(mpg, wt)) + geom_point()
ggsave("mtcars.png")
link <- imgur_upload("mtcars.png")
# should open the image on a browser
browseURL(link)
# should give XML attributes
str(attributes(link),1)
```

Not sure how to implement test with image upload... As it is not there already I did not create a test.

yihui:
Don't worry about the test. This function is called in https://github.com/yihui/knitr-examples/blame/master/010-upload.Rmd and the knitr-examples repo is automatically compiled on Travis. As long as this PR can pass the check on Travis, it should be fine.

BTW, I don't see your name in https://github.com/yihui/knitr/blob/master/DESCRIPTION. Could you add yourself as a contributor there?

cderv:
Yep! I was just thinking about that.  I forgot this file. It is not a habit yet as not every project ask you that. 
Sorry for the delay as I will do the change in the morning (Paris/France time). But we are not in a hurry. 

One more thing, I did not remove XML and RCurl as dependencies in DESCRIPTION. I leave that to you.

yihui:
Sounds good. No hurry.

cderv:
Done. I had myself as a contributor. It seems to be an alphabetical order of contributor in `Authors@R` so I respected it. 
Thank you for you review and letting me contribute! 😄

PR #1444, opened 2017-10-11
adds fig.ncol for latex output

### LINKED ISSUE #1327: Vertical alignment of subfigures
This question is related to the StackoverFlow post
http://stackoverflow.com/questions/30664820/subfigures-stacked-vertically-with-knitr
that never received an aswer and
http://stackoverflow.com/questions/12546365/subfigures-or-subcaptions-with-knitr
where the problem is also mentioned
I can use subfigures using knitr in the way described there, but they are always aligned horizontally. It would be an awsome feature to be able to introduce line breaks, so subfigures can be aligned vertically, can be aligned as a 2x2 grid etc...

http://stackoverflow.com/q/40932462/559676

eliocamp:
Adds new chunk option `fig.ncol` that controls the number of columns when using subfigures. (fixes #1327 )

Example and some tests: https://gist.github.com/eliocamp/7609fae98632b5af9a05d37262337bb9

example

### DIFF
--- a/R/hooks-latex.R
+++ b/R/hooks-latex.R
@@ -69,10 +69,11 @@ hook_plot_tex = function(x, options) {
   fig.cur = options$fig.cur %n% 1L
   fig.num = options$fig.num %n% 1L
   animate = options$fig.show == 'animate'
-  ncol = abs(options$fig.ncol %n% fig.num)
-  sep = options$fig.sep %n% c(rep('', ncol - 1), '\\newline')
-  ncol = length(sep)
-  col.cur = fig.cur - floor((fig.cur - 1)/ncol)*ncol
+  fig.ncol = options$fig.ncol %n% fig.num
+  if (is.null(sep <- options$fig.sep)) {
+    sep = character(fig.num)
+    if (fig.ncol < fig.num) sep[seq(fig.ncol, fig.num - 1L, fig.ncol)] = '\\newline'
+  }
   sep.cur = NULL
 
   # If this is a non-tikz animation, skip to the last fig.
@@ -115,10 +116,7 @@ hook_plot_tex = function(x, options) {
     if (usesub) {
       sub1 = sprintf('\\subfloat[%s%s]{', subcap, create_label(lab, fig.cur, latex = TRUE))
       sub2 = '}'
-      # Add separator command for floats if not last in set
-      # Add separator command for floats if not last in set
-      # Add separator command for floats if not last in set
-      if (!plot2) sep.cur = sep[col.cur]
+      sep.cur = sep[fig.cur]
     }
 
     # If pic is standalone/last in set:
--- a/R/hooks-latex.R
+++ b/R/hooks-latex.R
@@ -69,7 +69,11 @@ hook_plot_tex = function(x, options) {
   fig.cur = options$fig.cur %n% 1L
   fig.num = options$fig.num %n% 1L
   animate = options$fig.show == 'animate'
-  ncol = options$fig.ncol %n% .Machine$integer.max
+  ncol = abs(options$fig.ncol %n% fig.num)
+  sep = options$fig.sep %n% c(rep('', ncol - 1), '\\newline')
+  ncol = length(sep)
+  col.cur = fig.cur - floor((fig.cur - 1)/ncol)*ncol
+  sep.cur = NULL
 
   # If this is a non-tikz animation, skip to the last fig.
   if (!tikz && animate && fig.cur < fig.num) return('')
@@ -82,8 +86,6 @@ hook_plot_tex = function(x, options) {
   plot1 = ai || fig.cur <= 1L
   # TRUE if this picture is standalone or last in set
   plot2 = ai || fig.cur == fig.num
-  # TRUE if this picture is not standalone, nor last in set and is at the last column
-  plot3 = !plot2 && (fig.cur %% ncol == 0)
 
   # open align code if this picture is standalone/first in set
   align1 = if (plot1)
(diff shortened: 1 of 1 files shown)


R/hooks-latex.R
```diff
@@ -69,6 +69,7 @@ hook_plot_tex = function(x, options) {
   fig.cur = options$fig.cur %n% 1L
   fig.num = options$fig.num %n% 1L
   animate = options$fig.show == 'animate'
+  ncol = options$fig.ncol %n% 1L
```
yihui: Is one column a reasonable default? Perhaps the default should be no special layout?

R/hooks-latex.R
```diff
@@ -89,6 +92,9 @@ hook_plot_tex = function(x, options) {
   align2 = if (plot2)
     switch(a, left = '\\hfill{}\n\n', center = '\n\n}\n\n', right = '\n\n', '')
 
+  # initialize newline
+  newline = ''
```
yihui: Will `NULL` be more backward-compatible than `''`?

R/hooks-latex.R
eliocamp: Nope, and I realised my mistake when I went to sleep :upside_down_face: . If should (conceptually) be `Inf` .

R/hooks-latex.R
eliocamp: Since it's used in `paste()` I don't think there would be any difference. 
``` r
identical(paste0('a', '', 'b'), paste0('a', NULL, 'b'))
#> [1] TRUE
```
But I can change it if you think `NULL` fits better with the style of the code.

eliocamp:
I was thinking of a more general solution to the problem. A `fig.sep` option that is a vector of characters to put between figures. The equivalent of `fig.ncol = 2` would be `fig.sep = c(NULL, '\\newline')` (nothing between the first and second figure, and a new line between the second and third). But if one wants some separation between figures, one would use `fig.sep = c('\\hfill', '\\newline')`. 
What do you think?

yihui:
The general solution you proposed sounds good to me, but I'd like to keep `fig.ncol` as well because it is simpler to specify and is probably what users want most of time. If both `fig.ncol` and `fig.sep` are present, use the latter.

eliocamp:
I agree.
I'll make a PR implementing `fig.sep` and treating `fig.ncol` as a shortcut. Shouldn't be too hard, although I have to think how to treat the last plot.

eliocamp:
There. 
`fig.sep`'s default value is `c(rep('', fig.ncol -1), '\\newline\)` wich `fig.ncol`'s default value being `fig.num`. 
The last plot in a set has no separation otherwise it fails some tests. But I think it should be better for consistency if it used the same separation as the others. Otherwise this:
```
{r pressure, out.extra = "", echo = FALSE, fig.show = "hold", fig.cap = "Pressure", fig.subcap = c("one","two", "tree"), out.width = "0.48\\textwidth", fig.sep = c("1st\\newline", "2nd\\newline", "3rd\\newline")}
plot(pressure)
ggplot2::qplot(temperature, pressure, data = pressure)
plot(pressure)

```
Returns this: 
image

It feels weird not to render the "3rd" separator.

yihui:
Does the rewrite make sense to you? You don't need special treatment of the last plot.

```r
devtools::install_github('yihui/knitr#1444')
```

Test case:

````
---
title: "Test subfigures"
output:
  pdf_document: 
    extra_dependencies: subfig
---

```{r pressure, out.extra = "", echo = FALSE, fig.show = "hold", fig.cap = "Pressure", fig.subcap = c("one","two", "tree"), out.width = "0.48\\textwidth", fig.sep = c("1st\\newline", "2nd\\newline", "3rd")}
plot(pressure)
ggplot2::qplot(temperature, pressure, data = pressure)
plot(pressure)
```
````

image

**Comment by eliocamp** [comment]
Yep. Looks good! 🥇

**Comment by burnsla** [comment]
yihui, could you post an example of how this would work with include_graphics? I'm trying to use knitr to print multiple images (4) in a document without using par( ) to specify their positions ( e.g. 2 x 2) and add subcaptions or lettering for each. The r package 'multipanelfigure' does this but it's not elegant or easy to set the out.width dimensions so that each of the figures have the same dimensions.

**Comment by yihui** [comment]
@burnsla Sorry but I don't have time. You should know that `include_graphics()` is vectorized: https://stackoverflow.com/a/46836541/559676 Subfigures in multiple columns might work with a vector of image paths passed to `include_graphics()`. I'm not sure.

**Comment by eliocamp** [comment]
Can confirm that this works:

```
```{r pressure-ncol, fig.ncol = 2, echo = FALSE, fig.cap = "caption", fig.shor = "hold", fig.subcap = c("1", "2", "3", "4")}
library(knitr)
include_graphics(c(path1, path2, path3, path4))
```


image

yihui:
@eliocamp Excellent. [Thanks for trying it out and saving my time!!](https://yihui.name/en/2017/11/thanks-elio-campitelli/) Actually for the current version of **knitr**, I guess you still need the `out.extra = ""` hack?

eliocamp:
Yes, sorry. Was working with my local forked version. Very bad for reproducibility!!!! 👎

burnsla:
Yes, I've produced the same, but I can't seem to force the images to a new line. I'm currently using rmarkdown and can't take advantage of using fig.ncol or fig.sep. 

image

eliocamp:
@burnsla Can you provide the source code? In my machine, using pdf files, it works fine. Did you install the development version from github with `devtools::install_github("yihui/knitr")`?

burnsla:
@eliocamp, ah thank you. That was the issue. Works perfectly now.

lwtichem:
Is it possible to refer to specifically figure 4 (a) of for example the @burnsla example with the \@ref() function? 
I managed to refer to the equivalent of figure 4, but not to 4 (a).

darrenproj:
@yihui in the example of [this comment](https://github.com/yihui/knitr/pull/1444#issuecomment-346239114), if I add `fig.align='center'` to the second chunk, it produces

example

Only the last plot is aligned to center. What's wrong with it?

PR #1468, opened 2017-11-27
Allow use of _ in engine names

jimhester:
I would like to define a `glue_sql` engine, which works fine with `{r engine = "glue_sql"}` but fails if you use `{glue_sql}`. In this PR I added support for `.` separators as well, which seems natural.


jimhester:
Ok I updated the PR to only add `_`.

PR #1479, opened 2017-12-27
Update engine.R

sumandoc:
Added [Perl 6](https://perl6.org/) support.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -178,6 +178,24 @@ eng_python = function(options) {
   }
 }
 
+eng_perl6 <- function(options) {
+  # create a temporary file
+  f <- basename(tempfile("perl6", '.', paste('.', "p6", sep = '')))
+  on.exit(unlink(f)) # cleanup temp file on function exit
+  writeLines(options$code, f)
+  out <- ''
+  
+  # if eval != FALSE compile/run the code, preserving output
+  if (options$eval) {
+    out <- system(sprintf('perl6 %s', paste(f, options$engine.opts)), intern=TRUE)
+  }
+  
+  # spit back stuff to the user
+  engine_output(options, options$code, out)
+}
+
+
+
 ## Java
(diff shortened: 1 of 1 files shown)


yihui:
Thanks, but I think you can just do

    ```{perl, engine.path='perl6'}

https://yihui.name/knitr/options/#language-engines

sumandoc:
Thanks but
```
```{perl, engine.path='perl6'}
```

**doesn't** work in **windows** !!

The reason for this I believe could be:

- While `perl -e 'print "Hello"'` works for **Perl 5**
- `perl6 -e 'print "Hello"'` **doesn't** work for **Perl 6**

yihui:
What is the equivalent flag of `-e` in Perl 6?

sumandoc:
Looks like its `-e` only.

- https://docs.perl6.org/programs/00-running
- https://github.com/WildYorkies/Perl6-One-Liners/blob/master/README.md

yihui:
Since it is still `-e`, I don't know why `perl6 -e 'print "Hello"'` does not work on Windows. I don't know much about Perl, so you have to figure out what's wrong with `perl6 -e`.

sumandoc:
This was a reason to add `perl6` as a new engine in this PR. ;)

yihui:
Yes, I know that. I just don't like introducing new code when you can take advantage of existing code. If `perl6 -e` does not work, it is probably not my fault.

In general, I welcome pull requests, but I hesitate on accepting pull requests that introduce code that I'm not familiar with. Yes, this one is simple enough, but I just don't want to take the responsibility of answering possible future questions on `perl` vs `perl6`.

That said, knitr's language engine system is quite open, which means you can certainly modify it as you wish. `engine_output()` is an exported function, so you can just write your own engine function, and register it via `knitr::knit_engine$set(perl6 = whatever)`. You don't have to let me support it.

PR #1482, opened 2017-12-30
Added conversion of percent height to LaTex unit.

dataopt:
Before this patch, `out.height` for figures when specified as a percent leads to the percent sign get passed into pandoc causing parsing problems.  The patch translates the percent to the equivalent fraction of `\textheight`

Fixes https://github.com/rstudio/bookdown/issues/501

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -290,8 +290,8 @@ fix_options = function(options) {
 
   # turn x% to x/100\linewidth
   if (is_latex_output())
-    options['out.width'] = list(percent_latex_width(options[['out.width']]))
-    options['out.height'] = list(percent_latex_height(options[['out.height']]))
+    options['out.width'] = list(percent_latex_width_height('out.width', options))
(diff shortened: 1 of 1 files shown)


R/utils.R
```diff
@@ -355,6 +356,17 @@ percent_latex_width = function(x) {
   x
 }
 
+# turn percent height to LaTeX unit, e.g. out.height = 30% -> .3\textheight
+percent_latex_height = function(x) {
```
yihui: Apparently the code in this function pretty much duplicates `percent_latex_width()`. Can you try to merge the two into one and use an argument to control whether it is for width or height?

dataopt: Changes merging the two into one committed.

yihui:
Sorry for closing this PR. Normally I wouldn't do it, but this time the fix is simple enough, and I feel we probably need two more rounds of revisions in this PR, so I went ahead and did it by myself.

PR #1484, opened 2018-01-02
read_chunk doesn't ignore spin comments

ruaridhw:
I have an R source file which is formatted to contain commentary and code chunks for use with both `spin` and `read_chunk`. Currently when calling `read_chunk` the end of a chunk is recognised only by the start of the successive chunk. This means any commentary in between chunks is included in the previous chunk's code. Further, any "chunks" which are entirely comments are also created. This is not the case when calling `spin` which handles intermittent comments correctly.

This PR is to edit the behaviour of `read_chunk` to strip trailing `spin` comments from code chunks.

**example.R**
```r
#' # An R Script
#'
#' This is a valid R source file which is formatted
#' using the `knitr::spin` style comments and code
#' chunks.

# ---- chunk_one
print("Hello world!")

#' This is intermittent commentary that shouldn't
#' be included in the source of `chunk_one`

# ---- chunk_two
1 + 1
```

**Current behaviour**
```r
knitr::read_chunk("example.R")
knitr:::knit_code$get()
#> $`unnamed-chunk-1`
#> [1] "#' # An R Script"                                   
#> [2] "#'"                                                 
#> [3] "#' This is a valid R source file which is formatted"
#> [4] "#' using the `knitr::spin` style comments and code" 
#> [5] "#' chunks."                                         
#> 
#> $chunk_one
#> [1] "print(\"Hello world!\")"                                      
#> [2] ""                                                   
#> [3] "#' This is intermittent commentary that shouldn't"  
#> [4] "#' be included in the source of chunk_one"
#> 
#> $chunk_two
#> [1] "1 + 1"
```

**New behaviour**
```r
knitr::read_chunk("example.R")
knitr:::knit_code$get()
#> $`unnamed-chunk-1`
#> character(0)
#> 
#> $chunk_one
#> [1] "print(\"Hello world!\")"
#> 
#> $chunk_two
#> [1] "1 + 1"
```

Does this seem like an appropriate change? If so, should chunks that become empty post-strip (ie. `unnamed-chunk-1`) be avoided from being created via `knit_code$set` entirely?


yihui:
I'd like this new behavior to be optional. For example, you may add a new argument to `read_chunk()` to ignore roxygen comments `#'`.

ruaridhw:
Behaviour is now optional but the documentation for the additional parameter conflict with 7386aa3960e8126f8accaa35341b578b7af44277

yihui:
Yes, please resolve the conflicts (I don't have time for this kind of things). Ideally you should rebase against the master branch here, and force push, to make the diff in the PR as minimal as possible.

ruaridhw:
Done

R/parser.R
```diff
@@ -280,7 +282,8 @@ print.inline = function(x, ...) {
 #' knitr:::knit_code$get() # use this to check chunks in the current session
 #' knitr:::knit_code$restore() # clean up the session
 read_chunk = function(path, lines = readLines(path, warn = FALSE),
-                      labels = NULL, from = NULL, to = NULL, from.offset = 0L, to.offset = 0L) {
+                      labels = NULL, from = NULL, to = NULL, from.offset = 0L, to.offset = 0L,
+                      strip_comments = FALSE) {
```
yihui: I guess `roxygen_comments` may be a clearer name (defaults to `TRUE`).

R/parser.R
```diff
@@ -336,7 +339,19 @@ pattern_index = function(pattern, text) {
...
+  x <- x[-1]
+  if (!strip_comments) return(strip_white(x))
+  if (!length(x)) return(x)
+  while (is_blank(x[1]) || grepl("^#+'[ ]?", x[1])) {
+    x = x[-1]; if (!length(x)) return(x)
+  }
+  while (is_blank(x[(n <- length(x))]) || grepl("^#+'[ ]?", x[n])) {
+    x = x[-n]; if (n < 2) return(x)
+  }
+  x
```
yihui: The code is so similar to `strip_white()`, and it will be an excellent exercise for you to reuse code without copy-and-pasting (I stick firmly to [the DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)). Hint: you can add a function argument to `strip_white()`.

R/parser.R
ruaridhw: So the new behaviour is the default?

R/parser.R
ruaridhw: Will do, I was hoping to avoid the duplication but wasn't sure whether to edit `strip_white` given it is used throughout the package whereas `strip_chunk` is not.

R/parser.R
yihui: No, `roxygen_comments = TRUE` means `strip_comments = FALSE` (keep roxygen comments).

R/parser.R
yihui: Just use a reasonable default value for the new argument, so that the previous behavior is not changed by default, but customizable, e.g., `function(x, test = function(line) is_blank(line))`.

ruaridhw:
I've made the requested changes so let me know if there is anything else.

yihui:
Looks great. Thanks!

Issue #1485, opened 2018-01-02
Feature Request: More informative error for kable col.names


billdenney:
When `col.names` is specified for `kable`, but it is the wrong length, the error is relatively uninformative:

```r
> knitr::kable(data.frame(A=1), col.names=c("A", "B"))
Error in dimnames(x) <- dn : 
  length of 'dimnames' [2] not equal to array extent
> 
```

A simple modification here would make it much more informative:

https://github.com/yihui/knitr/blob/03682e0444457d4891474c1480c92a8eae558d9d/R/table.R#L160

Change that line to these lines (or similar):

```r
if (ncol(x) == length(col.names)) {
  colnames(x) = col.names
} else {
  stop("Argument col.names is not the same length as the number of columns of x.")
}
```

If interested, I can submit this as a pull request.

yihui:
Yes, please submit a PR, and I prefer `stop()` to appear earlier: https://yihui.name/en/2018/01/stop-early/

billdenney:
I put it right at the top.

(And, I prefer the same style-- do all input checking at the top so that more assumptions can be made within the code itself.  It simplifies the more complex parts of the code.)

billdenney:
As discussed within the PR, the change appears to be more work than it's worth.

PR #1486, opened 2018-01-02
Make col.names length mismatch more informative

billdenney:
Fix yihui/knitr#1485

### DIFF
--- a/R/table.R
+++ b/R/table.R
@@ -82,11 +82,6 @@ kable = function(
   x, format, digits = getOption('digits'), row.names = NA, col.names = NA,
   align, caption = NULL, format.args = list(), escape = TRUE, ...
 ) {
-  # check input
-  if (!identical(col.names, NA) && length(col.names) != ncol(x)) {
-    stop("Argument col.names is not the same length as the number of columns of x.")
-  }
-
   # determine the table format
   if (missing(format) || is.null(format)) format = getOption('knitr.table.format')
   if (is.null(format)) format = if (is.null(pandoc_to())) switch(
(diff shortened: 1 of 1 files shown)


R/table.R
```diff
@@ -82,6 +82,10 @@ kable = function(
   x, format, digits = getOption('digits'), row.names = NA, col.names = NA,
   align, caption = NULL, format.args = list(), escape = TRUE, ...
 ) {
+  # check input
```
yihui: The check is too early here, because `x` is not necessarily rectangular. It could be a list. Please move it to line 160 (before `colnames(x) = col.names`).

R/table.R
```diff
@@ -82,6 +82,10 @@ kable = function(
...
+  if (!identical(col.names, NA) && length(col.names) != ncol(x)) {
+    stop("Argument col.names is not the same length as the number of columns of x.")
```
yihui: _of_ the same length?

R/table.R
billdenney: Updated

R/table.R
billdenney: I modified the text to hopefully be clearer.  I think that's what you were asking for.  If not, can you suggest the exact text you'd like?

R/table.R
```diff
@@ -82,7 +82,6 @@ kable = function(
   x, format, digits = getOption('digits'), row.names = NA, col.names = NA,
   align, caption = NULL, format.args = list(), escape = TRUE, ...
 ) {
-
```
yihui: Could you not touch white spaces?

R/table.R
yihui: I meant

    Argument col.names is not of the same length as the number of columns of x.

I emphasized "_of_" in my comment but I understand it is easy to ignore it.

billdenney:
I agree that it's probably not worth the `R CMD check` headaches.  We can drop this change as it was mostly cosmetic.

yihui:
Perfect. Thank you!

Issue #1506, opened 2018-02-12
Suggestion: Replace non-standard Unicode characters with entities for HTML output


martinmodrak:
On Windows, locale sometimes messes up Unicode characters in HTML output from knitr. While this can be avoided with proper locale, for HTML, it can also be avoided by using HTML entities. HTML entities also could be somewhat preferable to raw Unicode (not sure about this, really). So my suggestion is to do this transformation by default.

If you agree, I am ready to implement this, but I am not sure, if this transformation should be implemented in knitr, or by modifying the ```escape_html``` function in [highr](https://github.com/yihui/highr/blob/master/R/utils.R).

This came up while working on an issue in skimr (https://github.com/ropenscilabs/skimr/issues/278) where ```kable``` can mess up histograms built from Unicode characters.

yihui:
I'm afraid this is not simple to fix, and **knitr** is probably not the best place to fix it (it goes back to the **evaluate** package and base R). There have been long-standing issues like yours:

- https://github.com/r-lib/evaluate/issues/59
- https://github.com/r-lib/evaluate/issues/66

martinmodrak:
I understand that the issue is more complex. But I know and have tested that taking UTF-8 string, converting it to HTML entities before passing to knitr and then pass to knitr (with ```escape=FALSE```), is a feasible workaround. The problem is that setting ```escape=FALSE``` introduces other problems, which is why it would make sense to implement the workaround in knitr. 

I understand if you consider this an ugly hack or don't like the penalty of increasing output size, but I believe this workaround could solve some real user problems (I may be oblivious to other downsides).

yihui:
Is the hack going to be something like `gsub("<U+2583>", "&#2583;", x)`? :)

martinmodrak:
For HTML, you can be a little cleaner - I noticed that when using ```kable``` (at least on my machine) UTF-8 strings arrive into ```kable``` just OK. So, when format is HTML, you could transform UTF-8 character to entities before R gets to mangle with them. This is (arguably) more correct than just replacing sequences of the form ```<U+XXXX>``` with the entities (since the user can safely but literal ```<U+XXXX>``` strings in the data they want to be processed).

But thinking some more, it might also make sense to do all the processing you do, and then replace the ```<U+XXXX>``` characters in the final output with actual UTF-8 chars. As long as the output string has UTF-8 encoding set, this could be OK (at least in RStudio and file output, but I dunno what it does to base R terminal), and work for all output formats.

[...]

PR #1508, opened 2018-02-14
Removed call to base::format to avoid mangling UTF-8 chars

martinmodrak:
Refs #1506


martinmodrak:
Also a small note - I scanned the code for other uses of ```base::format```. It is also called from ```format_args``` and ```format_matrix```, but if I read the code correctly, those are only ever called on numeric data and thus should be fine.

Issue #1505, opened 2018-02-11
knitr engine API and cache compatibility with reticulate engine


tmastny:
I think whether or not this is a knitr or [reticulate](https://github.com/rstudio/reticulate) bug depends on the knitr engine API, which I do not completely understand. I carefully searched for this bug in the knitr and reticulate issues and didn't see anything, so I apologize if this is already known. 

## The bug

Suppose we have the following file called `python_test.Rmd`:
````
---
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE, warning=FALSE, message=FALSE, 
                      results='show', cache=TRUE, autodep=FALSE)
knitr::opts_knit$set(progress = TRUE, verbose = TRUE)
knitr::knit_engines$set(python = reticulate::eng_python)
```

```{python chunk1}
x = 1
print(x)
```


```{python chunk2}
print(x + 9)
```
````
When you first press the **Knit** button the document compiles successfully.

Now suppose I change chunk2, so it has `print(x + 10)` and I save the file. If I try clicking the **Knit** button I get the following error:
```
Quitting from lines 19-20 (python_test.Rmd) 
Error in py_run_string_impl(code, local, convert) : 
  NameError: name 'x' is not defined

Detailed traceback: 
  File "<string>", line 1, in <module>
Calls: <Anonymous> ... force -> py_run_string -> py_run_string_impl -> .Call
Execution halted
```
----
## My efforts to debug
Here's what I've learned about the error:

It reliably happens when I call `knitr::knit('python_test.Rmd', envir = new.env())` *after a session restart*, so I don't think it is an `rmarkdown::render` error. 

The error message points to `py_run_string_impl`, which is a `reticulate` function. But I believe the problem arises before `knitr` reaches the python engine. 

When you call `knitr::knit('python_test.Rmd', envir = new.env())`, chunk1 eventually enters the `call_block` function in `knitr/R/block.R`. It passes `if (params$cache > 0)`, and the hash comes up with the same value. Then `cache$load` tries to bring the saved data into `knit_global`. 

At this point, if you `ls(knit_global())` you'll see `character(0)`. So it isn't clear if the python object `x=1` was even saved. 

[...]

PR #1518, opened 2018-02-26
cache engine API

tmastny:
My proposal is to add a hook for knitr language engines that allows an additional caching mechanism for that language's environment. This addresses issue #1505.

The basic idea is to add a [new object](https://yihui.name/knitr/objects/) that can attach cache engines:
```r
knitr::cache_engines$set(python = reticulate::cache_eng_python)
```

This is also relevant to pull request [#167](https://github.com/rstudio/reticulate/pull/167) on the `reticulate` repo, which implements a caching mechanism using this API. These pull requests are intertwined, because they use the other's new API. 

I wasn't sure how to implement testing, since these two issues overlap. So I refer you to the tests in the `reticulate` pull request. In particular, [this unit test](https://github.com/tmastny/reticulate/blob/master/tests/testthat/test-python-cache-engine.R) implements the issue described in #1505 and verifies that the solution works. 

This does not implement any `autodep` for the cache engine.


R/engine.R
```diff
@@ -31,6 +31,24 @@
...
+#' to the chunk cache and loads it into the language environment.
+#'
+#' The cache engine function has one argument \code{cache_path}. The
+#' argument is the path to the current chunk cache with the chunk's
+#' hash, but without any file extension.
+#'
+#' The cache engine function should load the cache environment and should
+#' know the extension appropriate for the language.
+#' @export
+#' @examples cache_engines$set(python = reticulate::load_python_session)
```
yihui: `reticulate::load_python_session` is not available in reticulate yet, so I cannot use this example at the moment. Once https://github.com/rstudio/reticulate/pull/167 is merged, I'll just set this caching engine for python in knitr.

R/engine.R
```diff
@@ -178,6 +196,19 @@ eng_python = function(options) {
   }
 }
 
+eng_python_cache = function(options) {
+  if (isFALSE(options$python.reticulate)) {
+    eng_interpreted(options)
```
yihui: In this case, you can just `return()` instead of running the python code chunk again. Results in a previously run [have been saved](https://github.com/yihui/knitr/blob/03b6e40d0969cdee4c01205bd31b096e85797a81/R/block.R#L111-L114) and [will be loaded](https://github.com/yihui/knitr/blob/03b6e40d0969cdee4c01205bd31b096e85797a81/R/block.R#L76).

R/engine.R
```diff
@@ -178,6 +196,19 @@ eng_python = function(options) {
...
+  } else {
+    if (!loadable('reticulate')) warning2(
```
yihui: I'd remove this warning, since users will see it in the first time. The whole function here can be

```r
eng_python_cache = function(options) {
  if (isFALSE(options$python.reticulate)) return()
  # TODO: change this hack to reticulate::load_python_session(options) after
  # https://github.com/rstudio/reticulate/pull/167 is merged and released
  if (!'load_python_session' %in% ls(asNamespace('reticulate'))) return()
  fun = getFromNamespace('load_python_session', 'reticulate')
  fun(options)
}
```

R/parser.R
```diff
@@ -251,6 +251,10 @@ print.inline = function(x, ...) {
 #' @param roxygen_comments Logical dictating whether to keep trailing
 #'   roxygen-style comments from code chunks in addition to whitespace
 #' @return As a side effect, code chunks are read into the current session so
+#'   that future chunks can (re)use the code by chunk label references.
+#'   If an external chunk has the same label as a chunk in the current session,
+#'   chunk label references by future chunks will refer to the
+#'   external chunk.
```
yihui: Do you know how to `git rebase` against my master branch here? If you do, please rebase and force push to get rid of the superfluous changes here.

tmastny: Thanks for the comments! I'm busy all weekend, so I incorporate the changes next week.

yihui: Sounds good. There is no hurry at all. Please take your time!

PR #1519, opened 2018-02-27
Fix #1487

### LINKED ISSUE #1487: kable: Vertical page position with "valign" option does behave faulty
First, thank you for the great work!

I like to produce a floating table in `latex` with the `kable()`-function. Next to that I tried to figure out how the page position is set. There is an option 'valign' which comes from a hidden function `kable_latex` and is passed through when it is applied in `kable(dat,valign = "!b")`.

Producing a simple table we can do it like this:

```{r}
require(knitr)
dat <- data.frame(A=c(1,2,3,4),B=c(T,F))

kable(dat,format = "latex")
```
which will give the output as expected:
```
\begin{tabular}{r|l}
\hline
A & B\\
\hline
1 & TRUE\\
\hline
2 & FALSE\\
\hline
3 & TRUE\\
\hline
4 & FALSE\\
\hline
\end{tabular}
```
Putting a caption to the table, will generate a floating object with latex - `table` environment. But in case we like to have the table at the bottom of the page, which is usually done by `\begin{table}[!b]` the '[!b]' switch appears besides the ´tabular´ environment an NOT aside the ´table´. Proof:

R-code
```
require(knitr)
dat <- data.frame(A=c(1,2,3,4),B=c(T,F))

kable(dat,format = "latex",caption = "abc",valign = "!b")
```
And its output:
```
\begin{table} % <-- here [!b] is supposed to be

\caption{\label{tab:}abc}
\centering
\begin{tabular}[!b]{r|l}   %% here is the wrong position for [!b]
\hline
A & B\\
\hline
1 & TRUE\\
\hline
2 & FALSE\\
\hline
3 & TRUE\\
\hline
4 & FALSE\\
\hline
\end{tabular}
\end{table}
```
Since the functionality is given, but accidentally wrong composed, I would also suggest to document this option in the help of `kable` function.

haozhu233:


haozhu233:
Added. Thanks! :)

DESCRIPTION
```diff
@@ -88,7 +88,8 @@ Authors@R: c(
     person("Viktoras", "Veitas", role = "ctb"),
     person("Weicheng", "Zhu", role = "ctb"),
     person("Wush", "Wu", role = "ctb"),
-    person("Zachary", "Foster", role = "ctb")
+    person("Zachary", "Foster", role = "ctb"),
+    person("Hao", "Zhu", role = "ctb")
```
yihui: Sorry I forgot to mention the alphabetical order by the first names (I thought it was obvious).

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# CHANGES IN knitr VERSION 1.21 (unreleased)
+
+## NEW FEATURES
+
+## BUG FIXES
+
+- `valign` in `kable_latex()` does not put the float alignment to the correct 
+location (#1487, #1519)
```
yihui: Could you not hard-wrap the line?

You could say `thanks, @haozhu233, #1487, #1519` (like other entries in NEWS), although I know it feels odd to thank yourself :)

NEWS.md
```diff
@@ -1,3 +1,12 @@
+# CHANGES IN knitr VERSION 1.21 (unreleased)
+
+## NEW FEATURES
```
yihui: No new features yet. I can add this header when necessary. For now, I don't need it.

PR #1520, opened 2018-02-27
Add classes to table rows in kable_html

jreduardo:
Just add classes to table rows (`<tr>`) html tag. This keep consistency with pandoc converter, see example:

**In example.Rmd file**

<pre lang="r">
---
output: html_document
---

```{r}

knitr::kable(head(cars, 3))
knitr::kable(head(cars, 3), format = "html")

```
</pre>

Compiling **example.Rmd** to **example.html** (`rmakdown::render` or `knitr::knit` + `knitr::pandoc`) we obtain (suppressing  others html definitions):

```
<!-- Pandoc to html -->                <!-- knitr::kable to html -->
  <table>                              |  <table>
    <thead>                            |    <thead>
      <tr class="header">              |      <tr>
        <th align="right">speed</th>   |        <th style="text-align:right;">speed</th>
        <th align="right">dist</th>    |        <th style="text-align:right;">dist</th>
      </tr>                            |      </tr>
    </thead>                           |    </thead>
    <tbody>                            |    <tbody>
      <tr class="odd">                 |      <tr>
        <td align="right">4</td>       |        <td style="text-align:right;">4</td>
        <td align="right">2</td>       |        <td style="text-align:right;">2</td>
      </tr>                            |      </tr>
      <tr class="even">                |      <tr>
        <td align="right">4</td>       |        <td style="text-align:right;">4</td>
        <td align="right">10</td>      |        <td style="text-align:right;">10</td>
      </tr>                            |      </tr>
      <tr class="odd">                 |      <tr>
        <td align="right">7</td>       |        <td style="text-align:right;">7</td>
        <td align="right">4</td>       |        <td style="text-align:right;">4</td>
      </tr>                            |      </tr>
    </tbody>                           |    </tbody>
  </table>                             |  </table>  

```

So, this pull request just keeps the same table row classes for both.

### DIFF
--- a/R/table.R
+++ b/R/table.R
@@ -289,7 +289,7 @@ kable_html = function(x, table.attr = '', caption = NULL, escape = TRUE, ...) {
     sprintf('<table%s>%s', table.attr, cap),
     if (!is.null(cn <- colnames(x))) {
       if (escape) cn = escape_html(cn)
-      c(' <thead>', '  <tr>', sprintf('   <th%s> %s </th>', align, cn), '  </tr>', ' </thead>')
+      c(' <thead>', '  <tr class="header">', sprintf('   <th%s> %s </th>', align, cn), '  </tr>', ' </thead>')
     },
     '<tbody>',
     paste(
--- a/R/table.R
+++ b/R/table.R
@@ -276,6 +276,7 @@ kable_latex_caption = function(x, caption) {
 }
 
 kable_html = function(x, table.attr = '', caption = NULL, escape = TRUE, ...) {
+  trclass <- rep(c("odd", "even"), length.out = nrow(x))
   table.attr = trimws(table.attr)
   # need a space between <table and attributes
   if (nzchar(table.attr)) table.attr = paste('', table.attr)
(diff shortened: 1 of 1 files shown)


jreduardo:
I undestand your concerns. Really, change `knitr` functions is like to walk on eggshells :smile:. 

I included an option to `knitr::kable_html` to add html classes (751b5d2). This seems ok?

yihui:
Yes, implementation-wise, it looks good to me. I just don't understand why we need it in the first place. I need a reason that is not related to CSS (for striped tables).

jreduardo:
It's just for style. It is for `knitr::kable(..., format = "html")` behave like markdown tables translated by pandoc.

yihui:
That does not sound convincing enough... If it is only for styling the rows, there is no need to use these classes, as I said in my first response.

jreduardo:
My point is: suppose you are writing a blog or an HTML document, it doesn't seem good that you have to change CSS when you use `knitr::kable(..., format = "html")`, whereas it isn't needed when you use `knitr::kable(...)`.

Of course, I can obtain the same style using `nth:child`, but I would expect the same HTML result for both (`knitr::kable(...)` and `knitr::kable(..., format = "html")`).

yihui:
My point is Pandoc does not need it, either :) The classes must be there for historical reasons (before CSS3), although I didn't confirm it with Pandoc's developer.

`:nth-child` works no matter if you have the odd/even classes or not.

yihui:
Many thanks for your understanding! 🙇

PR #1525, opened 2018-03-14
deals with xml2 >= 1.2.0 change

### LINKED ISSUE #1523: needs to import xml2 > 1.2.0 for imgur_upload
I believe the #1495 was due to a breaking change in xml2 1.2.0 in `as_list` that now return the root node. Fix in #1496 is a way to deal with that. 
However knitr now needs xml2 >= 1.2.0 for `imgur_upload` to work properly. 

I could to a PR with that addition Un description or test in `imgur_upload` for which version of xml2 and do the right thing accordingly. What do you think?

cderv:
In version xml2 1.2.0, there is a breaking change in the structure returned by as_list. The root node is now return that adds one more level on top.

This PR deals with that by

+ adding a minimal version to xml2 suggests
+ dealing with the breaking change if xml2 version is >= 1.2.0

this fixes #1523.

I did not had a bullet to NEWS for this small change.

---
This change is [Reviewable](https://reviewable.io/reviews/yihui/knitr/1525)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -120,7 +120,7 @@ Suggests:
     JuliaCall (>= 0.11.1),
     png,
     jpeg,
-    xml2,
+    xml2 (>= 1.2.0),
     httr,
     DBI (>= 0.4-1),
     showtext,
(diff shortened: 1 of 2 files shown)


cderv:
@jennybc This should also deal with tidyverse/reprex#170 for all version of xml2.

PR #1531, opened 2018-04-04
Increment version to 1.20-9000

ck37:
Hello,

This is a quick PR to update DESCRIPTION to be version 1.20-9000, which will differentiate daily builds from the 1.20 release. Then when 1.21 is ready to be released it can be converted to 1.21, and then 1.21-9000.

Thanks,
Chris


ck37:
Cool cool, thanks!

PR #1542, opened 2018-05-09
Fix the issue with SQL chunk when cache=TRUE and output.var is specified

yutannihilation:
fixes https://github.com/rstudio/rmarkdown/issues/914

IIUC, SQL chunk is special in the sense the result of it can be either a text output or a data.frame, depending on whether `output.var` is set or not. Since both can be saved as an R object, it may be possible to use the same cache file. But, for simplicity and consistency with other engines (non-R engines seem to have their own cache file other than one for text outputs), I implemented to cache using the separate file, suffixed `_sql_result.rds`.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -605,13 +605,23 @@ eng_sql = function(options) {
   if (options$results == 'hide') output = NULL
 
   # assign varname if requested
-  if (!is.null(varname)) assign(varname, data, envir = knit_global())
+  if (!is.null(varname)) {
+    assign(varname, data, envir = knit_global())
+    saveRDS(data, file = paste0(options$hash, "_sql_result.rds"))
+  }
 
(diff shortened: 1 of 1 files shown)


yutannihilation:
Thanks, confirmed! It looks perfect. I didn't understand the way how object is passed to cache.

Since this PR seems superseded by your idea, I'm closing this. I'll try to achieve some actual contribution next time! 💪

yihui:
Please go ahead to submit another PR with my approach above. Thanks! :)

yutannihilation:
Hmm, generally, I rather hesitate to be the author of other people's idea... But, if you say so, I will. Thanks for always trying to educate me :)

yihui:
Haha, actually I almost merged this PR, and I suddenly remembered I did something special for the `stan` engine. It turned out that the `sql` engine could use the same idea.

PR #1544, opened 2018-05-10
Cache the result of SQL chunk properly

yutannihilation:
(The original author of this idea is @yihui: https://github.com/yihui/knitr/pull/1542#pullrequestreview-119148409)

fixes rstudio/rmarkdown#914

When the engine is not R and the result of R object exists, the object should be passed to `block_cache`; SQL chunk and Stan chunk are the cases. Currently, Stan chunk does the right thing. This PR also make the result of SQL chunk to be cached.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 - `engine.path` does not work for `engine = 'dot'` (thanks, @billy34, #1534).
 
+- `sql` engine now caches the result properly when `output.var` is specified (thanks, @yutannihilation, #1544).
+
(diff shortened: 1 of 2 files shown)


yutannihilation:
Here you are!

BTW, "(thanks, @..., #...)" at the end seems the format, but I feel a bit weird to say thanks to myself... lol

yihui:
I guess it will be even more weird to `thank @yihui` 😀

yutannihilation:
Thanks!

Issue #1546, opened 2018-05-20
Encoding on Windows 7


crossaw:
I wrote a report detailing how output is encoded from the [Knit] button in RStudio.  My report is available <a href="https://crossaw.github.io/Knit-Win7-Encoding-Report/">here</a>.  The supporting files can be accessed on the <a href="https://github.com/crossaw/Knit-Win7-Encoding-Report">master branch of the github repo</a>.

I know that encoding on Windows 7 is an old issue, and there may not be much value in working on it.  However, I have not seen any discussions that describe the behavior with the detail that I provide in my report.  I'm hoping that with that level of detail, a culprit might become readily apparent, and you could notify the developers of the offending package or easily make an improvement to knitr.

I have sent an e-mail to R-core about my report, and I have posted the links on nabble and in the RStudio Community.

cderv:
Just for complete cross reference, the post on Rstudio community is here:
https://community.rstudio.com/t/encoding-of-knitr-output-on-windows-7/8612

lgallindo:
See this: https://github.com/yihui/knitr/pull/1553

lgallindo:
By the way, it also affects Windows 8 and 8.1

crossaw:
Yesterday I updated R, RStudio and all my packages, and I reran the tests for all the reported encoding settings.  There was no change in the results.  I put a note at the beginning of the report.

I opened this issue primarily as a way to circulate my report.  I'm not really expecting a specific solution, so I'm not sure at what point I should close the issue.  By now I figure that most people who care are aware of my report, so I figure I'll leave this issue open for maybe another month and then just close it, unless someone thinks it should be left open.

Issue #1343, opened 2017-01-09
Some non-ASCII characters that exist in native encoding are wrongly encoded when knit()-ing


huftis:
When knit() a UTF-8 file (at least under Windows), some characters disappear under some circumstances. This seems to happen with characters that exist in the native encoding (windows-1252 on Windows) but non in latin1 (ISO-8859-1) or UTF-8, and only happens when printing the characters from R.

Here’s an example document, saved as UTF-8:

    \documentclass{article}
    \begin{document}
    
    æ, ø, å
    1–2
    
    <<>>=
    "æ, ø, å"
    "1–2"
    @
    
    \end{document}

When compiling this (using `knit2pdf("test.Rnw", encoding="UTF-8")`), the first two lines containing

    æ, ø, å
    1–2

are output correctly, but for the version created **inside the R chunk**, the output for the `1–2` string is wrongly encoded for the en-dash (–). The output for the `æ, ø, å` string is *correctly* encoded. One can check this using:

    library(readr)
    d = read_lines_raw("test.tex")
    d[c(64,70)]

which results in

    [[1]]
     [1] 23 23 20 5b 31 5d 20 22 c3 a6 2c 20 c3 b8 2c 20 c3 a5 22

    [[2]]
     [1] 23 23 20 5b 31 5d 20 22 31 c2 96 32 22

This is meant to represent

    [1] "æ, ø, å"
    [1] "1–2"

As you can see, the en-dash is encoded as `c2 96`. The correct encoding in UTF-8 would be `e2 80 93`. (And the correct encoding in the native encoding, windows 1252, is `96`.) The *first* time the string occurs, i.e., outside the R chunk, it *is* correctly encoded:

    > d[56]
    [[1]]
    [1] 31 e2 80 93 32

Note that the en-dash is one of the rare characters that are *not* available in latin1 (ISO 8859-1) but *are* available in the native Windows encoding (Windows-1252, AKA cp1252). These are the letters listed in red in this image at Wikipedia: https://en.wikipedia.org/wiki/Windows-1252

The encoding problems seems to happen all of these characters. For example, `“` is written as `c2 93` by knitr. The Windows-1252 encoding would be `93`, while the *correct* UTF-8 encoding would be `e2 80 9c`. So it seems like the Windows-1252 encoding is used, but prefixed by `c2`, for some reason.

[...]

PR #1553, opened 2018-06-08
Workaround for UTF-8 issues on Windows 7/8.

lgallindo:
Added and boolean parameter to knit(): lyx_compatible. If false, knit() will skip the call to correct_encode(). Functions that call knit() also received the extra parameter. The default value for the lyx_compatible parameter is TRUE, maintaining compatibility.

This provides a workaround for knitr falling back to Windows 1252 encoding on a Windows 7-8 platform correctly configured for UTF-8.

A proper solution would be review correct_encode(), but I'm not a LyX user and am afraid to break compatibility.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -22,3 +22,4 @@ Makefile
 ^revdep$
 ^docs$
 ^README-ES\.md$
+^README-PT\.md$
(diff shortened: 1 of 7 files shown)


lgallindo:
Related issue: https://github.com/yihui/knitr/issues/1546

lgallindo:
Another related issue: https://github.com/yihui/knitr/issues/1343

yihui:
@lgallindo I'd rather figure out how `correct_encode()` returned a wrong encoding instead of adding a new argument. You could help me in two possible ways: 1) answer my question above, or 2) provide an example with which I can reproduce the issue, so I can answer the question by myself. Thanks!

lgallindo:
Ok. Will try to build a MWE tomorrow to check this.

yihui:
Just a friendly ping :)

Issue #1516, opened 2018-02-23
Use styler instead of formatR?


krlmlr:
We're about to submit an update for *styler* to CRAN, this would be a good point to use *styler* in addition to/instead of *formatR* for styling code chunks. What's the best way to proceed, would you like to review a PR?

CC @lorenzwalthert.

yihui:
Yes, PR please:
https://github.com/yihui/knitr/blob/1e9a4d51c53c764bb3b8d288a7535a8bb79131f5/R/block.R#L148-L154

A chunk option like `tidy.method = 'styler'` should be good (defaults to `formatR`).

PR #1581, opened 2018-07-28
Adding styler support

lorenzwalthert:
Finally, I got time to try to add styler support as suggested in #1516. I have the following questions: 

* As of now, there are two options `tidy` (defaults to  `FALSE`) and `tidy.method` (defaults to `"formatR"`). We could remove the `tidy` option. This would result in a backward-incompatible API change, which is probably not a good idea. I think it would make sense to simply issue a warning when someone sets `tidy.method = "styler"` and (either because of explicit specification or because of the default) `tidy = FALSE`, because he or she probably just wants `tidy = TRUE` and `tidy.method = styler`. I have not yet implemented that, so let me know if I should go ahead with that.
* I don't know what resources (e.g. documentation, cheat sheets etc.) need to be updated to reflect the API change, at least I could not find any help file where the `tidy` option was documented.
* Also, I wasn't sure about if and how I should add unit tests for this. 

Could you please advise on these points? Thanks.

cc: @krlmlr

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -127,7 +127,8 @@ Suggests:
     httr,
     DBI (>= 0.4-1),
     showtext,
-    tibble
+    tibble,
+    styler
 License: GPL
 URL: https://yihui.name/knitr/
 BugReports: https://github.com/yihui/knitr/issues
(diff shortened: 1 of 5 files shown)


krlmlr:
How about supporting `tidy = "styler"` and `tidy = "formatR"`, in addition to `TRUE` and `FALSE` ?

lorenzwalthert:
Cool, thanks @yihui.

Issue #712, opened 2014-01-22
Animated plots, hook_ffmpeg_html and video format


basille:
Since knitr 1.3, animated plots using `hook_ffmpeg_html` uses OGG video flies instead of MP4. I think it is a sensible default (although WebM would be equally good, according to [browser support](http://www.w3schools.com/html/html5_video.asp). However, would it be possible to have the file type as an option, so that people could, if needed, switch to MP4? 

It might be already possible, in which case I apologize: I couldn't find a way to do it. As far as I understand it, the file type is hard-coded in `hook_ffmpeg_html`:

```
mov.fname = str_c(sub(paste(fig.num, "$", sep = ""), "", 
    x[1]), ".ogg")
```

After a couple of trials, I couldn't find a way to do it manually. For instance, modifying directly `hook_ffmpeg_html` using:

```
knit_hooks$set(hook_ffmpeg_html = function(x, options) 
    ...
    mov.fname = str_c(sub(paste(fig.num, "$", sep = ""), "", 
        x[1]), ".mp4")
    ...
```

does nothing: output is still a OGG file. Or setting up a new `hook_ffmpeg_html_mp4`, which would be called by `opts_knit$get(animation.fun = hook_ffmpeg_html_mp4)`, results in an error. In the end, I'm not too sure what to modify or what option to change to make this choice possible.

By the way, I'm using knitr 1.5.
Mathieu.

yihui:
FYI, I just added a chunk option `ffmpeg.format` so you can easily change the video format from the default webm to mp4 by using the chunk option `ffmpeg.format = 'mp4'`.

basille:
Hey Yihui,

This looks great! Did the new option make it to v1.11? I couldn't have it work, with e.g. `ffmpeg.format = "mp4"`. Shall I use the dev version instead?

In addition, does the option allow to change other settings, or is it necessary to go through `opts_knit$set`/`hook_ffmpeg` for detailed settings?

Thanks for your work!
Mathieu.

yihui:
@basille It is in the development version: https://github.com/yihui/knitr/blob/master/NEWS.md (1.12 is not released yet) This option does not allow you to change any other settings.

petrelharp:
ffmpeg [currently](https://trac.ffmpeg.org/ticket/5108) has a bug in its png -> webm output, so having an easy way to fall back on mp4 is timely.

PR #1586, opened 2018-08-07
Add flags to ffmpeg mp4 hook for Safari support

brianzhang01:
With the current ffmpeg mp4 converter setup in knitr in-line
animation, I was not able to have my animations show up in
Safari. After adding the flags in this commit, it worked.

https://apple.stackexchange.com/questions/166553/why-wont-video-from-ffmpeg-show-in-quicktime-imovie-or-quick-preview

I found the Safari viewability to be sensitive to exactly where
the flags were placed. For instance, running

ffmpeg -y -r 1 -pix_fmt yuv420p -i foo-%d.png foo.mp4

did not work, otherwise I would consider aliasing my ffmpeg
command. It turned out to be necessary to include the -i flag
first, as so:

ffmpeg -y -r 1 -i foo-%d.png -pix_fmt yuv420p foo.mp4

### DIFF
--- a/R/hooks-html.R
+++ b/R/hooks-html.R
@@ -95,6 +95,9 @@ hook_ffmpeg = function(x, options, format = 'webm') {
 
   extra = if (format == 'webm') {
     paste('-b:v', options$ffmpeg.bitrate %n% '1M', '-crf 10')
+  } else if (format == 'mp4') {
+    # enables Safari support of .mp4
+    paste('-pix_fmt yuv420p')
   }
   ffmpeg.cmd = paste(
     'ffmpeg', '-y', '-r', 1 / options$interval, '-i', fig.fname, extra, mov.fname


brianzhang01:
Here is some more information about the motivation behind this pull request.

I really like knitr and I especially found the in-line animation feature appealing -- it saves code and is more intuitive to just make multiple figures within a for loop. It turns out that Safari, my default mobile browser, does not support webm or ogg, the two formats that have been used as the default for knitr output, so I was interested to look into mp4 output. (See https://github.com/yihui/knitr/issues/712, https://www.w3schools.com/html/html5_video.asp)

I then discovered that the existing knitr ffmpeg mp4 hook generates output that is not viewable by Safari. After adding the flags recommended in this link (https://apple.stackexchange.com/questions/166553/why-wont-video-from-ffmpeg-show-in-quicktime-imovie-or-quick-preview), things worked for me, so I modified knitr with this small change and re-installed it on my system, using these instructions: https://stackoverflow.com/questions/34800331/r-modify-and-rebuild-package.

With this updated knitr package, I was able to write this blog post (https://brianzhang01.github.io/2018/01/clustering-with-k-means-and-em/) which includes animation that I'm able to view in Safari. I tested on my MacBook Air running macOS Sierra Version 10.12.6, and my iPhone with iOS 10.3.1. I also tested on my MacBook Air with Chrome. I will try to create an example of what things are like if I build without the updated flags.

brianzhang01:
Here is the same post built with knitr 1.20 from CRAN. The animations show up for me in Chrome, but not in Safari. http://www.stats.ox.ac.uk/~bzhang/test/2018/01/clustering-with-k-means-and-em/. Source code is here: https://raw.githubusercontent.com/brianzhang01/blog/master/content/post/2018-01-30-clustering-with-k-means-and-em.Rmd

PR #1592, opened 2018-08-15
Handle output extensions and out_format in patterns

jrnold:
Currently a lot of the logic/data for filetypes is spread throughout the code. Two in particular are how the output name is generated in `auto_output_name` and how the
value of `opts_knit$get('out.format')` is determined. The logic for these decisions depends on the file extension and/or the pattern. This patch streamlines the code to depend on the current pattern to make those decisions.  This avoids having to maintain multiple lists of extensions throughout the codebase.

I added two elements to each pattern in `all_patterns` and the default values of `knit_patterns`.

-  `out.format`: The output format of the pattern, e.g. returned by `knitr:::out_format()`. Defaults to `unknown`.
-  `ext`: A named list. Each element is a recognized extension for that pattern.
   The names are input file extensions and will be used to automatically determine
   the pattern, replacing custom logic elsewhere. These values are used by `detect_pattern()`.  There is still logic in `detect_pattern()` for disambiguating `rtex` and `Rnw` files.

    The value of each element determines how to generate the output file name from the input file name. It can take three types of values.

   - `NA` or `NULL`: Use the default method in `auto_output_name()`. If `_knit_` is in the filename, remove it. Otherwise, if the extension is `.txt` apply the extension `-out.txt`, otherwise use the extension `.txt`.
   - A string: use this extension for the output filename. This replaces the logic
     of what to do with `rmd`, `rmarkdown`, etc. files.
   - A function with a single argument, the name of the input file. This is new and
     isn't used, but adds some flexibility to what patterns can do.

   If no matching `ext` are found for the pattern, then `auto_output_name` will default to the aforementioned defaults. The default is `list()`.

The patch adjusts the code to use these element. The biggest change was in the `knit()` function. It currently determines the output file name (by extension only) before loading a pattern. The code needed to be rearranged so that the pattern is chosen first, and then the output name is created.

The patch passes all the current tests. I should add a few more and complete the documentation if this looks appealing.


jrnold:
Closing open pull requests that I've forgotten about.

PR #1596, opened 2018-08-24
License

tlevine:


DESCRIPTION
```diff
@@ -131,7 +131,7 @@ Suggests:
     showtext,
     tibble,
     styler
-License: GPL
+License: GPL (>= 3)
```
yihui: I don't have time to think about the consequences of this change. From my past experience, some people did want me to license it under GPL-2. Could you leave out this change for now?

tlevine: The issue is that GPL has multiple versions, making simple "GPL" meaningless. If you intend for the license to be GPL 2, I will update all references to use GPL-2.

PR #1601, opened 2018-08-31
Cache fix for non-R engine chunks

nielsrhansen:
I noticed that even though I cached an Rcpp chunk (and no recompilation was done), chunks that were dependent on the Rcpp chunk via the 'dependson' argument were evaluated every time disregarding the caching option. 

The cause, I believe, is in 'block_exec' where the call of 'block_cache' whenever the engine is not R always triggers deletion of the cache for all dependent chunks via 'purge_cache'. 

I have modified the code to check if it's actually necessary to rewrite the cache via a call to 'block_cache' in exactly the same way as it is done further down in the function anyway. This appears to work correctly on the document I am working on with cached Rcpp chunks. I don't know if this will break something for other non-R engines, but I don't see any argument why Rcpp chunks that have not changed should trigger recomputation of dependent chunks.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -111,7 +111,7 @@ block_exec = function(options) {
     output = knit_hooks$get('chunk')(output, options)
     if (options$cache) {
       cache.exists = cache$exists(options$hash, options$cache.lazy)
-      if(options$cache.rebuild || !cache.exists) block_cache(options, output, switch(
+      if (options$cache.rebuild || !cache.exists) block_cache(options, output, switch(
         options$engine,
         'stan' = options$output.var, 'sql' = options$output.var, character(0)
         ))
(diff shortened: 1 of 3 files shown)


R/block.R
```diff
@@ -109,10 +109,13 @@ block_exec = function(options) {
     res.after = run_hooks(before = FALSE, options)
     output = paste(c(res.before, output, res.after), collapse = '')
     output = knit_hooks$get('chunk')(output, options)
-    if (options$cache) block_cache(options, output, switch(
-      options$engine,
-      'stan' = options$output.var, 'sql' = options$output.var, character(0)
-    ))
+    if (options$cache) {
+      cache.exists = cache$exists(options$hash, options$cache.lazy)
+      if(options$cache.rebuild || !cache.exists) block_cache(options, output, switch(
```
yihui: Need a space:

```suggestion
      if (options$cache.rebuild || !cache.exists) block_cache(options, output, switch(
```

PR #1633, opened 2018-11-08
Add a raw_latex function, which uses pandoc's raw_attribute extension…

hughjonesd:
This is just a suggestion. I'm not sure whether it really belongs in knitr or rmarkdown, but it seems close to `asis_output` and `raw_output`. 

Anyway, it would be useful for people who currently get 
\textbackslash{} everywhere when using e.g. xtable or huxtable to print TeX in rmarkdown.


hughjonesd:
Ace, thank you!

PR #1650, opened 2018-12-21
add figure captioning support for shiny.tag/shiny.tag.list objects

cpsievert:
Addresses https://github.com/rstudio/bookdown/issues/74#issuecomment-251561867


cpsievert:
@yihui is this something you might consider merging? Would love to know your thoughts on it.

yihui (commented):
Sorry for the late response. Since we are going to meet at the unconference, let's talk about it there (be sure to remind me if I forget).

PR #1661, opened 2019-01-22
support folding of rendering output when on Travis 

### LINKED ISSUE #1654: Include travis folding markers
When building large or many rmarkdown files on travis (e.g. https://travis-ci.org/hadley/adv-r/jobs/476364254), it is hard to find problems because the log file is thousands of lines long. Travis provides a way to handle this through [folding markers](https://github.com/travis-ci/travis-rubies/blob/9f7962a881c55d32da7c76baefc58b89e3941d91/build.sh#L38-L44) (these are undocumented but used by all build systems so can not change).  

I think it would be very useful if knitr detected that it was being called in travis (by checking the TRAVIS env var), and inserting the folding comments at the start and end of each document.

cderv:
This is a proposition to implement folding to close #1654 

The travis folding feature is not very well documented but I found the crandalf 📦 that implement it so it helped a lot reproduce it. 

Instead of taking the code as is, it just use the part that put the folding part, only if on TRAVIS (`Sys.getenv("TRAVIS") == 'true'`)

Also, crandalf implemented the timer feature to time each folding part. I did not reuse that part, put could be if interested.

This is working (see [example](https://travis-ci.com/cderv/adv-r/jobs/171850729)), however it is really slow to render in the page. Not sure adding timing will be a good idea.

Don't know if something could be done.

Still WIP but you may have thoughts on this one.

---
This change is [Reviewable](https://reviewable.io/reviews/yihui/knitr/1661)

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -964,3 +964,7 @@ digest3 = function(x) {
   close(s)
   unname(tools::md5sum(f))
 }
+
+on_travis <- function() {
+  Sys.getenv('TRAVIS') == 'true'
+}
--- a/R/utils.R
+++ b/R/utils.R
@@ -965,6 +965,12 @@ digest3 = function(x) {
   unname(tools::md5sum(f))
 }
 
-on_travis <- function() {
-  Sys.getenv('TRAVIS') == 'true'
+# if document is knit on travis, add the folding tags.
+# from functions in https://github.com/yihui/crandalf
+travis_fold <- function(type = c("start", "end"), job, msg = NULL) {
+  type = match.arg(type)
+  if (Sys.getenv('TRAVIS') == 'true') {
+    cat(sprintf('travis_fold:%s:%s\r', type, job))
+    if (!is.null(msg)) cat(msg, '\n')
+  }
 }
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -964,3 +964,13 @@ digest3 = function(x) {
   close(s)
   unname(tools::md5sum(f))
 }
+
+# if document is knit on travis, add the folding tags.
+# from functions in https://github.com/yihui/crandalf
+travis_fold <- function(type = c("start", "end"), job, msg = NULL) {
+  type = match.arg(type)
+  if (Sys.getenv('TRAVIS') == 'true') {
```
yihui: I'd like to give users an option to disable folding, e.g., through an environment variable `KNITR_TRAVIS_FOLD` (true/false). Perhaps default to `false` and let users opt in instead of out.

R/utils.R
```diff
@@ -964,3 +964,13 @@ digest3 = function(x) {
   close(s)
   unname(tools::md5sum(f))
 }
+
+# if document is knit on travis, add the folding tags.
+# from functions in https://github.com/yihui/crandalf
+travis_fold <- function(type = c("start", "end"), job, msg = NULL) {
```
yihui: `<-` cough cough 😄

cderv: 😱 What have I done !  I should know better :stuck_out_tongue_winking_eye:

cderv:
Thinking about this, it may be better to insert this in bookdown where the folding is interesting to have when we have several documents. 
I targeted knitr at the beginning maybe not the best place 🤔 
I see what I do with this,taking your comments into account. thanks!

PR #1666, opened 2019-02-03
Implement sass / scss engines

### LINKED ISSUE #1665: Sass / SCSS language engine?
Would there be interest in a PR for a [Sass / SCSS](https://sass-lang.com/) language engine? Sass is a CSS extension language which allows users to write CSS more efficiently / programatically and then convert it back to CSS.

Tactically, this would look similar to a `css` chunk. Where this chunk in RMarkdown:

```
{scss echo = FALSE}
$font-stack:    Helvetica, sans-serif
$primary-color: #00FF00

body
  font: 100% $font-stack
  color: $primary-color
```

Turns into this in-line CSS code:

```
<style type="text/css">
body {
  font: 100% Helvetica, sans-serif;
  color: #99badd;
}
</style>
```

However, behind the scenes, it's slightly more complicated than the `css` engine, since first the code is passed to the `sass` executable, and then that output subsequently receives treatment similar to `eng_html_asset()`. 

---

As I see it, potential pros include:

- since Sass / SCSS "feels" more like a programming language, it might appeal to R users
- similar to the existing `css` engine, this can help code be more self-contained
- a built-in Sass engine ensures code updates are captured (versus having a stand-alone Sass file which users must remember the re-convert so RMarkdown reads in the correct CSS file)

However, I realize it might be unnecessary because:

- Sass is probably overkill for the level of in-line CSS used in most RMarkdowns
- Sass could already be run inside an RMarkdown via the Javascript engine (however, this requires more Javascript knowledge to set-up, and the Sass website states that it is slower) or called from the bash engine (however, this is true of many of the engines!)

Please let me know what you think! Just for fun, I was able to get an SCSS engine to work on Windows, but before I worry more about other systems or some issues specific to the Sass side, I thought I would see if it was of interest.

Thank you for such a great (family of) tool(s)!

emilyriederer:
This PR closes #1665 and adds Sass / SCSS language engines.

When available and not specifically declined by the user (via an `r.sass = FALSE` chunk option), conversion is handled by LibSass via the `sass` R package. Otherwise, conversion is done by the standalone DartSass executable found either via PATH or provided via the `engine.path` option.

Following conversion, output is handled the same as in the `css` language engine.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -659,10 +659,18 @@ eng_go = function(options) {
 }
 
 # SASS / SCSS engine
-## converts SASS / SCSS -> CSS (with same treatments as CSS engine) using either:
-## LibSass sass R package (https://github.com/rstudio/sass) if installed and not sass.package = FALSE, or
-## dart-sass standalone executable (https://sass-lang.com/install)
-## CSS output is compressed by default. Pass "expanded" to engine.opts for more readable output.
+#
+# Converts SASS / SCSS -> CSS (with same treatments as CSS engine) using either:
+# LibSass sass R package (https://github.com/rstudio/sass) when
+#   + the package is installed
+#   + engine.opts does not set package = FALSE (e.g. engine.opts = list(package = FALSE))
+#   + an explicit path to the executable is not provided through engine.path, or
+# dart-sass standalone executable (https://sass-lang.com/install) otherwise
+#
+# CSS output is compressed by default but formatting can be set through style in engine.opts
+#  For the sass R package, valid styles are "compressed","expanded", "nested", and "compact"
+#  For the executable, valid styles are "compressed" and "expanded"
+#  Please refer to respective package / executable documentation for more details
 eng_sxss = function(options) {
 
   # early exit if evaluated output not requested
--- a/R/engine.R
+++ b/R/engine.R
@@ -669,9 +669,9 @@ eng_sxss = function(options) {
   }
 
   # create temporary file with input code
-  f = tempfile( pattern = 'code', tmpdir = '.', fileext = paste0(".",options$engine) )
-  writeLines( options$code , f )
-  on.exit( unlink(f), add = TRUE )
+  f = tempfile(pattern = 'code', tmpdir = '.', fileext = paste0(".",options$engine))
+  writeLines(options$code , f)
+  on.exit(unlink(f), add = TRUE)
 
   # convert sass/sxss -> css
   if( loadable("sass") & !isFALSE(options$r.sass) ){
@@ -680,16 +680,16 @@ eng_sxss = function(options) {
       sass::sass( sass::sass_file(f) ),
       error = function(e) {
        if (!options$error) stop(e)
-       message( paste('Error in converting to CSS using sass R package:', e, sep = "\n") )
+       message(paste('Error in converting to CSS using sass R package:', e, sep = "\n"))
       }
     )
   }
   else{
     cmd = get_engine_path(options$engine.path, options$engine)
     out = tryCatch(
-      paste( system2( command = cmd, args = f, stdout = TRUE) , collapse = "\n"),
+      paste(system2(command = cmd, args = f, stdout = TRUE), collapse = "\n"),
       error = function(e) {
-        if(!openions$error) stop(e)
+        if(!options$error) stop(e)
         message( paste('Error in converting to CSS using executable:', e, sep = "\n") )
       }
     )
--- a/R/engine.R
+++ b/R/engine.R
@@ -660,7 +660,7 @@ eng_go = function(options) {
 
 # SASS / SCSS engine
 ## converts SASS / SCSS -> CSS (with same treatments as CSS engine) using either:
-## LibSass sass R package (https://github.com/rstudio/sass) if installed, or
+## LibSass sass R package (https://github.com/rstudio/sass) if installed and not sass.package = FALSE, or
 ## dart-sass standalone executable (https://sass-lang.com/install)
 eng_sxss = function(options) {
 
@@ -674,8 +674,8 @@ eng_sxss = function(options) {
   on.exit(unlink(f), add = TRUE)
 
   # convert sass/sxss -> css
-  if( loadable("sass") & !isFALSE(options$r.sass) ){
-    message("Converting sass with R package. For executable, set chunk option r.sass = FALSE")
+  if( loadable("sass") & !isFALSE(options$sass.package) ){
+    message("Converting sass with R package. For executable, set chunk option sass.package = FALSE")
     out = tryCatch(
       sass::sass( sass::sass_file(f) ),
       error = function(e) {
@@ -690,7 +690,7 @@ eng_sxss = function(options) {
       paste(system2(command = cmd, args = f, stdout = TRUE), collapse = "\n"),
       error = function(e) {
         if(!options$error) stop(e)
-        message( paste('Error in converting to CSS using executable:', e, sep = "\n") )
+        message(paste('Error in converting to CSS using executable:', e, sep = "\n"))
       }
     )
   }
--- a/R/engine.R
+++ b/R/engine.R
@@ -664,12 +664,12 @@ eng_go = function(options) {
 ## dart-sass standalone executable (https://sass-lang.com/install)
 eng_sxss = function(options) {
 
-  if (!options$eval) {
-    return(engine_output(options, options$code, ''))
-  }
+  # early exit if evaluated output not requested
+  options$results = 'asis'
+  if (!options$eval) return(engine_output(options, options$code, ''))
 
   # create temporary file with input code
-  f = tempfile(pattern = 'code', tmpdir = '.', fileext = paste0(".",options$engine))
+  f = tempfile(pattern = 'code', tmpdir = '.', fileext = paste0(".", options$engine))
   writeLines(options$code , f)
   on.exit(unlink(f), add = TRUE)
 
--- a/R/engine.R
+++ b/R/engine.R
@@ -682,19 +682,25 @@ eng_sxss = function(options) {
     # add sass to Suggests
     sass = get("sass", asNamespace("sass"))
     sass_file = get("sass_file", asNamespace("sass"))
+    sass_options = get("sass_options", asNamespace("sass"))
+
+    style = if(isFALSE(options$sass.compressed)) "expanded" else "compressed"
 
     out = tryCatch(
-      sass( sass_file(f) ),
+      sass(sass_file(f), options = sass_options(output_style = style)),
       error = function(e) {
-       if (!options$error) stop(e)
-       message(paste('Error in converting to CSS using sass R package:', e, sep = "\n"))
+        if (!options$error) stop(e)
+        message(paste('Error in converting to CSS using sass R package:', e, sep = "\n") )
       }
     )
   }
   else{
-    cmd = get_engine_path(options$engine.path, options$engine)
+    message("Converting sass with executable.")
+    cmd = get_engine_path(options$engine.path, "sass")
+    style = if(isFALSE(options$sass.compressed)) "" else "--style=compressed"
+
     out = tryCatch(
-      paste(system2( command = cmd, args = f, stdout = TRUE) , collapse = "\n"),
+      paste(system2( command = cmd, args = c(f, style), stdout = TRUE) , collapse = "\n"),
       warning = function(w) {
         if(!options$error && !is.null(xfun::attr(w, "status"))) stop(paste(w, collapse = "\n"))
         message(paste('Error in converting to CSS using executable:', w, sep = "\n"))
--- a/R/engine.R
+++ b/R/engine.R
@@ -662,6 +662,7 @@ eng_go = function(options) {
 ## converts SASS / SCSS -> CSS (with same treatments as CSS engine) using either:
 ## LibSass sass R package (https://github.com/rstudio/sass) if installed and not sass.package = FALSE, or
 ## dart-sass standalone executable (https://sass-lang.com/install)
+## CSS output is compressed by default. Pass "expanded" to engine.opts for more readable output.
 eng_sxss = function(options) {
 
   # early exit if evaluated output not requested
@@ -672,6 +673,9 @@ eng_sxss = function(options) {
   f = tempfile(pattern = 'code', tmpdir = '.', fileext = paste0(".", options$engine))
   writeLines(options$code , f)
   on.exit(unlink(f), add = TRUE)
+
+  # determine preferred styling
+  style = get_engine_opts(options$engine.opts, engine, "compressed")
 
   # convert sass/sxss -> css
   if(loadable("sass") && !isFALSE(options$sass.package)){
@@ -684,7 +688,7 @@ eng_sxss = function(options) {
     sass_file = get("sass_file", asNamespace("sass"))
     sass_options = get("sass_options", asNamespace("sass"))
 
-    style = if(isFALSE(options$sass.compressed)) "expanded" else "compressed"
+    if(!options$error) stopifnot(style %in% c("nested", "expanded", "compact", "compressed"))
 
     out = tryCatch(
       sass(sass_file(f), options = sass_options(output_style = style)),
@@ -693,11 +697,15 @@ eng_sxss = function(options) {
         message(paste('Error in converting to CSS using sass R package:', e, sep = "\n") )
       }
     )
+
+    # remove final newline chars from output
+    if(!is.null(out)) out = sub("\\n$", "", out)
   }
   else{
     message("Converting sass with executable.")
     cmd = get_engine_path(options$engine.path, "sass")
-    style = if(isFALSE(options$sass.compressed)) "" else "--style=compressed"
+    if(!options$error) stopifnot(style %in% c('expanded', 'compressed'))
+    style = paste0("--style=", style)
 
     out = tryCatch(
       paste(system2( command = cmd, args = c(f, style), stdout = TRUE) , collapse = "\n"),
--- a/R/engine.R
+++ b/R/engine.R
@@ -674,10 +674,17 @@ eng_sxss = function(options) {
   on.exit(unlink(f), add = TRUE)
 
   # convert sass/sxss -> css
-  if( loadable("sass") & !isFALSE(options$sass.package) ){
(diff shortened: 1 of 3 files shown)


emilyriederer:
@cderv 

Let me know if I should go ahead and change the logic to use the `sass` package as suggested by @yutannihilation's [comment](https://github.com/yihui/knitr/issues/1665#issuecomment-460103469) . If so, is there a specific way that you would like this block commented to ensure it gets updated after the `sass` package is release? 

I noticed lines 210-211 contain a `# TODO:` comment for a similar issue related to `reticulate`

R/engine.R
```diff
@@ -658,6 +658,54 @@ eng_go = function(options) {
...
+## LibSass sass R package (https://github.com/rstudio/sass) if installed, or
+## dart-sass standalone executable (https://sass-lang.com/install)
+eng_sxss = function(options) {
+
+  if (!options$eval) {
+    return(engine_output(options, options$code, ''))
+  }
+
+  # create temporary file with input code
+  f = tempfile( pattern = 'code', tmpdir = '.', fileext = paste0(".",options$engine) )
```
cderv: I think it is best to follow the style of the rest of the code here
```suggestion
  f = tempfile(pattern = 'code', tmpdir = '.', fileext = paste0(".", options$engine))
```

R/engine.R
```diff
@@ -658,6 +658,54 @@ eng_go = function(options) {
...
+  writeLines( options$code , f )
+  on.exit( unlink(f), add = TRUE )
+
+  # convert sass/sxss -> css
+  if( loadable("sass") & !isFALSE(options$r.sass) ){
```
cderv: In a condition it is best to use `&&` instead of `&`. 

Also, I think I would name the option beginning with `sass.` instead of `r.` 🤔 
There is already some options with pattern `options$<engine>.<optname>` like `options$python.reticulate` and `options$html.*`. Maybe `sass.package = FALSE` ?

R/engine.R
```diff
@@ -658,6 +658,54 @@ eng_go = function(options) {
...
+       message( paste('Error in converting to CSS using sass R package:', e, sep = "\n") )
+      }
+    )
+  }
+  else{
+    cmd = get_engine_path(options$engine.path, options$engine)
+    out = tryCatch(
+      paste( system2( command = cmd, args = f, stdout = TRUE) , collapse = "\n"),
+      error = function(e) {
+        if(!openions$error) stop(e)
```
cderv:
```suggestion
        if(!options$error) stop(e)
```

R/engine.R
```diff
@@ -658,6 +658,54 @@ eng_go = function(options) {
...
+        message( paste('Error in converting to CSS using executable:', e, sep = "\n") )
+      }
+    )
+  }
+
+  # wrap final output for correct rendering
```
cderv: I think the all part below can be replace by `eng_css()` that calls `eng_html_asset('<style type="text/css">', '</style>')()` (those functions are in `engine.R`

Something like
```r
options$code <- out
eng_css(options)
```
it just a thought reading through the code - I let you check this works.

R/engine.R
```diff
@@ -658,6 +658,54 @@ eng_go = function(options) {
...
+  # create temporary file with input code
+  f = tempfile( pattern = 'code', tmpdir = '.', fileext = paste0(".",options$engine) )
+  writeLines( options$code , f )
+  on.exit( unlink(f), add = TRUE )
+
+  # convert sass/sxss -> css
+  if( loadable("sass") & !isFALSE(options$r.sass) ){
+    message("Converting sass with R package. For executable, set chunk option r.sass = FALSE")
+    out = tryCatch(
+      sass::sass( sass::sass_file(f) ),
```
cderv: Yes I could replace here `::` following the advices in the comments' issue

R/engine.R
emilyriederer: I considered this initially, but in the (extremely rare case) that `echo = TRUE`, this would produce an undesirable result because the "echoed" code would not be the raw user-input anymore but the final CSS. 

Then again, I will admit I can't imagine a case that it would be useful to have `echo = TRUE` for sass (short of someone using RMarkdown to write a sass tutorial? 🤔 ). 

Please let me know what you think - of course, I'm happy to change if you think the value of standardization is greater than the cost of missing this one weird corner case.

cderv: Ok I missed that. I believe it is more important to keep the spirit of knitr above the standardization of code. This is just a few lines of code. 
The other solution would be to modify existing function to make it work for an output that is different as code chunk.
```r
  eng_html_asset = function(prefix, postfix) {
    function(options, out = NULL) {
      if (is.null(out)) out <- options$code
      out = if (options$eval && is_html_output(excludes = 'markdown')) {
        code = c(prefix, out, postfix)
        paste(code, collapse = '\n')
      }
      options$results = 'asis'
      engine_output(options, options$code, out)
    }
  }
```
I do not think it would break something, but not sure there would other case like sass or scss where it would be needed... so we leave it as you proposed ?

R/engine.R
emilyriederer: Makes sense - sorry! I'll clean this up here and other similar places throughout the function.

R/engine.R
emilyriederer: No idea how that snuck in there. Sorry - that's just sloppy on my part. Very embarrassing 😳

R/engine.R
emilyriederer: Sounds good. I'll switch is to `sass.package`. I should have thought of that from the outset.

R/engine.R
emilyriederer: Either way sounds good to me, but I have to agree that I can't think of other cases where this would come up. I erred on the side on only making self-contained changes to the code, but definitely happy to make this edit if you'd like. Just let me know!

emilyriederer:
Thanks @cderv for the review! And apologies -- there are definitely a few things in there that you shouldn't have had to catch. A few follow-up questions:

- I think I have the changes ready now. Do you have a preference how I add them (e.g. more commits versus force pushing an amended version of the initial commit)? Looking back at past PRs in this repo, I couldn't get a strong sense 
- I also have begun to think that it would be better to have the default CSS output be compressed (no whitespace) for the (admittedly minimal) efficiency it provides. It's trivial to add a `sass.compressed` chunk option so that users could set `sass.compressed = FALSE` to get expanded (and more human-readable) output. 
    + Do you agree with the decision for compression to be the default? 
    + Is it worth adding an additional chunk option or is the potential utility too low to justify introducing a new option?
- Thank you for the call-out on `system2` status codes versus R errors. I see that I should have "learned" it from the `eng_interpreted` code. Locally, I have a fix in for that now. I couldn't replicate the specific "dart.exe" error you had on either my Windows 8 or 10, but I am able to induce other status code-related errors so I believe it should work. 

Per the last comment, I'll also include an updated `NEWS.md` with the revised PR.

cderv:
Just a quick reply as it is late but that will keep you going

> I think I have the changes ready now. Do you have a preference how I add them (e.g. more commits versus force pushing an amended version of the initial commit)? Looking back at past PRs in this repo, I couldn't get a strong sense

Do not force push as it is not necessary. When we can avoid I think it is better.  We will deal with that later. You can add more commits to this branch. We should be able to "squash and merge" the PR, in order to reduce the amount of commit. Or I'll deal myself with git stuff at the end - I am knew as a reviewer for this 📦, so we both learn in the process 😄 

> I also have begun to think that it would be better to have the default CSS output be compressed (no whitespace) for the (admittedly minimal) efficiency it provides. It's trivial to add a `sass.compressed` chunk option so that users could set `sass.compressed = FALSE` to get expanded (and more human-readable) output.

If you want to try add this, feel free. I don't think `css` does "compress" the output, nor it is minimal in pandoc template. Do we have a big gain compare to the cost of maintaining ? 
Does sass package and sass exe can both handle this ? Is this a command line argument you add ? 
I don't know if it should be the default, but it could be interesting to pass arguments to sass engine through the `engine.opts` options. Look at this option in [engine.R](https://github.com/yihui/knitr/blob/master/R/engine.R). 
If a user wants to change the default, it just needs knitr option settings.


>    Thank you for the call-out on `system2` status codes versus R errors. In retrospect, this need should have been obvious to me based on the `eng_interpreted` code. I couldn't replicate the specific "dart.exe" error you had on either my Windows 8 or 10, but I was able to induce other status code-related errors to test my fix so I believe it should work now.

Don't worry about `dart.exe`. I know how to reproduce so I could test when you are ready to share code.

emilyriederer:
Thanks for taking the time to reply!

> If you want to try add this, feel free. I don't think `css` does "compress" the output, nor it is minimal in pandoc template. Do we have a big gain compare to the cost of maintaining ?

The `css` engine definitely doesn't naturally. I thought perhaps this could be one other use case for the `scss` engine - to provide normal css and have the whitespace stripped. Admittedly, for most single-file knitr uses, I can't really imagine the performance boost would be noticeable unless someone is doing something particularly long and complicated (at which point it might be a better idea to convert their Sass externally anyway?). However, there's also limited downside. I don't think maintenance is a real problem since its outsourced to either LibSass or Dart Sass, but there is also definitely a readability tradeoff.

> Does sass package and sass exe can both handle this ? Is this a command line argument you add ?

Yes! Both versions can handle it. 

> I don't know if it should be the default, but it could be interesting to pass arguments to sass engine through the `engine.opts` options. Look at this option in [engine.R](https://github.com/yihui/knitr/blob/master/R/engine.R).
> If a user wants to change the default, it just needs knitr option settings.

Thanks - that's a good point. It really does make more sense for this to be an engine option than a chunk option. I'll update this in the code before pushing my new commits.

R/engine.R
cderv: Don't be embarrassed ! it is what reviews are for. 😉

R/engine.R
cderv: > I can't think of other cases where this would come up.

I think you're right. We can go with your current code. Thanks!

cderv:
I let you push your changes and we'll see what default to choose. 🤔

emilyriederer:
Hi @cderv ,

I just pushed a number of changes. Let me know what you think!

I'm still not sure if I love the API for output styling, so definitely open to more feedback here. In the end I considered four options. Happy to switch if you think one of the others is better.

`sass sass.package = FALSE, sass.compressed = FALSE`

Most readable / typeable but felt "aggressive" somehow to add a new chunk option for something so specialized.

`sass sass.package = FALSE, engine.opts = 'expanded'` (**<- how it works now**)

Feels weird to not have the option labeled "style" in anyway, but that is also unneeded since there is only one option. This felt most in keeping with the main usage of `engine.opts` to pass additional command line arguments (although I pass 'expanded' versus '--style=expanded' so the syntax seems reasonable for either the R package or sass implementation.)

`sass sass.package = FALSE, engine.opts = list(style = 'expanded')`

Most explicit but also unnecessarily verbose. 

`sass engine.opts = list(package = FALSE, style = 'expanded')`

Most compact but no longer follows same convention as `python.reticulate` (engine implementation as chunk option)

R/engine.R
```diff
@@ -658,6 +658,79 @@ eng_go = function(options) {
...
+    message("Converting sass with R package. For executable, set chunk option sass.package = FALSE")
+
+    # TODO: after sass R package (https://github.com/rstudio/sass) is released on CRAN
+    # delete calls to get and replace sass, sass_file, sass_options with sass::function_name()
+    # add sass to Suggests
+    sass = get("sass", asNamespace("sass"))
+    sass_file = get("sass_file", asNamespace("sass"))
+    sass_options = get("sass_options", asNamespace("sass"))
+
+    if(!options$error) stopifnot(style %in% c("nested", "expanded", "compact", "compressed"))
```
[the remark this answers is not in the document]
emilyriederer: Thanks -- will do. That's a much better idea!

R/engine.R
```diff
@@ -658,6 +658,79 @@ eng_go = function(options) {
...
+  else{
+    message("Converting sass with executable.")
+    cmd = get_engine_path(options$engine.path, "sass")
+    if(!options$error) stopifnot(style %in% c('expanded', 'compressed'))
+    style = paste0("--style=", style)
+
+    out = tryCatch(
+      paste(system2( command = cmd, args = c(f, style), stdout = TRUE) , collapse = "\n"),
+      warning = function(w) {
+        if(!options$error && !is.null(xfun::attr(w, "status"))) stop(paste(w, collapse = "\n"))
```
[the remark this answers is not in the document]
emilyriederer: Thank you so much for the patient explanation. This makes a lot of sense and was an oversight on my part. It also helped me to realize why my tests intended to check this piece of the code didn't catch the problem. Sorry for your trouble and thank you for the help!

R/engine.R
```diff
@@ -658,6 +658,79 @@ eng_go = function(options) {
...
+      }
+    )
+
+    # remove final newline chars from output
+    if(!is.null(out)) out = sub("\\n$", "", out)
+  }
+  else{
+    message("Converting sass with executable.")
+    cmd = get_engine_path(options$engine.path, "sass")
+    if(!options$error) stopifnot(style %in% c('expanded', 'compressed'))
```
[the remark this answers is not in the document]
emilyriederer: Sounds good - I will change this. 

I think I was naively thinking of `error` is in the R engine, where `error = FALSE` completely stops execution upon errors while `error = TRUE` unabashedly proceeds and just includes the effects / output of errors in the output.

Now that you mention it, it makes a ton more sense for this styling engine to be less "extreme". I like the idea of setting a default much better! 

Your note also made me question my use of `message()` versus `warning()` within my error handling. I'm thinking now that `warning()` would be more appropriate here, also?

emilyriederer:
Hi @cderv,

I've just pushed a number of updates:

- Moved all options into `engine.opts` to be more consistent with `knitr` 
- Updated engine selection so that if `engine.path` is provided, the executable is used regardless of whether or not `package = FALSE` (since clearly this seems to be the user intent)
- Changed handling of `system2()` call to explicitly check for `status`, as in `eng_interpreted()`

While these aren't "official" tests or in the style of `knitr`, I'm also including my local `testthat` script (as txt versus R, per GitHub limitations) in case it is at all helpful. I know there's a lot to look over with all of the sources of variations (package vs executable) x (style) x (error options).  While it doesn't include your specific errors, I am able to trigger some status code issues to test handling there. 

At minimum, hopefully this restores a bit of confidence after I know I've made some pretty stupid mistakes; I'm very grateful for your help and patience! 

[test_eng_sxss.txt](https://github.com/yihui/knitr/files/2898269/test_eng_sxss.txt)

R/engine.R
```diff
@@ -658,6 +658,112 @@ eng_go = function(options) {
...
+  on.exit(unlink(f), add = TRUE)
+
+  # process provided engine options
+  package = options$engine.opts$package %n% TRUE
+  style = options$engine.opts$style %n% "compressed"
+  cmd = get_engine_path(options$engine.path, "sass")
+
+  # validate provided engine options
+  if(!is.logical(package)) {
+    if(!options$error) stop(paste("package option must be either TRUE or FALSE"))
```
cderv: I see you used `warning2` 👍 
Should we use `stop2` also here ? The wrapper with `call. = FALSE`.
As you prefer.

R/engine.R
```diff
@@ -658,6 +658,115 @@ eng_go = function(options) {
...
+      if (!options$error) stop(paste(out, collapse = '\n'))
+      out = NULL
+    } else if (!is.null(out)) {
+      out = paste(out, collapse = "\n")
+    }
+  }
+
+  # wrap final output for correct rendering
+  final_out = if (!is.null(out) && is_html_output(excludes = 'markdown')) {
+    out_tagged = paste(c('<style type="text/css">', out, '</style>'), collapse = "\n")
```
cderv: what is `out_tagged` ? it does not seem to be used. 
Did I miss something ? 🤔

R/engine.R
emilyriederer: Ah! Very nice. I hadn't noticed that wrapper as much in the other engines, but I like it. I will switch over. Thanks for the recommendation!

R/engine.R
emilyriederer: Ah, thanks. You didn't -- I did. Previously I had more validation steps after this, but it appears I failed to remove this assignment when I took those out.

emilyriederer:
Hi @cderv ,

Thanks for the last polishing thoughts! I switched to `stop2` throughout and removed the extraneous variable assignment. Also appreciate the styling changes. I tried to stay consistent to `knitr` but definitely missed some things along the way. (I guess my mental energy there was too focused on repeatedly changing all my `<-`s to `=`s ? 🙂 )

Please let me know if there's anything else I should do on my end, like eventually adding examples to the `knitr_examples` repo, etc. I'll also keep an eye out for `sass`'s CRAN release so we can remove the calls to `get()`.

Thank you again so, so much for your help and great feedback at every step of the way!

cderv:
Awesome thank you ! I'll need to check if I missed anything in the process. It is my first review for knitr 📦 and first time merging - I'll do it very soon. 

I will also add you as contributor to the 📦 in DESCRIPTION. 
Thanks a lot ! This will be a super new feature! 🎉

cderv:
@yihui,  review is done. and this is working well. Just checking how you get things done usually.

From what I observed until now
* I added @emilyriederer as contributor because she wasn't
* I increased the version number in DESCRIPTION as it is new feature

Is this ready to squash and merge ? 

As it is my first review for knitr, I prefer to go through this time to learn the workflow for this 📦 and be at ease next time ☺️ 

Thanks.

R/engine.R
```diff
@@ -136,7 +136,7 @@ eng_interpreted = function(options) {
           Darwin = paste('-q < %s >', shQuote(xfun::normalize_path(logf))),
           Linux = '-q -e do %s',
           '-q -b do %s'
-         ), shQuote(normalizePath(f)))
+        ), shQuote(normalizePath(f)))
```
yihui: Good eye. Thanks!

emilyriederer:
Thanks, @yihui! And don't worry -- I was just joking, not complaining, about `=` and `<-`.  I really didn't mind! After all, a blogger I really admire once [wrote a great post about adhering to package style in PRs](https://yihui.name/en/2018/11/cosmetic-changes/)? 😄 It 100% makes sense, and that's actually the easiest one of all since its so readily apparent from the rest of the code that that is the preference. I just wish I'd done a better job noticing some of the other things that @cderv caught!

Thanks again to you both!

PR #1680, opened 2019-03-10
Fix Stata engine output

Hemken:
Accomodate difference between linux and Windows log files.  It turns out there is one tiny difference between Stata log files on linux versus Windows:  the number of blank lines before the "running ..." note!  This modification gives priority to linux, making it easier to work across platforms.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -93,7 +93,7 @@ engine_output = function(options, code, out, extra = NULL) {
     options$engine
   )
   if (options$engine == 'stata') {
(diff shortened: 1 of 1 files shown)


Hemken:
Thanks!

Hemken:
Thanks for generalizing my code.

Just FYI, I installed your merged knitr and all my Stata engine tests work just fine.

yihui:
Perfect. Thanks!

PR #1683, opened 2019-03-16
added "rmarkdown" engine to process code blocks

dgkf:
I ran into a situation where I wanted to be able to turn on or off different pieces of commentary in the markdown throughout a document. For example, I would ideally like to set a flag to include more technical commentary for an analytic audience, but then strip them out for a general audience.  

Although the `asis` engine works in most cases, it doesn't process inline code blocks that I often use for communicating results. To address this, I added a new `rmarkdown` engine to pre-process the text using `knitr::knit`.

Here is an example document where this is useful:

`````
---
title: "test"
output: pdf_document
---

```{r, echo=FALSE}
# create a flag to easily turn on/off markdown 'code' blocks
.show_analysis_commentary <- TRUE
```

this is some general text

```{r analysis, echo = FALSE}
# derive answer to life, the universe and everything
var <- 42
```

```{rmarkdown, echo = .show_analysis_commentary}
# print out analysis result only if `.show_analysis_commentary` is `TRUE`
the answer to life, the universe and everything is `r sprintf('%.f', var)`
```

```{r}
hist(mtcars$mpg)
```
`````

---
_let me know if any tests should be added. I couldn't find any for the `asis` engine and figured this would only necessitate a similar level of testing._

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -392,6 +392,12 @@ eng_asis = function(options) {
   if (options$echo && options$eval) one_string(options$code)
 }
 
+## output the code, processing as though it's a markdown block
+eng_rmarkdown = function(options) {
+  if (options$echo && options$eval)
+    one_string(knit(text = options$code, envir = knit_global()))
+}
+
 # write a block environment according to the output format
 eng_block = function(options) {
   if (isFALSE(options$echo)) return()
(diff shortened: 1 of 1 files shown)


yihui (commented):
Thanks! I don't have time to think about it carefully at the moment. You should probably call `knit_child()` instead of `knit()` here. Anyway, I guess the problem should not really block you, because there are at least two ways out:

1. You can register the engine by yourself via `knitr::knit_engines$set(rmarkdown = function(options) ...)`;

1. Or use child documents (which will require the commentary to be written to external files, and you may not like that).

simonschoe:
I was exactly looking for this use case! How would I implement it via ``knitr::knit_engines$set(rmarkdown = function(options) ...)`?  When I use `function` from the committed code, R objects that it couldn't find `one_string`.

yihui:
Closing this as stale. As I mentioned above, there are a couple of ways out without adding a new engine (self-registering the engine via `knitr::knit_engines$set()`, or using child documents). Here's a third that may suit you better: conditionally wrap the text in HTML comments — output `<!--` and `-->` around the block to comment it out when your flag is off, and emit nothing when it's on. Thanks for the idea!

Issue #1687, opened 2019-03-23
Minor suggestion (with .diff) which allows to run scripts on the knit()ed .tex


mhofert:
Suggestion for improvement (of ./R/header.R): Replace %% by % on line 126 and 127, see the following `.diff` (which I would have attached as a file if that file type was supported... do you prefer `.diff.zip`?). 

--- header.R	2018-01-02 11:48:28.000000000 -0500
+++ header_new.R	2019-03-23 09:22:03.000000000 -0400
@@ -123,8 +123,8 @@
 '
 
 .header.maxwidth =
-'%% maxwidth is the original width if it is less than linewidth
-%% otherwise use linewidth (to make sure the graphics do not exceed the margin)
+'% maxwidth is the original width if it is less than linewidth
+% otherwise use linewidth (to make sure the graphics do not exceed the margin)
 \\makeatletter
 \\def\\maxwidth{ %
   \\ifdim\\Gin@nat@width>\\linewidth

The reason for this suggestion is that this is (seems to be) the only comment that's introduced in the knitted `.tex` which starts with two '%%' -- and that messes up scripts which grab (= grep) for that (there are various reasons why one would want to run such scripts on the generated `.tex`, I can give more details if necessary).

Minor bug found: The word "Pleaes" in the default text when opening an issue.

yihui:
I'm fine with the change. Would you submit a pull request? Thanks!

PR #1692, opened 2019-03-26
Suggested improvement of LaTeX header to be compatible with scripts on the knit()ed .tex files

mhofert:
see Issue #1687, essentially:

--- header.R	2018-01-02 11:48:28.000000000 -0500
+++ header_new.R	2019-03-23 09:22:03.000000000 -0400
@@ -123,8 +123,8 @@
'

.header.maxwidth =
-'%% maxwidth is the original width if it is less than linewidth
-%% otherwise use linewidth (to make sure the graphics do not exceed the margin)
+'% maxwidth is the original width if it is less than linewidth
+% otherwise use linewidth (to make sure the graphics do not exceed the margin)
\makeatletter
\def\maxwidth{ %
\ifdim\Gin@nat@width>\linewidth

### DIFF
--- a/R/header.R
+++ b/R/header.R
@@ -125,8 +125,8 @@ rm(list = c('.default.sty', '.knitr.sty')) # do not need them any more
 '
 
 .header.maxwidth =
-'%% maxwidth is the original width if it is less than linewidth
-%% otherwise use linewidth (to make sure the graphics do not exceed the margin)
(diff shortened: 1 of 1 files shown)


yihui:
Thanks!

mhofert:
Thanks, Yihui!

PR #1699, opened 2019-04-03
Enable optional features for Perl code chunks

kiwiroy:
This PR aims to enable optional [features](https://perldoc.pl/feature) when running Perl code chunks.

From [perlrun](https://perldoc.pl/perlrun#-E-commandline)

```
  -E commandline
            behaves just like -e, except that it implicitly enables all optional features (in the main compilation unit). See feature.
```

## Current behaviour

An attempt to use `say`, as in `say "Hello, World!";`, is met with the following error.

```
Error in engine(options) : 
  String found where operator expected at -e line 2, near "say "Hello, $user""
	(Do you need to predeclare say?)
syntax error at -e line 2, near "say "Hello, $user""
Execution of -e aborted due to compilation errors.
Calls: <Anonymous> ... process_group.block -> call_block -> block_exec -> in_dir -> engine
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,8 @@
 
 - Fixed a bug introduced when fixing #1648: `include_url()` no longer works at all unless the chunk option `out.extra` is not empty (thanks, @gabrielvazdemelo, #1691).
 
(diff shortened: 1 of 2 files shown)


kiwiroy:
Awesome. `NEWS.md` updated. Thanks.

PR #1701, opened 2019-04-08
R latex pattern

braunm:
This pull request adds a new pattern for defining R code chunks within a LaTeX environment, replacing the <<>>= ... @ syntax.
```
\begin{Rcode}[options]
## code goes here
\end{Rcode}
```
The pattern is associated with files with an .RLatex extension. Ideally, I would like this syntax to be just another option for .Rnw files, but for now, I think this approach is less likely to create conflicts with existing code.

Why did I do this? My development environment is emacs/ESS, and when writing ordinary LaTeX documents, I use the preview-latex emacs package to preview LaTeX math inline. However, preview-latex chokes on R code chunks, and I have not found a way to tell it to skip everything between <<>>= and @. But I can tell preview-latex to skip an environment. So preview-latex thinks the code chunk is just an environment to skip, and knitr knows that the Rcode environment contains code to be knitted. Everyone is happy.

I created a gist that goes into more detail, with examples (including some additions to the LaTeX preamble). The gist also contains a new polymode, so ESS can recognize the contents of the Rcode environment as code, and execute syntax highlighting, indentation, etc., appropriately. That's not part of knitr, but I mention it here for those who are interested.

Of course, I could just define the pattern in a hook to run whenever knitr loads. But that would make my documents non-portable. So I am hoping this contribution is worthy of addition to the package, and that others can benefit.


yihui:
@braunm Are you still interested in getting this merged? If so, the branch now has merge conflicts that would need to be resolved.

Also — before introducing a new `.RLatex` format, does the existing `.Rtex` format work for your use case? It lets you `%` comment out all R code chunks, so you can keep valid LaTeX. I'd rather not add a new format if `.Rtex` already covers this. Thanks!

braunm:
In theory, yes, but TBH I had forgotten all about it.   I have a personal workaround that kind of works, and it's been a couple of years since I've needed it anyway.  I'm not even sure this is the best solution anymore, given more recent versions of emacs and AucTeX. 

Can you keep the issue open, and I could revisit the next time I'm embedding R code in LaTeX?  Or you could close it and I'll just resubmit if it becomes a gating factor.

yihui:
Okay, I'll close it for now (to keep the backlog clean), but it will never be deleted and can be reopened anytime in future. Thanks again!

PR #1739, opened 2019-08-13
Treat out.width|height as CSS when output is html

atusy:
This PR treats chunk options of `out.width` and `out.height` as CSS rather than `img` tag's attributes.

Thus,

````
```{r out.width="50%"}
knitr::include_graphics("foo.png")
```
````

becomes

```html
<img src="foo.png" style="width: 50%;"  />
```

rather than

```html
<img src="foo.png" width ="50%;"  />
```

This update has two benefits:

1. To support any valid CSS units including `px` and `%`
2. To fix a conflict in `rmarkdown::html_document` and its extentions. Currently, `out.height` is ignored in output because `img` tag's `height` attribute has less priority than CSS height defined as:
    
    ````css
    img {height: auto}
    ````

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # CHANGES IN knitr VERSION 1.25
 
-
+- When output format is html, `out.width` and `out.height` are treated as CSS's `width` and `height` properties rather than `img` tag's `width` and `height` tags. This update enables to specify output size with any valid CSS unit (thanks, @atusy).
 
 # CHANGES IN knitr VERSION 1.24
 
--- a/R/hooks-html.R
+++ b/R/hooks-html.R
@@ -34,7 +34,12 @@ hook_animation = function(options) {
 }
 
 .img.attr = function(w, h, extra) {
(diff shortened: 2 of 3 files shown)


atusy:
I just notice a difference of the `%` unit in `img` tag's attributes and CSS properties.
In the `img` tag's attributes, `%` is relative to the raw size of image.
In the CSS properties, `%` is relative to the size of containers.

Thus, this PR is partly a breaking change...

atusy:
I feel I can close this PR because another approach as followings, and because I don't like breaking change and complicated exceptions.

> 1. To support any valid CSS units including px and %

This can be approached by the `out.extra` chunk option (e.g., `out.extra="style='width: 3in;'"` )

> 2. To fix a conflict in rmarkdown::html_document and its extentions. Currently, out.height is ignored in output because img tag's height attribute has less priority than CSS height defined as:

I opened another PR in `rmarkdown`,  and this fixes the conflict.
https://github.com/rstudio/rmarkdown/pull/1620

atusy:
Thanks.
I think this PR is no more necessary, and close it.

Issue #1755, opened 2019-09-04
Support `fig.show='hold'` for office output


atusy:
When output is office formats (e.g., docx) and `fig.show='hold'`, then knitr coerces `fig.show='asis'` with warning.
This is introduced by https://github.com/yihui/knitr/commit/f6e3b62bf3090671fe47282a4f70b635df8368bb#diff-4910cdfe600c3874a655c6951e12daa0 to fix https://github.com/rstudio/bookdown/issues/249 and is expanded by https://github.com/yihui/knitr/commit/575e18e8961e16bee1595cca44ca7b93da006b82#diff-4910cdfe600c3874a655c6951e12daa0 .

However, regardless of the coercion, `fig.show='hold'` seems working.
In addition, removing the coercion still works.
I do not face error reported by https://github.com/rstudio/bookdown/issues/249 

Although I don't know what made things possible, can we remove the relevant lines?
If not, I think we should also add the treatment here: https://github.com/yihui/knitr/blob/master/R/hooks-md.R#L20

## Example Rmd

````
---
title: "Untitled"
output :
  word_document: default
  powepoint_presentation: default
  odt_document: default
  rtf_document: default
  bookdown::word_document2: default
---

# fig.show='hold'

```{r base, fig.show='hold', fig.cap=.cap}
.cap = c('a', 'b')
plot(1)
'a'
plot(2)
```

# fig.show='asis'

```{r ref.label='base', fig.show='asis', fig.cap=.cap}

```
````

BTW, when `fig.show='hold'`, captions are not working in the above example.
I'm currently working to fix this.

##  xfun::session_info('knitr')

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux 9 (stretch), RStudio 1.2.1335

Locale:
  LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8   
  LC_MESSAGES=C              LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C            
  LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

Package version:
[...]

PR #1756, opened 2019-09-05
support holding figures for office formats

atusy:
See discussion in #1755


atusy:
Thanks.
About disappeared captions, I know the reason and will describe later with a PR ; )

PR #1760, opened 2019-09-06
fix disappearance of figure captions when a chunk generates multiple figures continuously

### LINKED ISSUE #1524: Figure Captions with Multiple Figures and echo=FALSE Does not Print the Caption
When printing multiple figures per chunk (and maybe in other scenarios with multiple outputs per chunk), figure captions do not show up unless extra line breaks are inserted between them.  In the example below, the first and third chunks caption as expected, but the second does not.  (This is similar to #820, but it still appears to be an issue.)

    ---
    title: "Test2"
    author: "William Denney"
    date: "March 14, 2018"
    output:
      pdf_document:
        keep_tex: yes
        fig_caption: yes
    ---

    ```{r setup, include=FALSE}
    knitr::opts_chunk$set(echo = TRUE)
    cap_vec <- c("howdy", "foo")
    ```

    # With echo=TRUE

    When `echo = TRUE`, the captions print as expected.

    ```{r fig.cap=cap_vec, echo=TRUE}
    plot(pressure)
    plot(temperature~1/pressure, data=pressure)
    ```

    \clearpage

    # With echo=FALSE

    When `echo = FALSE`, the captions do not print.

    ```{r fig.cap=cap_vec, echo=FALSE}
    plot(pressure)
    plot(temperature~1/pressure, data=pressure)
    ```

    \clearpage

    # With echo=FALSE and new lines

    When `echo= FALSE` and new lines are added between figures via cat, it works.

    ```{r fig.cap=cap_vec, echo=FALSE}
    cat("\n\n")
    plot(pressure)
    cat("\n\n")
    plot(temperature~1/pressure, data=pressure)
    cat("\n\n")
    ```

atusy:
Knitr generates multiple figures in one line when a chunk generates multiple figures continuously.

For example,

````
```{r echo=FALSE, fig.cap = c('cap1', 'cap2')}
plot(1)
plot(2)
```
````

becomes something like

```md
![cap1](plot1.png)![cap2](plot2.png)
```

In such case, Pandoc will suppress figure captions.
Yihui pointed out this in #1756 and I did in #1755 .
These issues were about office formats, but the issue occurs also in HTML and PDF.

I think users expect figure captions be shown rather than figures stay in one line.

This PR fixes this problem by adding blank lines between captioned figures.

```md
![cap1](plot1.png)

![cap2](plot2.png)
```

The current implementation requires `fig.show='hold'` or `echo=FALSE`, which is maybe too strict because

````
```{r, fig.cap = c('cap1', 'cap2')}
g <- ggplot2::qplot(1, 1)
purrr::walk(list(g, g), print)
```
````

is still like

```md
![cap1](plot1.png)![cap2](plot2.png)
```

Unfortunately, I have no idea how to detect such a case.
A workaround is to add blank lines regardless of what is specified to  `fig.show` and `echo`.
A disadvantage of the workaround is that needless blank lines will be added when multiple figures are generated by a chunk but are separated from each other by something, i.e.

````
```{r fig.cap = c('cap1', 'cap2')}
plot(1)
plot(2)
```
````

will be like

````
```r
plot(1)
```

![cap1](plot1.png)



```r
plot(2)
```

![cap2](plot2.png)
````

IMO, For the most of the formats, this is okay because the markdown format is just the intermediate format.
Some may dislike the needless blank lines when they want the markdown format as the output format.

Any thoughts?

## Example Rmd

````
---
title: "Untitled"
output :
  html_document: default
  rtf_document: default
  odt_document: default
  powerpoint_presentation: default
  word_document: default
  pdf_document: default
---

# echo is FALSE

```{r base, fig.cap=.cap, echo=FALSE}
.cap <- c('cap1', 'cap2')
plot(1)
plot(2)
```

# fig.show is hold

```{r ref.label='base', fig.cap=.cap, fig.show='hold'}
```
````

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - The chunk option `collapse = TRUE` now works as expected when the chunk option `attr.*` or `class.*` is provided. By this change, The chunk option `collapse = TRUE` forces `attr.*` and `class.*` be `NULL` except for the chunk options `attr.source` and `class.source` (thanks, @aosavi @cderv @atusy, #1902 #1906).
 
+- Captions disappeared when a chunk generates multiple figures continuously (thanks, @atusy, #1760)
+
 # CHANGES IN knitr VERSION 1.30
 
 ## NEW FEATURES
--- a/R/hooks-md.R
+++ b/R/hooks-md.R
@@ -1,27 +1,37 @@
 #' @rdname hook_plot
 #' @export
 hook_plot_md = function(x, options) {
+  add_lines = function(x) {
+    if (
+      (isTRUE(options$echo) && options$fig.show != 'hold') ||
+      is.null(options$fig.cap) || options$fig.num %in% c(1L, options$fig.cur)
+    ) {
+      return(x)
+    }
+    paste0(x, "\n\n")
+  }
   # if not using R Markdown v2 or output is HTML, just return v1 output
   if (is.null(to <- pandoc_to()) || is_html_output(to))
-    return(hook_plot_md_base(x, options))
+    return(add_lines(hook_plot_md_base(x, options)))
   if ((options$fig.show == 'animate' || is_tikz_dev(options)) && is_latex_output())
     return(hook_plot_tex(x, options))
+
   office_output = to %in% c('docx', 'pptx', 'rtf', 'odt')
   if (need_special_plot_hook(options)) {
     if (is_latex_output()) {
       # Pandoc < 1.13 does not support \caption[]{} so suppress short caption
       if (is.null(options$fig.scap)) options$fig.scap = NA
-      return(hook_plot_tex(x, options))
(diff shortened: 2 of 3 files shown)


atusy:
I see I need some improvements.
I would happy to work on, but want to make sure if we want this solved as this PR has been open for longer than a year.

Current implementation tries to split sub-figures when fig.cap is specified.
However, I now think the split should take place only if the length of `fig.cap` is equal to the number of figures.

Also, in `html_document`, Splitting inlined figures into multiple lines will not show figure captions for each figures.

cderv:
We've been focus on other subject and did not have time to look into it yet. I'll try to look at this issue and suggested fix later this week. Sorry for the delay.

atusy:
Don't worry :) I'm not rush about this.

PR #1772, opened 2019-11-16
change domain for website yihui.name to yihui.org

cderv:
Following https://github.com/rstudio/pagedown/issues/154 this changes the domain of the website in `travis.yml` where redirection are not handled by curl currently

@yihui should we change also anywhere in the source code ?
I am not sure as there is redirection

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
         exit 0;
       fi;
     fi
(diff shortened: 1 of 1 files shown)


cderv:
FWIW knitr has a lot of  `yihui.name` still in its source code 
https://github.com/yihui/knitr/search?q=yihui.name

yihui:
Thanks! I just replace the rest of them.

PR #1775, opened 2019-11-19
Simple path.expand for plot_crop

muschellij2:
Simple `path.expand` so that:
```r
knitr::plot_crop("~/Desktop/myplot.pdf")
knitr::plot_crop("~/Desktop/myplot.png")
```
work.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -280,10 +280,11 @@ fig_process = function(FUN, path, options) {
 #' @return The original filename.
 plot_crop = function(x, quiet = TRUE) {
   ext = tolower(file_ext(x))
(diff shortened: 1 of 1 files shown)


R/plot.R
```diff
@@ -280,6 +280,7 @@ fig_process = function(FUN, path, options) {
 #' @return The original filename.
 plot_crop = function(x, quiet = TRUE) {
   ext = tolower(file_ext(x))
+  x = path.expand(x)
```
yihui: Could you backup the original `x` first? e.g. `x2 = x`, then `return(x2)` below.

Issue #1442, opened 2017-10-07
export 'wrap' for use by other packages defining custom engines?


kevinushey:
Motivation: I'm working on defining a custom Python engine using the `reticulate` package, and I'd like to be able to interleave input and output in a similar way as to what we get when running R code chunks. This implies mimicing the output format produced by `evaluate::evaluate()`, if I understand correctly.

These outputs are generally post-processed using the `knitr` `wrap()` function. Would it be sensible for `knitr` to export this function? Or is there another API that I should be considering here? (Note that `engine_output()` seems to assume that a single block of source code is followed by a single block of output; I'd like to do something a bit more nuanced if possible)

yihui:
I'm happy to export it. The only reason that I'm hesitating is that `wrap()` is a sloppy function name :) Let me think a little more about it.

yihui:
I just extended `engine_output()` so that you can pass a list of the similar structure as `evaluate::evaluate()` output to it, e.g.

```r
engine_output(CHUNK_OPTIONS, out = LIST)
```

where `LIST` can be a list like this:

```r
list(
  structure(list(src = '1 + 1'), class = 'source'),
  '2',
  structure(list(src = 'print("hello world!")', class = 'source'),
  'hello world!'
)
```

Please let me know if that suffices.

kevinushey:
Great! This is exactly what I need. Thanks Yihui!

PR #1799, opened 2020-01-21
export `wrap` generic

kbodwin:
Would you be willing to export this generic?  Happy to provide more info on how I'm using it if you're not sure.  Thank you!

### DIFF
--- a/R/output.R
+++ b/R/output.R
@@ -428,6 +428,7 @@ knit_log = new_defaults()  # knitr log for errors, warnings and messages
 #' @param x output from \code{\link[evaluate]{evaluate}}
 #' @param options List of options used to control output
 #' @noRd
+#' @export
 wrap = function(x, options = list(), ...) {
   UseMethod('wrap', x)
 }


cderv:
Hi, 

I think this request is related to https://github.com/yihui/knitr/issues/1442 and `engine_output` has been adapted with an expert mode in order avoid exporting `wrap`. I guess it should support your usecase already, but happy to discuss it if you can give some details.

See `?knitr::engine_output` 
> For expert users, an advanced usage of this function is engine_output(options, out = LIST) where LIST is a list that has the same structure as the output of evaluate::evaluate(). In this case, the arguments code and extra are ignored, and the list is passed to an internal function knitr:::wrap() to return a character vector of final output.

Please tell us if it seems to fit your usage.
Hope it helps.

kbodwin:
I got that to work, after some fiddling with the options.

It'd still be cool to have wrap exported, I think, for situations like this one where I just want the html wrapping, not necessarily something embedded in an engine.  But I'll leave that up to you all - my problem is solved. :)

Thank you!

cderv:
Good it works for you too !
 
> It'd still be cool to have wrap exported, I think, for situations like this one where I just want the html wrapping, not necessarily something embedded in an engine. 

@yihui should an issue be opened to maybe reconsider this wrap export one day, and have a place to compile the different usage in the meantime maybe ?

yihui:
I could consider exporting it after I see a convincing use case. I guess the S3 generic `knit_print()` (which has been exported) and the function `engine_output()` can do the job in most cases. `wrap()` is too low-level and messy (forgive my R skills 8 years ago), and I always hesitate when users request me to export it.

cderv:
It makes sense. Thank you for the precisions !

PR #1810, opened 2020-02-11
Add propagate option to knit_exit()

davidchall:
When calling `knit_exit()` from within a child document, this child document stops rendering and the parent document continues to render. There are certainly circumstances where this is the desired behavior. However, there are also cases where I'd prefer the entire document to stop rendering immediately.

This PR adds a `propagate` option to `knit_exit()`. This indicates whether the early termination signal should be propagated from child documents up to the top-level document. The default value (`FALSE`) retains existing behavior.


yihui (commented):
Sorry for the late response! I'd like the propagation to be the default behavior. I'm okay with making this breaking change, since I feel it makes a lot more sense for `knit_exit()` to exit the whole process. Thank you!

yihui (approved):
I just simplified the PR and made the breaking change. Will merge after tests pass. Thanks!

davidchall:
@yihui - thanks for reviewing! 

I quite liked the ability to choose the behavior with the `propagate` argument though... The previous behavior could support fault-tolerant documents, when this was desirable.

yihui:
Okay. I just added the argument back. Thanks again!

davidchall:
Thanks!

PR #1812, opened 2020-02-11
Add html_to_r()

### LINKED ISSUE #1811: Extract R code from R Markdown HTML file
There appears to be no fast and easy way to extract the R code from HTML files generated via R Markdown. 

### Example 
Max and Davis's [applied-ml](https://github.com/rstudio-conf-2020/applied-ml) workshop is a good example. 

We can easily get the R code for 'Part_1.html', since we have access to the original `.Rmd` file, and can hence call 

```R
knitr::purl("Part_1.Rmd")
readLines("Part_1.R") %>% paste0(collapse="\n\n") %>% cat
# Displays R code...
```
But we cannot so easily get the R code for parts 2 through 5, as the originating `.Rmd` is not available. 

### Possible solution
`html_to_r()` extracts the R code from R Markdown generated HTML files. 

I provide an implementation in a PR.

### Using in the applied-ml example

We can now easily retrieve the R code from the `.html` files, like so

```R
# from inside applied-ml
dir() %>% grep("Part_{1}.*html", ., value = T) %>% sapply(., html_to_r) -> a
dir() %>% grep("Part_{1}.*html", ., value = T) %>% mapply(html_to_r, inc_out=F, .) -> b

# Randomly inspect the second file with / without output to ensure it worked as expected
a[[2]] %>% cat # with output
b[[2]] %>% cat # without output
```

This can be merged if relevant or disregarded if not relevant.

stevecondylios:
Regards #1811 

- Adds `html_to_r()` to easily extract R code from `.html` files generated through R Markdown
- Adds tests for `html_to_r()`


yihui:
Thanks for this, @stevecondylios, and apologies for the very long delay in getting back to you.

Coming back to this now, I don't think we should merge it, for a few reasons:

1. **It doesn't match what the discussion in #1811 converged on.** There, @atusy proposed a much simpler pandoc-based approach (HTML → commonmark → `purl()`), and you agreed it was "a great simplification and improvement on the DIY solution." That thread also resolved the two open questions — pandoc handles the character-entity decoding for free, and `inc_out` was deemed unnecessary. But this PR is still the original regex-based implementation, not the agreed direction.

2. **The regex approach doesn't work on real R Markdown HTML.** `html_to_r()` searches for Markdown code fences (```` ```r ````, ```` ``` ````) inside `<body>`, but `rmarkdown::render()` produces `<pre class="r"><code>...</code></pre>`, not Markdown fences. The tests here pass because their input is hand-written HTML containing literal fences, which isn't what a knitted document looks like — so on the actual motivating example (the applied-ml `.html` files), nothing would be extracted.

3. **The bare ```` ``` ```` branch also captures non-R output blocks**, which is the ambiguity @atusy flagged.

There's also a new `stringr` dependency, which we try to avoid (knitr leans on `xfun`).

Given that the working approach from #1811 is essentially a short pandoc round-trip that anyone can run ad hoc, and that reconstructing source from rendered HTML is a bit outside knitr's core scope, I'll close this and #1811. If you (or anyone) still wants a helper for this, the pandoc-based `purloc()` sketch in #1811 is the right starting point. Thanks again for the idea and the write-up.

PR #1823, opened 2020-03-23
Add Rust Engine

TianyiShi2001:
Adding support for running Rust code in ```` ```{rust} ```` and ```` ```{cargo} ```` chunks.

```` ```{rust} ```` chunks call `rustc`, while ```` ```{cargo} ```` chunks call `cargo run`. The main difference is that, ```` ```{cargo} ```` chunks allow using external crates, but this also comes at a cost of additional configuration (see below). 

### ```` ```{rust} ```` implementation details:

This is achieved by copying code into a temporary file `foo.rs` and then run `rustc foo.rs && ./foo`.

You may need to specify the path to `rustc` in `~/.Renviron` [because `Sys.getenv('PATH')` in R is different from `echo $PATH` in shell.](https://stackoverflow.com/questions/31121645/rstudio-shows-a-different-path-variable)

Demo:

图片

You can even omit `fn main()` in simple cases:

图片

(this is achieved by first `grep`ping 'fn main', and if it's not found, wrapping your code with `fn main(){` and `}`

### ```` ```{cargo} ```` implementation details:

1. the user sets `options(knitr.cargo.path= "PATH_TO_CARGO")`, where `PATH_TO_CARGO` is the directory that contains `Cargo.toml` **Please suggest a better name for `knitr.cargo.path`!**
2. `eng_rust_cargo()` extracts the cargo directory using `getOption("knitr.cargo.path")`
3. the content of each ```` ```{cargo} ````block is copied to `CARGO_DIR/src/main.rs` (`fn main() {}` added if absent)
4. `setwd()` to the cargo directory, and `cargo run -q` is executed by `system2()`, then wd is reset

Demo:

Screenshot 2020-04-05 at 16 18 03

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -91,6 +91,7 @@ Authors@R: c(
     person("Thibaut", "Assus", role = "ctb"),
     person("Thibaut", "Lamadon", role = "ctb"),
     person("Thomas", "Leeper", role = "ctb"),
+    person("Tianyi", "Shi", role = "ctb", comment = c(ORCID = "0000-0002-9204-9949")),
     person("Tim", "Mastny", role = "ctb"),
     person("Tom", "Torsney-Weir", role = "ctb"),
     person("Trevor", "Davis", role = "ctb"),
--- a/R/engine.R
+++ b/R/engine.R
@@ -650,6 +650,30 @@ eng_go = function(options) {
   engine_output(options, code, extra)
 }
 
+# rust engine, added by @TianyiShi https://github.com/yihui/knitr/pull/1823
+eng_rust = function(options) {
+  src = wd_tempfile('code', '.rs')
+  bin = strsplit(src, '\\.rs')[[1]]
+  write_utf8(source <- options$code, src)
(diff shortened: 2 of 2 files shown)


TianyiShi2001:
Similarly, many other compiled languages can be added to knitr. I’ll try to implement more of them as well as their corresponding cachers.

TianyiShi2001:
A new engine that uses `cargo run` has been added. This allows usage of external crates!

cderv:
FWIW there is a rust knitr engine in the R package **extender**

https://github.com/extendr/rextendr

Not yet on CRAN but part of a rust for R ecosystem. A good home for such an engine.

yutannihilation:
Feedback is welcome! R Markdown-related usages can be found here:

https://extendr.github.io/rextendr/articles/rmarkdown.html

yihui:
The **extendr** package looks much more promising than calling the command directly via `system2()` in R, so I'll close this PR and recommend users to try **extendr** instead. Thanks!

PR #1827, opened 2020-04-02
Update kable to allow custom formating

thebioengineer:
With the coming of vctrs, it is probably inevitable that there will be a number of new vector classes. And with that, a number of custom printing methods. With the current format of kable, it uses `as.matrix()` to handle the conversion, but it might be better to handle the conversion explicitly.

As far as I could test, this would not change the general behavior.


thebioengineer:
With the latest addition, I added a test to confirm custom formatting of non-numeric vectors works.

thebioengineer:
Thanks @yihui! 

I found this while working on [{colortable}](https://github.com/thebioengineer/colortable) and playing around with vctrs. Hoping this will resolve this for everyone!

PR #1829, opened 2020-04-04
Adds a Generic Engine

TianyiShi2001:
A generic engine that make it possible to run any language using custom commands.

Screenshot 2020-04-05 at 05 58 48

This is just a quick-and-dirty implementation (from conception to implementation in one night). If you find this interesting, please post your comments here so I can go ahead and add more features/make it more robust.

PR #1833, opened 2020-04-14
Support for Typescript

TianyiShi2001:
Yet another engine.

图片


yihui:
As I just mentioned in https://github.com/yihui/knitr/pull/1829#issuecomment-1011801414 about the new `exec` engine, now you can use

````
```{exec, command='ts-node'}
```
````

to run TypeScript code. Thanks!

PR #1834, opened 2020-04-14
Support for ragg png device

cpsievert:
cc @thomasp85 @csgillespie (https://github.com/r-lib/ragg/issues/29)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## NEW FEATURES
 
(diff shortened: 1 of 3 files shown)


csgillespie:
Neat! It's probably worth adding the other __ragg__ devices as well, such as jpeg? Also, should __ragg__ be added to Suggests?

cpsievert:
> should ragg be added to Suggests?

It shouldn't be necessary...FWIW, Cairo and svglite are also not in suggests

cpsievert:
@yihui done in 8f5b81a

Issue #1709, opened 2019-05-03
Kable typesets negative numbers incorrectly in tables


Selbosh:
The current version of `kable` generates incorrectly typeset numeric columns, especially in LaTeX output, wherever they contain negative numbers.

Example
---------

Consider this minimal example.
```r
df <- data.frame(x = c('Foo', 'Bar', 'Baz')
                 y = c(1.2345, -6.5432, 1.0001))
kable(df, digits = 2)
```
The output is 
```
|x   |     y|
|:---|-----:|
|Foo |  1.23|
|Bar | -6.54|
|Baz |  1.00|
```
which renders as follows here on GitHub:

|x   |     y|
|:---|-----:|
|Foo |  1.23|
|Bar | -6.54|
|Baz |  1.00|

Negative numbers and minus signs
---------------------------------------

Looks OK, right? Well, not quite. The symbol before the `6` is a hyphen-minus (-, U+002D), not an explicit minus sign (−, U+2212).

In HTML output the difference is not so noticeable to some, but in LaTeX-PDF output it looks extremely dodgy as far as kerning is concerned, because the numbers are incorrectly typeset in text mode and therefore the '-' symbol is misinterpreted as a hyphen rather than a minus sign.

The *correct* output, at least for LaTeX, should be to wrap all numbers in in-line math tags, like 
 `$-123$` or `\(-123\)`, as a hyphen-minus entered in math mode is interpreted and correctly rendered as a minus symbol, with the correct kerning.

Compare these two columns, where column `y` is rendered by the default `kable` settings from a numeric column, and column `z` has been wrapped in math tags.

```markdown
|x   |     y|       z|
|:---|-----:|-------:|
|Foo |  1.23|  $1.23$|
|Bar | -6.54| $-6.54$|
|Baz |  1.00|  $1.00$|
```

![Numeric columns in text mode and in math mode](https://i.postimg.cc/qqNKvVYB/textmode.png)

Possible fix
------------

One way to accomplish this might be to replace the line
https://github.com/yihui/knitr/blob/00ffce24b08f79fc15e2b77309bc0b34a0def647/R/table.R#L145
with
```r
if (is_numeric(x[, j])) x[, j] = sprintf(sprintf('$%%.%sf$', digits[j]), x[, j])`
```
however this may only appropriate for LaTeX output, as though it should look fine on HTML, it would be an unwelcome surprise for people not using MathJax to suddenly need it. Maybe we can make it an option, disabled by default?

Another solution
------------------

[...]

PR #1850, opened 2020-05-20
kable() in LaTeX mode to typeset numeric columns in math mode.

krivit:
This is a partial fix for #1709, addressing the `LaTeX` case by typesetting numeric columns in math mode and wrapping commas (as decimal or thousands separators) in braces (`{}`) for correct spacing. This is, in my opinion, the cleanest option that does not require additional `LaTeX` packages.

I've looked into the other output types (`pandoc`, `html`, `md`, etc.), but I am not quite sure what *should* be their output.

I've updated `NEWS.md` and `DESCRIPTION` to save time, and apologise for the presumption of adding myself as a contributor and adding "thanks" to myself. If this PR doesn't merit a line in the `DESCRIPTION`, I'll be happy to remove it.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
-Version: 1.52.1
+Version: 1.52.2
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666", URL = "https://yihui.org")),
     person("Abhraneel", "Sarma", role = "ctb"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 - For the `tikz` engine, the class options of the `standalone` document classs can be specified via the chunk option `engine.opts$classoption` (thanks, @XiangyunHuang, #1985). The default value is `tikz`, i.e., `\documentclass[tikz]{standalone}` is used by default.
 
+- `kable()` operating in LaTeX mode now typesets numeric columns in math mode for improved rendering of minus signs; decimal and thousands separator commas are wrapped in braces (`{}`) to preserve spacing (thanks, @krivit, #1709).
+
 ## MAJOR CHANGES
 
 - An error is now thrown when an inline code result is not coercible to character. This has always been the assumed behavior but it happens to be different in certain formats with unknown automatic coercion. This is now enforced to prevent any unexpected output. An inline code expression must evaluate to a character vector or an object coercible by `as.character()` (#2006).
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## NEW FEATURES
 
-- `kable()` operating in LaTeX mode now typesets numeric columns in math mode for improved rendering of minus signs; decimal and thousands separator commas are wrapped in braces (`{}`) to preserve spacing (thanks, @krivit, #1709).
+- `kable()` operating in LaTeX mode can now optionally typeset numeric columns in math mode for improved rendering of minus signs, infinite values, and scientific notation; in particular, decimal and thousands separator commas are wrapped in braces (`{}`) to preserve spacing (thanks, @krivit, #1709).
 
 # CHANGES IN knitr VERSION 1.52
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## NEW FEATURES
 
-- `kable()` operating in LaTeX mode can now optionally typeset numeric columns in math mode for improved rendering of minus signs, infinite values, and scientific notation; in particular, decimal and thousands separator commas are wrapped in braces (`{}`) to preserve spacing (thanks, @krivit, #1709).
+- `kable()` operating in LaTeX mode can now optionally typeset numeric columns in math mode for improved rendering of minus signs, infinite values, and scientific notation; in particular, decimal and thousands separator commas are wrapped in braces (`{}`) to preserve spacing. To enable, use `kable(..., numeric.math = TRUE)` or set `options(knitr.table.numeric.math = TRUE)` globally (thanks, @krivit, #1709).
 
 # CHANGES IN knitr VERSION 1.52
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # CHANGES IN knitr VERSION 1.53
 
+## NEW FEATURES
+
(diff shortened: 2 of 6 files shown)


krivit:
Looks like the `TARGET=integration DISPLAY=:99.0` failed with
```
The command "sudo add-apt-repository -y "ppa:marutter/rrutter4.0"" failed and exited with 1 during .
```
I don't think it's anything in the PR.

krivit:
It no looks like the problem is with `knitr-examples`, which makes sense. I will make a PR for that.

krivit:
I've now updated the `knitr-examples` to match and opened a PR there, though I haven't been able to get the [integration tests](https://github.com/yihui/knitr-examples#integration-tests) to work under Travis-CI, since the repository and branch appear to be hardcoded to `yihui/knitr-examples`. @yihui, am I doing something wrong?

yihui:
Thanks for the PR! Typesetting numbers in math mode for better minus signs is a nice improvement!

My main concern is that this changes the **default** LaTeX output: every numeric column now emits `\(...\)` instead of a bare number (e.g. `1.20` → `\(1.20\)`). knitr has 13K+ reverse dependencies on CRAN, and many of them snapshot `kable()` output in their tests or post-process it downstream. A default change this broad is very likely to break some of them, so I'd prefer this be **opt-in** rather than on-by-default.

Could you gate it behind an argument that defaults to off, following the same pattern as the other `kable_latex()` arguments (a formal whose default comes from `getOption()`)? For example:

```r
numeric.math = getOption('knitr.table.latex.numeric.math', FALSE)
```

That way users can flip it per-call via `kable(..., numeric.math = TRUE)` or globally via `options(knitr.table.latex.numeric.math = TRUE)`, and the default output stays unchanged for existing users. Compare `toprule`, `midrule`, etc., which already work this way.

Two smaller things:
- The new test uses `format.arg=` — should be `format.args=`; it only works today via partial argument matching.
- `NA` cells get wrapped as `\(NA\)`. Worth skipping the wrap for `NA` (and any non-numeric-looking token) so text doesn't land in math mode.

Also, the branch currently conflicts with `master` — only in `NEWS.md` (the `VERSION 1.53` header); `R/table.R` and the tests merge cleanly. Please merge master and resolve that one header.

krivit:
Thanks for looking into this! Happy to make all these changes. While we are at it, would it make sense to also pretty-print the scientific notation, i.e., replace `"2e20"` with `"2 \\times 10^{20}"`?

krivit:
Also, `Inf` and `-Inf` should be handled as well.

krivit:
@yihui , I've pushed changes addressing the comments. For consistency with the argument name, I went with option name `knitr.table.numeric.math`.  I ended up creating a helper function `transform_num_for_latex()` that takes a string containing a typeset number (such as that received by `kable_latex()`) and re-typesets it for LaTeX using string operations.

In the longer term, a similar function can be made for Markdown and HTML output.

yihui:
Two follow-ups on the latest changes:

**Reuse `format_sci()`.** We already have `format_sci()` / `format_sci_one()` in `R/utils.R`, used by the inline hooks for all output formats. It handles infinities, the `scipen` threshold, the `1 \times 10^n` -> `10^n` simplification, and the same `knitr.inline.times` option — so `transform_num_for_latex()` largely duplicates it. The catch is that `format_sci()` takes **numeric** input while your helper takes already-formatted strings, so reusing it would mean applying it on the numeric column before stringification rather than after. Bonus: reusing it means eventual HTML/Markdown support falls out of the existing `format` argument instead of needing a second helper.

**Scientific notation.** I'd rather keep this PR focused on the math-mode typesetting only — math-mode wrapping, minus signs, comma-bracing, and Inf/-Inf. Rendering `1.23e6` as `1.23 \times 10^6` changes the number's *appearance*, not just its typesetting, and deserves its own discussion. Could you split the scientific-notation conversion out into a separate PR (reusing `format_sci()` there)? That keeps this one small and easy to reason about for the revdep impact.

krivit:
> **Reuse `format_sci()`.** We already have `format_sci()` / `format_sci_one()` in `R/utils.R`, used by the inline hooks for all output formats. It handles infinities, the `scipen` threshold, the `1 \times 10^n` -> `10^n` simplification, and the same `knitr.inline.times` option — so `transform_num_for_latex()` largely duplicates it. The catch is that `format_sci()` takes **numeric** input while your helper takes already-formatted strings, so reusing it would mean applying it on the numeric column before stringification rather than after. Bonus: reusing it means eventual HTML/Markdown support falls out of the existing `format` argument instead of needing a second helper.

This might require quite a bit of rearchitecting:

1. The mathification needs to happen after the escaping of special characters, and the escaping needs to happen after the conversion to string, and also the escaping depends on the output format, which means that there isn't really a clean and simple way to separate format-specific parts from format-agnostic.

    Probably the solution with minimal change would be for the escaping code to skip the numeric columns. But then...

2. `format_sci()` formats each element in a vector individually, whereas `format_args()` that's used by `kable()` formats the whole column jointly. This is important for tables, because we generally want to have consistent number of digits after the decimal point and consistent use of scientific notation within each column:

    ``` r
    x <- c(1.1, 1.11)
    knitr:::format_args(x)
    #> [1] "1.10" "1.11"
    knitr:::format_sci(x)
    #> [1] "1.1"  "1.11"
    
    y <- c(1.1e-10, 1.11)
    knitr:::format_args(y)
    #> [1] "1.10e-10" "1.11e+00"
    knitr:::format_sci(y)
    #> [1] "1.1\\times 10^{-10}" "1.11"
    ```

    So, the `format_sci()` family would need to be rearchitected as well.

Though I am not very happy with the regex approach either, it has the benefit of outsourcing the scientific notation and the trailing zero decisions to `format()`.

> **Scientific notation.** I'd rather keep this PR focused on the math-mode typesetting only — math-mode wrapping, minus signs, comma-bracing, and Inf/-Inf. Rendering `1.23e6` as `1.23 \times 10^6` changes the number's _appearance_, not just its typesetting, and deserves its own discussion. Could you split the scientific-notation conversion out into a separate PR (reusing `format_sci()` there)? That keeps this one small and easy to reason about for the revdep impact.

I can split it out, but I would rather keep it in the same patch. If we keep the regex approach, it's just another branch in the `ifelse()` in the `transform_num_for_latex()`. If we get `format_sci()` to work for this case, the patch will produce the same change anyway via `format_sci()`, unless a separate code-path is created to **prevent** scientific notation from being typeset in tables.

You are the maintainer, so I defer to you on the revdep issues, but the scientific notation typesetting will only kick in if `numeric.math == TRUE` anyway, so revdeps shouldn't be affected either way.

yihui:
Good points and largely agreed. Thanks! I pushed one more fix: `0e+00` (which `format()` produces for zero in a column that uses scientific notation) now renders as `\(0\)` instead of `\(0\times 10^{0}\)`.

Issue #1876, opened 2020-07-15
fix_options should run after chunk hooks


atusy:
I set chunk options and hooks so that I can specify `fig.width` and `fig.height` in centimeters.
What I did is to divide `fig.width/height` by 2.54.
In order to keep the default size of figures, I also set chunk options to multiply default `fig.width/height` by 2.54.
However, I realized that `out.width` is calculated before dividing `fig.width/height` by 2.54 in `rmarkdown::html_document`.

It happens when `keep_md` is `FALSE`.
I read the source code and found that `fix_options` is calculating `out.width` before applying chunk hooks on `fig.width`.
I suggest run `fix_options` after chunk hooks, though I'm not sure about side effects...

When `keep_md` is TRUE, `fig.retina` becomes `NULL` and avoids the problem.
https://github.com/rstudio/rmarkdown/blob/ff1b279e795b62b1ffeabcc9aa5bf7386a7ebb83/R/html_document.R#L475-L476

I will draft a PR and see if it passes the test.

# Reprex

```r
rmarkdown::render('example.Rmd')
```

<details><summary>example.Rmd</summary>

````md
---
title: "Untitled"
output: html_document
---

```{r}
f = knitr::knit_hooks$get('plot')
knitr::knit_hooks$set(
  plot = function(x, options) {
    print(options$out.width)
    f(x, options)
  }
)
```


```{r before-hook}
knitr::opts_current$get()[c('fig.height', 'fig.width')]
plot(iris)
```

```{r hook}
knitr::opts_chunk$set(
  fig.height = knitr::opts_chunk$get('fig.height') * 2.54,
  fig.width = knitr::opts_chunk$get('fig.width') * 2.54
)

knitr::opts_hooks$set(
  fig.height = function(options) {
    options$fig.height <- options$fig.height / 2.54
    options
  },
  fig.width = function(options) {
    options$fig.width <- options$fig.width / 2.54
    options
  }
)
```



```{r after-hook}
knitr::opts_current$get()[c('fig.height', 'fig.width')]
plot(iris)
```
````

</details>

<details><summary>
Result: out.width differs before (672) and after (1706.88) setting chunk hooks
</summary>

````
rmarkdown::render('example.Rmd')
Loading knitr


processing file: example.Rmd
  |..........                                                                                    |  11%
  ordinary text without R code

  |.....................                                                                         |  22%
label: unnamed-chunk-1
  |...............................                                                               |  33%
  ordinary text without R code

[...]

PR #1877, opened 2020-07-15
run fix_options after hook

atusy:
solves #1876


atusy:
rmarkdown also passes test when installing knitr from this PR
https://github.com/atusy/rmarkdown/pull/1

yihui (approved):
That makes sense. Thank you!

Issue #1708, opened 2019-04-28
cache is rebuild when it should not be


fountainer:
A minimal example is as follows:
```
\documentclass{article}

<<setup, echo = FALSE>>=
knitr::opts_chunk$set(
  cache = TRUE,
  autodep = TRUE
)
@

\begin{document}
<<chunk1>>=
n <- 6
print(seq_len(n))
print(Sys.time())
@

<<chunk2>>=
n <- 10
print(seq_len(n))
print(Sys.time())
@

\end{document}
```
Run the code once, the results are cached. When I change the code in the first chunk, say `n <- 10` and rerun the code, the chunk2 is also rebuild. I expect that the chunk2 will not be rebuild since it should not depend on the `n` in the first chunk.

Session information is:
```
R version 3.5.2 (2018-12-20)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Mojave 10.14.4

Locale: C

Package version:
  evaluate_0.13   glue_1.3.0      grDevices_3.5.2 graphics_3.5.2 
  highr_0.7       knitr_1.21      magrittr_1.5    markdown_0.9   
  methods_3.5.2   mime_0.6        stats_3.5.2     stringi_1.4.3  
  stringr_1.4.0   tools_3.5.2     utils_3.5.2     xfun_0.5       
  yaml_2.2.0  
```

I have posted the issue in [stackoverflow](https://stackoverflow.com/questions/55883038/why-cache-is-rebuild-when-it-should-not-be).

yihui:
I just posted a reply on StackOverflow. Thanks for the report!

Issue #1403, opened 2017-05-17
Cache dependencies


knokknok:
Not sure if it is an issue with codetools or knitr (and where it should/could be "fixed") but:

```r
codetools::findGlobals(function(){x<-2*x})
```

doesn't return `x`.

The side effect is that, in the following, the "b" chunk does not depend on the "a" chunk. So changing the assignment in the "a" chunk will always report a result of `2`.

```r
<<knitr-config, echo=FALSE, message=FALSE, results='hide'>>=
opts_chunk$set(cache=2, echo=FALSE, results='hide', autodep=TRUE)
@

<<a>>=
x <- 1
@

<<b>>=
x <- 2*x
@

<<res, results='asis'>>=
cat(x)
@

<<cache-dep, cache=FALSE, results='asis'>>=
cat('\n%\n% Chunk dependencies:\n')
deps <- sapply(knitr:::dep_list$get(), paste, collapse=', ')
cat(paste0('% ', names(deps), ": ", deps), sep='\n')
@
```

yihui:
Since this issue comes from **codetools**, I think you should report it to bugs.r-project.org and probably let Luke know. There is no need to mention **knitr** when you report the issue. As long as this upstream issue can be fixed, **knitr** should work automatically. Thanks!

knokknok:
Looks like something equivalent was already reported one year ago...
https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16884

yihui:
Okay, then I'll wait for it to be fixed in base R, unless you can come up with a small patch, in which case I'll be happy to incorporate it in knitr.

yihui:
The ticket in R was closed, which makes sense. You will have to manually specify the dependency via `dependson='a'` in `<<b>>=`. knitr's documentation also said that automatic detection of dependencies is not reliable: https://yihui.name/knitr/options/#cache

knokknok:
Would the following work? I think it's "erring on the 'there are dependencies' side" ;-)

```diff
 find_globals = function(code) {
-  fun = eval(parse_only(c('function(){', code, '}')))
-  setdiff(codetools::findGlobals(fun), known_globals)
+  unique(subset(getParseData(parse(text = paste('function(){', paste0(code, collapse="\n"), '}', sep="\n"), keep.source=TRUE), includeText=NA), terminal & token %in% c("SYMBOL", "SYMBOL_FUNCTION_CALL"))$text)
 }
-known_globals = c(
[...]

PR #1882, opened 2020-08-03
new cache.strict chunk option

knokknok:
Allow the user to chose between preferring false positives or false negatives for cache invalidation (#1403 vs #1708).


yihui (commented):
Instead of introducing a new chunk option (which will break existing cache), I prefer overloading `options$cache.globals`: In theory, it should take a character vector. We can use `find_globals()` when it is `TRUE`, and `find_symbols()` when it is `FALSE`. What do you think?

Issue #1767, opened 2019-10-23
Could the label prefix in kable() be user defined?


AnttiSim:
If I make a tex file of a table with the `kable` function, it adds a prefix `tab:` to the label by default.

I find this a bit awkward, because I have accustomed to use a different label myself: `TAB:`. It is confusing to have tables from two different sources with the labeling either being upper or lower case depending on its source. Easiest would be if the `kable` argument `label` would include the whole user defined label, including the prefix that suits the user best.

This is just a small suggestion for improvement, and I would be surprised if noone had thought about it before.

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('knitr')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('yihui/knitr')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

luispfonseca:
Also struggling with this, as I am using pandoc-crossref, which uses tbl, making it impossible (I think) to cross-ref properly tables coming from kable.

yihui:
Done with #1890 (thanks, @luispfonseca).

```r
remotes::install_github('yihui/knitr')
```

Please see the NEWS item in 3f58f675605e65ad8dda632d405fabef101a81f7.

Note that for table cross-references in **bookdown**, it currently only supports `tab:`. If you customize it, you won't be able to cross-reference tables in **bookdown**.

PR #1890, opened 2020-08-26
allows for a custom label for a table (#1767)

luispfonseca:
A first attempt at a fix for #1767

I don't know if this is something that would interest you but in my case it appears to be necessary to make kable compatible with pandoc-crossref using a latex export. It doesn't work right yet with html, but I am willing to give it a go if you are interested in eventually merging this.


R/table.R
```diff
@@ -96,7 +97,8 @@
 #' kables(list(kable(d1, align = 'l'), kable(d2)), caption = 'A tale of two tables')
 kable = function(
   x, format, digits = getOption('digits'), row.names = NA, col.names = NA,
-  align, caption = NULL, label = NULL, format.args = list(), escape = TRUE, ...
+  align, caption = NULL, label = NULL, label_prefix = 'tab:',
```
atusy, line 100: I suggest `label_prefix` look for its value from `getOptions` or `knitr::opts_knit$get` unless it is explicitly specified.
I think most users want the same `label_prefix` for all tables.

It would be also nice if we could do something like below so that we can also change prefix for other elements in the future.

`knitr::opts_knit$set(label_prefix = c(table = 'tab:', figure = 'fig:'))`

luispfonseca: I agree with the suggestion. Just made a push making a step in that direction.

atusy: Great.
I just noticed that `label.prefix` might be better than `label_prefix` in terms of consistency with names of the other options.
I'd like to hear the voice of @yihui about this.

R/defaults.R
```diff
@@ -169,6 +169,7 @@ opts_knit = new_defaults(list(
   upload.fun = identity, global.device = FALSE, global.par = FALSE,
   concordance = FALSE, documentation = 1L, self.contained = TRUE,
   unnamed.chunk.label = 'unnamed-chunk', highr.opts = NULL,
+  label_prefix = c(table = 'tab:', figure = 'fig:'),
```
yihui, line 172:
```suggestion
  label.prefix = c(table = 'tab:', figure = 'fig:'),
```

R/defaults.R
```diff
@@ -181,7 +182,7 @@ opts_knit = new_defaults(list(
 opts_knit_names = c(
   'progress', 'verbose', 'upload.fun', 'animation.fun', 'global.device',
   'eval.after', 'concordance', 'documentation', 'aliases', 'self.contained',
-  'unnamed.chunk.label'
+  'unnamed.chunk.label', 'label_prefix'
```
yihui, line 185:
```suggestion
  'unnamed.chunk.label', 'label.prefix'
```

R/table.R
```diff
@@ -160,7 +160,8 @@ kable_caption = function(label, caption, format) {
   # create a label for bookdown if applicable
   if (is.null(label)) label = opts_current$get('label')
   if (!is.null(caption) && !is.na(caption)) caption = paste0(
-    create_label('tab:', label, latex = (format == 'latex')), caption
+    create_label(opts_knit$get("label_prefix")[["table"]],
+                 label, latex = (format == 'latex')), caption
```
yihui, line 164:
```suggestion
    create_label(
      opts_knit$get('label.prefix')[['table']],
      label, latex = (format == 'latex')), caption
```

yihui (approved):
I renamed `label_prefix` to `label.prefix` as @atusy suggested. Thanks!

R/table.R
```diff
@@ -160,7 +160,9 @@ kable_caption = function(label, caption, format) {
   # create a label for bookdown if applicable
   if (is.null(label)) label = opts_current$get('label')
   if (!is.null(caption) && !is.na(caption)) caption = paste0(
-    create_label('tab:', label, latex = (format == 'latex')), caption
+    create_label(
+      opts_knit$get('label.prefix')[['table']],
+      label, latex = (format == 'latex')), caption
```
yihui, line 165:
```suggestion
    create_label(
      opts_knit$get('label.prefix')[['table']],
      label, latex = (format == 'latex')
    ), caption
```

yihui (commented):
Another cosmetic change

luispfonseca:
I am no sure if the Contributor License Agreement is a roadblock here, but I have to admit I don't feel comfortable with all the permissions requested, including the ability to read gists, including private ones. Is it possible to agree to the license without having to share personal data?

Thank you.

yihui:
Yes, it suffices to just verbally agree to the CLA here if you have read the content of the CLA. Thanks!

luispfonseca:
I agree to the CLA. Thank you @yihui

cderv:
> I am no sure if the Contributor License Agreement is a roadblock here, but I have to admit I don't feel comfortable with all the permissions requested, including the ability to read gists, including private ones. Is it possible to agree to the license without having to share personal data?

@luispfonseca there should not be all this required to you only to sign it. Those permissions are asked when setting up cla-assistant in an organization or repo. So for us. You shouldn't need to sign in to their app only to sign. For example, this link should not asked you for permission in your Github account: https://cla-assistant.io/yihui/knitr

We'll investigate. Sorry for the trouble.

Issue #1881, opened 2020-08-01
Reconsider interaction between knitr and evaluate


hadley:
I've found it challenging to create custom output formats that want to care deeply about syntax highlighting and code display because the responsibilities of evaluate and knitr are intertwined in a way that makes it difficult to control the details. I think it might be worth reconsidering exactly how these two packages interact, with an eye to moving more code into evaluate so it could become responsible for the complete "run code and display it as if you had run from the console, but in any output format". This would include moving the code responsible for rendering plots to files from knitr to evaluate, and rethinking the extensibility mechanisms. I think this could make it easier to write new output formats, and would allow the code to be re-used in other non-document situations (e.g pkgdown, reprex) where you also want to run code and display the results to the user.

dmurdoch:
This is somewhat related:  in the `rgl` package devel version, I've needed to use the `evaluate` hook to distinguish between "high level" and "low level" plot actions, the way `knitr` does for base graphics.  (There's some tricky code in `knitr:::is_low_change` that figures out whether an R statement drew a base graphics plot that just adds to an old one; if so, it doesn't include the old one, only the newer one.  I didn't see a way for `rgl` to use the existing mechanism, so I basically duplicated it.)  Since the evaluate hook is undocumented as far as I can see, I'm not too confident about the resilience of this code.

This is related to @hadley's comment, in that if code evaluation and display is changed, it is likely that the processing of output will also end up changed (or at least it could be).   It would be nice to be able to specify a class on an output value that can pass the tests in `evaluate::is.recordedplot` and `knitr:::is_low_change` so that I wouldn't need to duplicate all that code.

yihui:
Thanks both. @dmurdoch's part is largely handled: since #1892, `is_low_change()` is an exported S3 generic (and `wrap()` is exported too), so rgl registers its own method rather than duplicating the logic.

The broader refactor — moving plot rendering and code display out of knitr into evaluate — sounds like a large project. I'll close this for now, but happy to reopen: @hadley, if you can share more about the specific challenges you hit and which extension points felt hardest to control, that'd help scope it concretely.

Issue #1853, opened 2020-05-23
Suggestion:  export `knitr:::wrap`


dmurdoch:
In #1852, I described trouble creating a `knit_print` method for the result of `rglwidget()`, because it's an `htmlwidget` descendant, and they are handled by the generic `knit_print` method, rather than delegating to a method.  In the meantime, I think I've found a better solution to the underlying problem, but this solution requires use of the internal function `knitr:::wrap`, so I am suggesting `wrap` should be exported.

Here is the issue:  I would like `rgl` outputs to be automatically included in `knitr` documents, using the same rules as standard R graphics, i.e. respecting the different settings for `fig.keep` in the chunk options.  With the current `knitr` code, I believe the following is a description of how graphics are normally included:

- all graphics operations (both low level and high level) result in `grDevices::recordPlot` being called to record the plot state.   These are stored in a list with the echoed R code, text output, etc., in objects with class `"recordedplot"`.
- after fully processing the code chunk, `knitr` goes through the list and removes things that shouldn't be displayed. 
- It then calls the internal `wrap()` on the remaining items to produce the final text to include in the output. 

So if I have `fig.keep = "high"` (the default), the intermediate low level plot changes are removed in the second step, and `wrap()` is only called after the last change to each plot.

I can mimic this in `rgl` by having a `knit_print` method that creates an object descending from `"recordedplot"` but containing data that can be turned into an `rglwidget()` result.  However, if I don't have a `wrap()` method for my object, the standard one will be called and will give an error because my object can't be replayed the way a standard `"recordplot"` object can.  So I need a `wrap()` method, which is problematic because `wrap()` is not exported.  I can still create it using undocumented tricks, but it's bad practice to use package internals, and CRAN would justifiably reject the change.

Here's my current code to implement this, which appears to work, but uses `knitr:::wrap`:

```
knit_print.rglId <- local({
  plotnum <- 0
  function(x, options, ...) {
    if (getOption("rgl.printRglwidget", FALSE))	{
      scene <- scene3d()
      args <- list(...)
      if (inherits(x, "rglHighlevel"))
        plotnum <<- plotnum + 1
      structure(list(plotnum,  # knitr uses this to decide whether to suppress the plot
                     scene = scene,
                     width = figWidth(),
                     height = figHeight(),
                     options = options, args = args),
                class = c("rglRecordedplot", "recordedplot"))  
    }
  }
})

wrap.rglRecordedplot <- function(x, options = list(), ...) {
[...]

PR #1892, opened 2020-09-01
Changes to allow rgl (and other?) graphics systems to work like base graphics

dmurdoch:
This patch is related to my comments on #1881 and my issue #1853, to allow `rgl` to work more seamlessly with `knitr`.  The idea is that, like base graphics, `rgl` has "high level" and "low level" commands to change the current display, and users typically only want to see output when a new high level command is run.

To do this, I needed to export two internal functions from `knitr`:  `wrap()` and `is_low_change()`.  I also needed to change `is_low_change()` to a generic function, so the test for low versus high could take place in `rgl`, and I introduced a new class name, `"knit_other_plot"` so that `knitr` could tell that my `rgl` graphics objects are actually plots that need handling.  The access to `wrap()` allows `rgl` graphics to efficiently share data:  if part of a plot has previously appeared, it won't be written to the HTML output.

This PR shouldn't have bad effects on anyone as far as I know.  The current development version of `rgl` on R-forge doesn't take advantage of the changes; it uses a complicated (and fragile, I worry) set of hooks to replace a bunch of code in `knitr`.  You can see the code that does work with this PR in https://github.com/dmurdoch/rgl branch `rglpatch`.


dmurdoch:
Travis seems confused:  some checks have been sitting there all day, stuck at "Queued -- Build Created".

dmurdoch:
Similar changes to these were merged into `pkgdown` in https://github.com/r-lib/pkgdown/pull/1548 and https://github.com/r-lib/downlit/pull/78 .  Some differences:

- In this PR, I exported `wrap` and `is_low_change`, but in the pkgdown changes the corresponding functions (`downlit:::replay_html` and `downlit:::is_low_change`) weren't exported. 

- The new class name `"knit_other_plot"` was called `"otherRecordedplot"`.

If you'd like one or both of those changes in this PR, I'd be happy to make them.

yihui (approved):
Looks good to me overall. I'll make some minor and doc changes before merging. Thanks!

dmurdoch:
Great!

PR #1896, opened 2020-09-22
Improved SQL dbExecute vs dbGetQuery

rnorberg:
Adds additional SQL "update" keywords (`ALTER`, `GRANT`, `MERGE`, etc) to `is_sql_update_query` (which is a package internal function and not exported) and allow users to force a query contained in a SQL chunk to be dispatched to `DBI::dbExecute` or `DBI::dbGetQuery` via the new `sql.is_update_query` chunk option.

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -519,7 +519,7 @@ is_sql_update_query = function(query) {
     # comes out looking like "^\\s*(CREATE|ALTER|...).*"
     "^\\s*(",
     paste(sql_update_keywords, collapse = "|"),
-    ".*"
+    ").*"
   )
   grepl(pattern, query, ignore.case = TRUE)
 }
--- a/R/engine.R
+++ b/R/engine.R
@@ -575,7 +575,7 @@ eng_sql = function(options) {
   query = interpolate_from_env(conn, sql)
   if (isFALSE(options$eval)) return(engine_output(options, query, ''))
 
-  if (is_sql_update_query(query)) {
+  if (options$sql.is_update_query || is_sql_update_query(query)) {
     DBI::dbExecute(conn, query)
     data = NULL
   } else if (is.null(varname) && max.print > 0) {
--- a/R/engine.R
+++ b/R/engine.R
@@ -492,7 +492,36 @@ is_sql_update_query = function(query) {
   query = gsub('^\\s*--.*\n', '', query)
   # remove multi-line comments
   if (grepl('^\\s*\\/\\*.*', query)) query = gsub('.*\\*\\/', '', query)
-  grepl('^\\s*(INSERT|UPDATE|DELETE|CREATE|DROP).*', query, ignore.case = TRUE)
+  sql_update_keywords <- c(
+    # DDL
+    "CREATE",
+    "ALTER",
+    "DROP",
+    "GRANT",
+    "DENY",
+    "REVOKE",
+    "ANALYZE",
+    "AUDIT",
+    "COMMENT",
+    "RENAME",
+    "TRUNCATE",
+    # DML
+    "INSERT",
+    "UPDATE",
+    "DELETE",
+    "MERGE",
+    "CALL",
+    "EXPLAIN PLAN",
+    "LOCK",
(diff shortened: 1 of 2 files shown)


dpprdan:
I've also run into this issue. 

<details>
<summary>Details</summary>

I.e. the DB driver emits a warning, when you have a SQL statement without one of the current keywords inside a sql chunk. E.g. RPostgres [emits](https://github.com/r-dbi/RPostgres/blob/c68d38f3b721d34ba9b190ce5c796a0100a056b2/src/PqResultImpl.cpp)

```r
In result_fetch(res@ptr, n = n) :
  Don't need to call dbFetch() for statements, only for queries
```

Other driver may emit warnings with different wordings, but they are essentially the same, e.g. [RSQLite](https://github.com/r-dbi/RSQLite/blob/ec96c48d4ec4f1e438ff013c40e6dbd31948559d/src/SqliteResultImpl.cpp#L309).

</details>

I therefore support this PR. However, I would suggest two changes

1. The chunk option should be called something more "formal" e.g. `sql.is_statement` (`is_update_query` is a contradiction IIUC, since a statement with `UPDATE` (usually) does not return a result set, so it isn't a query). 
2. For PostgreSQL support I'd add [`SET`](https://www.postgresql.org/docs/15/sql-set.html), [`REFRESH`](https://www.postgresql.org/docs/15/sql-refreshmaterializedview.html) and [`VACUUM`](https://www.postgresql.org/docs/15/sql-vacuum.html) as keywords. (But this isn't strictly necessary with the `sql.is_*` option).

NB: It's possible to have `SELECT` statements that are not queries (e.g. [`SELECT INTO`](https://www.postgresql.org/docs/15/sql-selectinto.html)), as well as statements with a keyword that are queries (e.g. [`UPDATE [...] RETURNING`](https://www.postgresql.org/docs/15/sql-update.html)), which makes the `sql.is_*` option the most important element of this PR.

@cderv do you have time to look into this?

cderv:
@dpprdan thanks for pinging on this issue. I believe this was on long hold due to https://github.com/r-dbi/DBI/issues/321#issuecomment-700787844 - It would be better for the long term that the SQL engine for **knitr** leaves elsewhere than inside **knitr**. However, it seems the DBI is not a good place. Engine can be in their own package and SQL feature is not core of **knitr** directly. 

I think this could make sense to have an improved and new engine in its own package now, and that it is maintained with the help of SQL users and expert. What do you think ?

dpprdan:
@cderv That sounds very reasonable. I can see myself in "with the help of SQL users", so feel free to ping me if you think I can assist somewhere. 

You probably thought of this already, but it may make sense to consider the glue_sql engine here as well.

yihui:
Since we have gotten another vote from users, I can definitely consider merging this PR (since it doesn't appear to be too complicated). In the long run, I still think it will be great if we have one or two helpers who could have some soft commitment to maintain this engine in **knitr** (it may not require a lot of effort, and Posit/RStudio can also send gifts as incentives if necessary).

> 1. The chunk option should be called something more "formal" e.g. `sql.is_statement`

I think that's a great suggestion.

yihui:
I've rebased this PR onto the current `master` and tidied it up so it can be merged:

- Applied the review suggestion to rename the chunk option from `sql.is_update_query` to **`sql.is_statement`** (a query with a keyword may still return a result set, e.g. `UPDATE ... RETURNING`, and vice versa for `SELECT ... INTO`, so "statement" is more accurate).
- Kept the expanded keyword list (`ALTER`, `GRANT`, `MERGE`, `TRUNCATE`, `CALL`, etc.) for auto-detection, and the `sql.is_statement` option now overrides that detection when it is wrong.
- Added tests for the new keywords and verified all four dispatch paths (auto/forced × statement/query) plus the input validation via a mocked DBI connection.
- Added a NEWS item.

Thanks @rnorberg for the original contribution, and @dpprdan for the naming suggestion. The branch was force-updated to a clean single commit on top of current `master`.

yihui (approved):

PR #1900, opened 2020-09-25
Add a fig.alt chunk option to provide alternative text different than fig.cap

### LINKED ISSUE #1879: Add a fig.alt chunk option to provide alt text different from fig.cap
This is related to FR in https://github.com/rstudio/rmarkdown/issues/1867

Adding a `fig.alt` chunk option independent of `fig.cap` could be an easier solution to improve accessibility in HTML content produce. 

It is good practice to provide an alt text different than the caption: https://www.brandeis.edu/web-accessibility/understanding/images-alt-text.html

It seems like something doable and important toward more accessible content.

cderv:
This will close #1879 and https://github.com/rstudio/rmarkdown/issues/1867

the new `fig.alt` chunk option will be used if provided instead of `fig.cap` to populate `alt=` attribute in HTML for alternative text. 
This will improve the creation of better accessible content for screen reader.


cderv:
Travis issue is fixed in https://github.com/yihui/knitr-examples/pull/70/commits/a7d296f1d7b610f4f19172240a02cf64ba4f55b5

I added https://github.com/yihui/knitr/pull/1900/commits/a47c6eb05a8be81bfadc2fb77e6444597b101d28 because it seems that markdown syntax only cannot fully be used to provide alt-text and caption. 

I looked into it to understand better what becomes `text` and here is what I found: 

* Markdown strict will use the text as alt text
````bash
$ pandoc -f markdown_strict -t html <<< '![alt-tex](src)'
<p><img src="src" alt="alt-tex" /></p>
```` 
* Pandoc markdown won't 
```bash
$ pandoc -f markdown -t html4 <<< '![alt-tex](src)'
<div class="figure">
<img src="src" alt="" />
<p class="caption">alt-tex</p>
</div>
```
* Unless you deactivate implicite figure
````bash
$ pandoc -f markdown-implicit_figures -t html <<< '![alt-tex](src)'
<p><img src="src" alt="alt-tex" /></p>
````

So in markdown syntax, it seems `text` will lead to use the text as an alt text. 
I wondered if it should be used instead of `cap` in here 
https://github.com/yihui/knitr/blob/0a73970eea732332b0ac4cef0504ab4c7fc0b1ea/R/hooks-md.R#L55-L62
or maybe return another `res` if `fig.alt` is provided (but in that case it will replace `fig.cap`

What do you think about those different output for `hook_plot_md_base` and `fig.alt` support ?

**Comment by cderv** [comment]
Tested also by @jooyoungseo, thank you ! https://github.com/rstudio/rmarkdown/issues/1867#issuecomment-716207575

**Comment by jooyoungseo** [comment]
@cderv, I have just noticed that the `fig.alt` does not take any effect for `rmarkdown::word_document()` output. Maybe we need to write a lua filter to conditionally address word output? I have not had a chance to take a look at the XML structure for Word alt text, but it could be more complex. Anyhow, alt text is certainly needed for Word output at the end.

**Comment by atusy** [comment]
@jooyoungseo I'd suggest ask the feature on Pandoc side so that we can write something like

```md
![caption](image.png){alt=alternative-text}
```

cderv:
@atusy Alt text are already supported for docx it seems since a long time ! 
See https://github.com/jgm/pandoc/issues/2754. 

But the syntax though is 
```md
![alt text](foo.jpg "title text")
```
because this is how an image is parsed by Pandoc, but that is far from perfect across format I think
```sh
$ echo '![alt-text](foo.jpg "title text")' | pandoc -t native
[Para [Image ("",[],[]) [Str "alt-text"] ("foo.jpg","fig:title text")]]

$ echo '![alt-text](foo.jpg "title text")' | pandoc -t html
<figure>
<img src="foo.jpg" title="title text" alt="alt-text" /><figcaption aria-hidden="true">alt-text</figcaption>
</figure>

$ echo '![alt-text](foo.jpg "title text")' | pandoc -t html4
<div class="figure">
<img src="foo.jpg" title="title text" alt="" />
<p class="caption">alt-text</p>
</div>
```

I guess some improvement must be made regarding accessibility in Pandoc itself so that it is ok with all kind of output.

Can you try with the markdown syntax to see how it outputs in the docx and tell us if this is what is expected ? 
If so, we could use this markdown syntax specifically for docx output. Thank you.

Otherwise, doing that ourself would mean to generate some XML (https://stackoverflow.com/a/11472038/3436535), so redo exactly what Pandoc is doing (https://github.com/mb21/pandoc/blob/139fa54d48a878c91f6e56c17ee50e9d589d379f/src/Text/Pandoc/Writers/Docx.hs#L1152-L1158).

cderv:
@jooyoungseo , we may want to discuss that in another issue too. To see if something can be done here, or if we should wait / push for something different upstream in Pandoc.

atusy:
@cderv Thanks! It seems to be working. Just note that I opened docx with LibreOffice.

image

jooyoungseo:
I have opened [Pandoc issue #6782 ](https://github.com/jgm/pandoc/issues/6782). Please visit it at your best convenience.

NEWS.md
```diff
@@ -1,6 +1,6 @@
 # CHANGES IN knitr VERSION 1.31
 
-
+- Added a new `fig.alt` option that will be used for alternative text in `alt` HTML attribute instead of `fig.cap`, if provided. It will be evaluated after a chunk has been evaluated and recycled if multiple plots / images are in the chunk, as with `fig.cap` (#1900)
```
hadley, line 3: I assume this can be a vector so you can provide one alt label per plot?

cderv: Yes, it will work like `fig.cap` and others, so you can pass a vector and it will be recycled if needed too. 
I have set an example in `knitr-examples` to demo it (and test it) (https://github.com/yihui/knitr-examples/pull/70)

R/hooks-html.R
```diff
@@ -37,9 +37,9 @@ hook_animation = function(options) {
   paste(c(sprintf('width="%s"', w), sprintf('height="%s"', h), extra), collapse = ' ')
 }
 
-.img.tag = function(src, w, h, caption, extra) {
+.img.tag = function(src, w, h, caption, extra, alt = caption) {
   caption = if (length(caption) == 1 && caption != '') {
-    paste0('title="', caption, '" alt="', caption, '" ')
+    paste0('title="', caption, '" alt="', alt, '" ')
```
hadley, line 42: I wonder if setting the title to the caption is correct — you generally expect the caption to apply to the entire chunk, not individual plots. Maybe it should only be used if the figure doesn't have a caption displayed otherwise?

hadley: Ah got it, that makes sense.

yihui (approved):
Just a small question, which you can answer later. I'm going to merge this now. Thanks!

R/hooks-html.R
```diff
@@ -37,9 +37,9 @@ hook_animation = function(options) {
   paste(c(sprintf('width="%s"', w), sprintf('height="%s"', h), extra), collapse = ' ')
 }
 
-.img.tag = function(src, w, h, caption, extra) {
+.img.tag = function(src, w, h, caption, extra, alt = caption) {
```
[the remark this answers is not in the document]
cderv: It seems you're right. I may have not gone at the bottom of my idea.
It was to differentiate `title=` from `alt=` in the `` tag. 

Currently 
```r
> .img.tag("img-src", "10", "10", "cap", "extra = true")
[1] "<img src=\"img-src\" title=\"cap\" alt=\"cap\" width=\"10\" height=\"10\" extra = true />"
````

But it seems I forgot to pass the argument in the function using `.img.tag`.

PR #1902, opened 2020-09-28
Fix collapse = TRUE when attr.source or classe.source is set

### LINKED ISSUE #1893: `attr.source` and `class.source` conflict with `collapse = TRUE` option
Found while discussion https://github.com/rstudio/rmarkdown/issues/1876 

Using `attr.source` or `class.source` with `collapse=TRUE` will create multiple collapsed blocks instead of one block for the chunk as without these `*.source` option set

See reprex:

```` r
options(knitr.duplicate.label = "allow")
# correct result with collapse=TRUE
chunk <- c(
  '```{r, collapse=TRUE}',
  '1 + 1',
  '1:10',
  '```')
xfun::raw_string(knitr::knit(text = chunk))
#> 
#> ```r
#> 1 + 1
#> #> [1] 2
#> 1:10
#> #>  [1]  1  2  3  4  5  6  7  8  9 10
#> ```

# Adding attr.source or class.source creates 2 blocks
chunk[1] <- '```{r, collapse=TRUE, attr.source=".foo"}' 
xfun::raw_string(knitr::knit(text = chunk))
#> 
#> ```{.r .foo}
#> 1 + 1
#> #> [1] 2
#> ```
#> 
#> ```{.r .foo}
#> 1:10
#> #>  [1]  1  2  3  4  5  6  7  8  9 10
#> ```
chunk[1] <- '```{r, collapse=TRUE, class.source="foo"}' 
xfun::raw_string(knitr::knit(text = chunk))
#> 
#> ```{.r .foo}
#> 1 + 1
#> #> [1] 2
#> ```
#> 
#> ```{.r .foo}
#> 1:10
#> #>  [1]  1  2  3  4  5  6  7  8  9 10
#> ```
````

I think we expect this
```r
#> ```{.r .foo}
#> 1 + 1
#> #> [1] 2
#> 1:10
#> #>  [1]  1  2  3  4  5  6  7  8  9 10
#> ```
```
like the first example using only `collapse=TRUE`

cderv:
This will fix #1893 

This was a regex to change so that the header is correctly match

* Without class or attributes  we need to match 
````
```r
code1
```
```r
code2
```
````
* With the class or attr provided it becomes
````
```{.engine .class1 .class2 attr=val}
code1
```
```{.engine .class1 .class2 attr=val}
code2
```
````

Added some tests too

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # CHANGES IN knitr VERSION 1.31
 
-- `collapse = TRUE` now works as expected if `attr.source` or `class.source` is provided (thanks, @aosavi, #1902)
+- The chunk option `collapse = TRUE` now works as expected when the chunk option `attr.source` or `class.source` is provided (thanks, @aosavi @cderv, #1902).
 
 # CHANGES IN knitr VERSION 1.30
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # CHANGES IN knitr VERSION 1.31
 
-
+- `collapse = TRUE` now works as expected if `attr.source` or `class.source` is provided (thanks, @aosavi, #1902)
 
 # CHANGES IN knitr VERSION 1.30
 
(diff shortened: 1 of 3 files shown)


yihui (approved):
Looks good. Thank you!

PR #1905, opened 2020-10-05
Added more load_cache() arguments to avoid clobbering global environment

LTLA:
Bioconductor's [OSCA book](https://github.com/Bioconductor/OrchestratingSingleCellAnalysis) uses `load_cache()` to pull **knitr**-cached content from already-compiled chapters to enable their re-use when compiling later chapters. By and large, this works pretty well but `load_cache()` has the unfortunate side effect of loading everything in the labelled chunk into the global environment, not just the `object` I asked for. To illustrate, let's say we have a very simple Rmarkdown document:

````
---
title: Testing
---


```{r blah, cache=TRUE}
var1 <- 1
var2 <- 2
```
````

When I compile it with `rmarkdown::render`, restart the R session and load the chunk, I get:

```r
knitr::opts_knit$set(output.dir=".")
knitr::load_cache(path="blah_cache/html/", label="blah", object="var1") # good
##  [1] 1
var2 # I didn't ask for this
## [1] 2
```

This can be annoying when an object in my current environment (in an interactive R session, or during the course of compiling an Rmarkdon document) is silently and unintentionally overwritten. The solution presented in this PR adds a few arguments to `load_cache()` that give the caller a bit more control. The first is an `envir=` option that defines a workspace for cache loading, other than the default environment from `knit_global()`.

```r
knitr::load_cache(path="blah_cache/html/", label="blah", object="var1", envir=new.env())
## [1] 1
var2
## Error: object 'var2' not found
```

The second argument (`dir=`) is just a quality-of-life change so that `load_cache()` can be more easily called outside of a **knitr** context (e.g., in an interactive session), specifically by ignoring the (lack of) an `output.dir` setting.

In both cases, the defaults should preserve the current behavior of the function.

<details>
<summary>Session information</summary>

```
R version 4.0.0 Patched (2020-05-01 r78341)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS:   /home/luna/Software/R/R-4-0-branch-dev/lib/libRblas.so
LAPACK: /home/luna/Software/R/R-4-0-branch-dev/lib/libRlapack.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] knitr_1.30.1

loaded via a namespace (and not attached):
[1] compiler_4.0.0 tools_4.0.0    xfun_0.18     
```
</details>


yihui (commented):
I definitely agree this is an annoying problem and should be fixed. I'll take a closer look at your PR tomorrow. Thanks a lot!

LTLA:
Whoops, lost track of this. Any thoughts?

yihui:
Sorry for dropping the ball. I'll try my best to include this PR into the next version of **knitr**!

yihui (approved):
The changes look good to me. I'll do a few cosmetic changes and merge later. Thank you!

PR #1909, opened 2020-10-17
Let collapse work even if class.* or attr.* are provided (closes #1906)

### LINKED ISSUE #1906: Collapse should work even if class.output is provided ?
Currently, `collapse = TRUE` won't work if `class.output` and friends are provided. This is related to #1902 where we dealt with only a subset of the conflicts between those options.

### Reprex

When no `class.output` is provided
``` r
knitted <- knitr::knit(
  text = c(
    '```{r, class.source = "source", class.output = NULL, collapse=TRUE}',
    'str(1)',
    '```'
  ), quiet = TRUE)
xfun::raw_string(knitted)
#> 
#> ```{.r .source}
#> str(1)
#> #>  num 1
#> ```
```

When `class.output` is provided

``` r
knitted <- knitr::knit(
  text = c(
    '```{r, class.source = "source", class.output = "output", collapse=TRUE}',
    'str(1)',
    '```'
  ), quiet = TRUE)
xfun::raw_string(knitted)
#> 
#> ```{.r .source}
#> str(1)
#> ```
#> 
#> ```{.output}
#> #>  num 1
#> ```
```

<sup>Created on 2020-10-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

* Should we give priority to `collapse = TRUE` and ignore when attr of class is set to the output ? In that case, both output should be the same. 
* Or should we consider that `collapse = TRUE` is not compatible with the options `class.output`, `class.warning`, `class.error`, `class.message` ? And it that case, the output is currently correct.

Let's note that if we choose to ignore `collapse`, priority can be given by using an option hook
```r
knitr::opts_hooks$set(collapse = function(options) {
  if (isTRUE(options$collapse)) {
    opts <- paste("class", c("output", "warning", "message", "error"), sep = ".")
    options[opts] <- NULL
  }
  options
})
```
If we choose to give priority, I guess some hooks can also be done to set `collapse = FALSE` when those class options are set.

I don't know if there is other incomptatible option and what is done for them. 

@atusy, this is related to your issue in atusy/minidown#22. I opened this so that we discussed it before doing anything.

atusy:
#1906 mentions `class.output` and `attr.output`, but I also added those for message, warning, and error.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # CHANGES IN knitr VERSION 1.31
 
-- The chunk option `collapse = TRUE` now works as expected when the chunk option `attr.source` or `class.source` is provided (thanks, @aosavi @cderv, #1902).
+- The chunk option `collapse = TRUE` now works as expected when the chunk option `attr.*` or `class.*` is provided. By this change, The chunk option `collapse = TRUE` forces `attr.*` and `class.*` be `NULL` except for the chunk options `attr.source` and `class.source` (thanks, @aosavi @cderv @atusy, #1902 #1906).
 
 # CHANGES IN knitr VERSION 1.30
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -309,7 +309,7 @@ fix_options = function(options) {
 
   if (options$collapse) {
     options[unlist(lapply(
-      c("class.", "attr."), paste0, c("output", "message", "warning", "error")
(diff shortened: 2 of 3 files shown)


yihui (approved):
Great! Thanks!

cderv:
Thanks @atusy !

atusy:
My pleasure!! @cderv

PR #1910, opened 2020-10-22
Capture error in eng_sql if error = TRUE

cderv:
This will fix rstudio/rmarkdown#1208

We capture the error of SQL execution if `error = TRUE` as for some other engine. 

This allows this Rmd to work now without stoping

````md
---
title: "Error Honored"
output: html_document
---

```{r, message=FALSE}
knitr::opts_chunk$set(error = TRUE)
library(DBI)
con <- dbConnect(RSQLite::SQLite(), ":memory:")
```

```{r}
stop('error message that I want to print')
```

```{r}
dbWriteTable(con, "mtcars", mtcars)
```

```{sql, connection=con}
SELECT * FROM mtcars WHERE cyl = 4
```

```{sql, connection=con}
DROP TABLE mtcars
```

```{r}
dbListTables(con)
```

```{sql, connection=con}
select test;
```
````

I'll do a PR in knitr-example to complement this PR with tests

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # CHANGES IN knitr VERSION 1.31
 
+- `eng_sql` now correctly captures error if `error = TRUE` in chunk options. (thanks, @colearendt, rstudio/rmarkdown#1208)
+
 - The chunk option `collapse = TRUE` now works as expected when the chunk option `attr.*` or `class.*` is provided. By this change, The chunk option `collapse = TRUE` forces `attr.*` and `class.*` be `NULL` except for the chunk options `attr.source` and `class.source` (thanks, @aosavi @cderv @atusy, #1902 #1906).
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
Thanks!

cderv:
One thought after having pushed that... 

How does that work in knitr so that the `error` hook is used ? For example if someone want to use a specific `class.error`, I am not sure what I have done is enough. More generally, I don't think for non R engine that `wrap.error` will be called correctly. 
Should this be improve ?

yihui:
This could be improved. The `out` argument of `engine_output()` can take a list. You could add the error message to the list, and make its class to be `error`, then `knitr:::wrap` should handle it.

The work is probably not trivial, so I'd rather wait for someone to request it before we spend the time on the improvement.

cderv:
Thanks. 

Let's wait then.

PR #1917, opened 2020-11-06
add lock bot

cderv:
This add the lock thread action. 

I set it up every hour for a start in order to lock every existing closed thread. We need that because it locks by batch of 50 max. 
After that, we could set it every week or every month only. 

I let you merge when you decide it is time to run the bot 😄

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ name: 'Lock threads'
 
 on:
   schedule:
-    - cron: '10 * * * *'
+    - cron: '*/10 * * * *'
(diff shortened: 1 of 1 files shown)


cderv:
We could also set it every 10 minutes to be settled quicker., There is ~1700 issues to lock... 
https://github.com/yihui/knitr/issues?q=updated%3A%3C2020-05-10+is%3Aclosed+

So 34 jobs to be run ! 

Would be ~5-6 hours, maybe more interested

yihui (approved):
Thanks!

Issue #1858, opened 2020-06-03
Suggestion: Support for RStudio/Webshot2 instead of Webshot?


LouisStAmour:
Webshot2 is a drop-in replacement for Webshot, though the requirements have changed and phantomjs isn't available, instead it attempts to use Chrome via some complicated dependencies.

The current code explicitly loads "webshot" and requires "find_phantom" -- so I hacked together my own version of webshot that renames the package from "webshot2" to "webshot" and adds that method ("find_phantom") at https://github.com/LouisStAmour/webshot2 installable via `Rscript -e 'remotes::install_github("LouisStAmour/webshot2")'`

The drop-in replacement appears to work as expected with Bookdown workflows for DiagrammeR Mermaid to PDF (though the PDF has a whitespace cropped diagram, smaller dimensions, compared to the GitBook HTML display of the SVG). Haven't tested anything too complicated, but I figure Chrome is more compatible than PhantomJS long-term.

So this ticket is a request to modify Knitr to look for Webshot2 first, then Webshot second. So far the only folks I've seen support both are https://github.com/davidgohel/flextable/blob/7d9f14aecb015d6403ce0b44091e586e45490142/R/printers.R#L282-L358 so maybe that code will help some. Not sure, this is my first time writing R code in almost a decade. :D

Thanks very much for your work on Bookdown and related projects!

dmurdoch:
This change might allow `rgl` to be a bit simpler:  `webshot::webshot` can't handle WebGL displays (as in https://rpubs.com/murdoch/652451), whereas a quick test shows that `webshot2::webshot` can.

yihui:
The development version prefers **webshot2** over **webshot** now:

```r
remotes::install_github('yihui/knitr')
```

Thank you for the suggestion (and @atusy for the implementation)!

PR #1918, opened 2020-11-11
prefer webshot2 to webshot and solve #1858

atusy:
With this PR, html_screenshot` tries to use **webshot2** and fallbacks to **webshot**.

Also, user may prioritize **webshot** by `opts_chunk$set(webshot = "webshot")`.
Another possibility is to use `opts_knit$set(webshot = "webshot")`.
I'm fine with both ways, but choose the `opts_chunk` because it has controlls on each chunks.


R/plot.R
```diff
@@ -488,18 +488,21 @@ need_screenshot = function(x, ...) {
...
 }
 
 webshot_available = local({
-  res = NULL  # cache the availablity of webshot/PhantomJS
+  res = NULL  # cache the availablity of webshot2 and webshot/PhantomJS
   function() {
     if (is.null(res))
-      res <<- loadable('webshot') && !is.null(getFromNamespace('find_phantom', 'webshot')())
+      res <<- c(
+        webshot2 = loadable('webshot2'),
```
yihui, line 503: Should we test the availability of Chrome/Chromium?

yihui (approved):
Looks good to me. Just one small question. Thank you!

R/plot.R
atusy: I agree. Currently, **chromote** does not export `find_chrome`, so I'll open an issue there.
https://github.com/rstudio/chromote/blob/3a5329220d84de5f3a442cf05057d1f4ed7d3a7a/R/chrome.R#L17

atusy: Oops, we do not need `chromote:::find_chrome` be exported as we use `getFromNamespace`

atusy: Done via [`01d7c28` (#1918)](https://github.com/yihui/knitr/pull/1918/commits/01d7c28dd0073be1e97c7ca011c987ed743e089e)

yihui (approved):
Great. I'll carry on from here and merge in a few minutes. Thank you!

R/plot.R
cderv: @atusy I think it is still a good idea to suggest they export this function to allow test of chrome existence in other package. Worth a try ? 
Or suggest they add a `is_chrome_available()` in **webshot2**  ? 🤷

yihui:
I forgot to mention that another possible idea is to use `pagedown::chrome_print()`. We have exported `pagedown::find_chrome()` there, and it is a CRAN package. I guess the only problem is the compatibility of arguments.

R/plot.R
atusy: @cderv I agree. It is better to have the function in **webshot2**. Then we can incorporate Yihui's refactoring https://github.com/yihui/knitr/commit/997398230c976bd10d24423fc84268a0a9be88a5#. I'll open issue tomorrow.

cderv:
The function are a bit different in their way to find chrome, and specifically with the env var mechanism which is not the same. They both will find chrome but their could be a mismatch of version if you have several locally, as **webshot2** will use the one that **chromote** will use. 

Also `webshot:::find_phantom()` was unexported - does feel bad to do the same. 
I would stick to the one in **chromote** for now, and maybe ask them to put one in **webshot2** as a helper to test availability of chrome for other package ? or better that they do the check themselves maybe.

cderv:
`find_chrome` is now exported from **chromote**. Not yet from **webshot2** I believe.

PR #1920, opened 2020-11-17
Use github actions instead of travis

cderv:
* Using r-lib/actions as template
* Windows Linux and MacOS
* **We won't publish to XRAN** - could be added later

## Todod

- [x] Add a workflow to test knitr-examples - same as it is working now.
- [ ] Debug knitr-examples workflow

## Idea

- [x] It would be great to cache package installed for knitr-examples. It is currently 9 minutes.


cderv:
## Status & Progress: 

We got this after knitr-examples checks
````
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   025-engine-ruby.md
	modified:   027-engine-bash.md
	modified:   042-highlight-matlab.tex
	modified:   098-highlight-python.tex
````

- [ ] 025 example is not working because Ruby throw deprecated warnings (https://github.com/yihui/knitr/pull/1920/checks?check_run_id=1414097721##step:11:2657) 
- [x] `027-engine-bash` is failing because other engine docs have changed - modified example
- [ ] `highlight` examples have changed - don't know why. Something is different with highlight in GHA compare to travis

Moreover

- [ ] `asy` chunk is not working (knitr example 093

cderv:
On other note, I also removed some apt installation that where not needed (maybe they are already available in the runner)

* The dev dependencies for packages are now found automatically by `remotes::system_requirements("ubuntu", "18.04")`
* Some tools are [already installed ](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md) like cargo, ghc, texinfo. 

There is only `qpdf` that does not seem to be installed already but was not required.

cderv:
It would be better to merge this after merging knitr-examples, which I can't do myself. 

Next step would be 
* Do we need to fix highlight example ? Should we modify the resulting .md file due to the change we are seing in output ? 
* Merge change in knitr-examples once ok. 
* Modify the checkout of master in knitr-examples in the workflow in this PR
* Merge this PR in knitr.

yihui (approved):
The highlight examples have been updated: https://github.com/yihui/knitr-examples/commit/0262a05f38eb530bd0280344d5f8ff72aef61454

Thanks!

PR #1922, opened 2020-11-25
add an option for table.attr argument

cderv:
To close https://github.com/rstudio/rmarkdown/issues/1954, I find it interesting to have an option to set a table attribut for all kable using format html. 

Do you fing that interesting ?


cderv:
FWIW checks are failing on WINDOWS and MACOS because rgl is not available as binary yet.

atusy:
two ideas

1. As there are chunk options starting with "attr" but  "table", I would suggest "attr.table".
2. it would be nice to have "attr.fig" or "fig.attr" as a more self-explaining alternative of "out.extra". Of course we can keep "out.extra" for backward compatibilities.

2 is slightly off-topic, but worth discussing because the decision for 1 may affect the decision for 2.
Based on 1, I would say "attr.fig" > "fig.attr", however, there are already options starting with "fig".
So I am not sure which is better.
This might be slightly off-topic.

cderv:
@atusy, I am not sure to follow your comment. This is not a chunk option, this is an R option (to set using `options()`) for `knitr:::kable_html` which is called internally by `kable(..., format = 'html')`. 

This is following the discovery that you need to set the `table.attr` in the function is you want the table to be displayed as bootstrap table 
```r
knitr::kable(, 'html', table.attr = 'class="table table-condensed"')
```
I was thinking about using an R option to set this globally if someone needs it for all table instead of needed a copy-paste or a wrapper function. the argument name is `table.attr` not `attr.table`, so I kept the order.

This is like `knitr.table.format` option which is used to control `format` argument in kable. Following the pattern would lead to `knitr.table.table.attr` but I added `html` as it is only for this format and removed the second `table`

Options for knitr's chunk options are under `knitr.chunk.foo` - so I don't think this is confusing and there is no `attr.table` chunk option currently, so we should keep the name in case we want to add one. 

Does this explanation change your thoughts on this ?

atusy:
Oops. Sorry. I totally misunderstood.

yihui (approved):
Sounds good to me. Thanks!

R/table.R
```diff
@@ -327,7 +327,7 @@ kable_latex_caption = function(x, caption) {
   ), collapse = '')
 }
 
-kable_html = function(x, table.attr = '', caption = NULL, escape = TRUE, ...) {
+kable_html = function(x, table.attr = getOption('knitr.table.html.attr', ''), caption = NULL, escape = TRUE, ...) {
```
yihui, line 330:
```suggestion
kable_html = function(
  x, table.attr = getOption('knitr.table.html.attr', ''), caption = NULL, escape = TRUE, ...
) {
```

PR #1936, opened 2020-12-21
Update tweak_bib.csv

beanumber:
added `sessioninfo` tweak for R Core Team


yihui (changes requested):
Thanks! I'd rather do it this way (`Temple Lang` has exactly the same problem):
https://github.com/yihui/knitr/blob/f8f90baad99d873202b8dc8042eab7a88fac232f/R/citation.R#L99
so we don't have to deal with `{R Core Team}` on a package by package basis.

beanumber:
That makes sense. I'll close this.

yihui:
Please feel free to submit another PR. Thanks!

PR #1951, opened 2021-01-26
export and document pandoc_to() and pandoc_from()

### LINKED ISSUE #1950: Generalize is_format_output to test more than HTML and LaTeX
From a tweet https://twitter.com/graemeleehickey/status/1352947038690938880

There are currently only `is_html_output()` and `is_latex_output()` but no `is_docx_output()` or other. 

Would that be interesting to extend ? 

But maybe **knitr** is not the right place for this ? 

However, one solution could be to export `knitr:::pandoc_to()` and let users test on this.

cderv:
this closes #1950 

And I took the opportunity to start changing styles of tests.


yihui (commented):
I'd merge the doc of these functions to

```r
#' @rdname output_type
```

since they are essentially the same thing.

cderv:
This is what I made at first but think back and document in its home page. I have seen that as lower level and I thought it would be interesting to have them link (by `See also`) but separate.

I should have stick with my first feeling.

yihui (approved):
Thanks!

PR #1954, opened 2021-01-27
fix setting col in par() with global.pars = TRUE

### LINKED ISSUE #1603: Possible bug in knitr::opts_knit$set
Dear Yihui,

In the attached code below, the first plot is red while the second is black even though I'm using the global.par = TRUE .

Any error on my side?
(I'm using knitr 1.2 and R version 3.5.0)

````
---
title: "Global.par error"
date: "9/11/2018"
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_knit$set(global.par = TRUE)
```

```{r plotA, echo=FALSE}
par(col = "red")
with(mtcars, plot(qsec, disp))
```

```{r plotB, echo=FALSE}
with(mtcars, plot(qsec, disp))
```

````

Thank you!

### PR DESCRIPTION
**Pull request by cderv**
This PR fix #1603 by adding the special handling of `fg` option in `par2()`

I took the opportunity to add some test for this function - it seems one of the parameters is changed by others when we use the equivalent of `par(par(no.readonly = TRUE)`. 
There are dependencies between parameters and setting all of them at once seems to have side effect. Usually you do 
```r
op = par(x)
# do something
par(op)
```
and not 
```r
par(x)
all = par(no.readonly = TRUE)
# do something
par(all)
```

It does not seem to be equivalent. But we can't really do the former.

### DIFF
--- a/tests/testit/test-plot.R
+++ b/tests/testit/test-plot.R
@@ -130,3 +130,39 @@ if (xfun::loadable('tikzDevice') &&
 
 # https://github.com/yihui/knitr/issues/1166
 knit(text = "\\Sexpr{include_graphics('myfigure.pdf', error = FALSE)}", quiet = TRUE)
+
+with_par <- function(expr, ...) {
+  # set par
+  op = graphics::par(...)
+  # reset on exit
+  on.exit(graphics::par(op))
+  # save changed state
+  global.pars = par(no.readonly = TRUE)
+  # reset par
+  graphics::par(op)
+  # simulate what happens when global.par = TRUE by restoring pars
+  par2(global.pars)
+  # evaluate in this state
+  force(expr)
+}
+
+assert("par2 correctly handles specific pars", {
+  (par2(NULL) %==% NULL)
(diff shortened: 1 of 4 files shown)


tests/testit/test-plot.R
```diff
@@ -130,3 +130,39 @@ if (xfun::loadable('tikzDevice') &&
 
 # https://github.com/yihui/knitr/issues/1166
 knit(text = "\\Sexpr{include_graphics('myfigure.pdf', error = FALSE)}", quiet = TRUE)
+
+with_par <- function(expr, ...) {
```
yihui, line 134:
```suggestion
with_par = function(expr, ...) {
```

yihui (approved):
Looks good. Add a NEWS item and it should be ready to go. Thanks!

tests/testit/test-plot.R
cderv: Force of habits ! I missed one... Thanks! 
I should make a specific style to use with **styler** and check this before pushing 😄

PR #1955, opened 2021-01-28
issue with `fig.keep`: don't try to remove figures if all should be kept

### LINKED ISSUE #1579: fig.keep=3 suppresses code and output for markdown chunk
N.B. This was placed on Stack Overflow 24 hours ago.

I want to have only one plot generated by the following chunk. It currently generates two identical plots.

````
```{r addGridLines, fig.dim=c(6,6), fig.cap="Scatter plot of Ozone versus Wind with grid lines added.", fig.keep=3}
attach(airquality)
plot(Wind, Ozone, pch = 4)
abline(v=min(Wind)+c(0.25,0.5,0.75)*(max(Wind)-min(Wind)), col=6)
abline(h = min(Ozone, na.rm=TRUE) +c(0.25,0.5,0.75)*(max(Ozone, na.rm=TRUE)-min(Ozone, na.rm=TRUE)), col=6)
library(BrailleR)
WTF()
detach(airquality)
```
````

The library command is added here, but is in an earlier chunk in my document.

If I use the chunk argument fig.keep="first", then the scatter plot with no grid lines is generated. Using fig.keep="last" means the plot appears after the WTF() command. I know why this happens. I want the figure to appear before the library() command instead.

The WTF() command grabs the current graph window and re-creates the content using the grid system if it isn't in grid form already. In this case, the base graphics system is being used and so this chunk generates two graphs for me, both being given the same  caption.

My problem is that if I use fig.keep=3, I get the right graph, but none of the commands or the text output from WTF() are returned.

### PR DESCRIPTION
**Pull request by cderv**
when using `fig.keep` with index vector it could happen. This should fix #1579 

I took the opportunity of this PR to 

1. Refactor a bit with a new internal function `rearrange_figs()` to apply then logic previously in `block_exec()`
2. Add some unit tests thanks to this new function
3. Change the style of the other tests

About the unit test, it took me some times to understand the plot logic with `recorderplot` and low level plot. I believe I got it right, but you set in the start of the test file
```r
op = options(device = function(file = NULL, ...) {
  pdf(file, ...)
  dev.control('enable')  # important! otherwise plots get discarded
})
```
which is unset with `options(op)` a bit after

So I am unsure if I can use plot functions below this. It seems to work though. 

Last step would have been a refactoring between files as I think it could be interested to have all the plot related function in a file `plot.R` like the test files. 

That is for another time.


yihui (approved):
Looks good. Thanks!

PR #1958, opened 2021-01-29
In Rnw LaTeX files, suppress the paragraph break at the end of the chunk if chunk option end.paragraph=FALSE.

krivit:
Fixes yihui#1851.

@cderv , here's the PR. A few questions:

1. Should I also update `NEWS.md`, and is this a sufficient contribution for the author list in the `DESCRIPTION`?
2. Is `end.paragraph` a good name for the option?
3. Where does this need to be documented? (I don't think the `knitr` R help documents the chunk options anywhere.)
4. Do tests need to be added anywhere?

Also closes #1092.

### DIFF
--- a/R/hooks-latex.R
+++ b/R/hooks-latex.R
@@ -191,7 +191,13 @@ hook_plot_tex = function(x, options) {
   k2 = '\\end{kframe}'
   x = .rm.empty.envir(paste0(k1, x, k2))
   size = if (options$size == 'normalsize') '' else sprintf('\\%s', options$size)
-  if (!ai) x = sprintf('\\begin{knitrout}%s\n%s%s\n\\end{knitrout}', size, x, if (options$end.paragraph %n% TRUE) '' else '%')
+  if (!ai) {
+    # if the chunk content starts with \n, don't add \n; similarly, if it ends with \n, don't append \n
+    n1 = n2 = '\n'
+    if (grepl('^\\s*\n', x)) n1 = ''
+    if (grepl('\n\\s*$', x)) n2 = ''
+    x = sprintf('\\begin{knitrout}%s%s%s%s\\end{knitrout}', size, n1, x, n2)
+  }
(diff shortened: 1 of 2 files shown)


R/hooks-latex.R
```diff
@@ -191,7 +191,7 @@ hook_plot_tex = function(x, options) {
   k2 = '\\end{kframe}'
   x = .rm.empty.envir(paste0(k1, x, k2))
   size = if (options$size == 'normalsize') '' else sprintf('\\%s', options$size)
-  if (!ai) x = sprintf('\\begin{knitrout}%s\n%s\n\\end{knitrout}', size, x)
+  if (!ai) x = sprintf('\\begin{knitrout}%s\n%s%s\n\\end{knitrout}', size, x, if (options$end.paragraph %n% TRUE) '' else '%')
```
yihui, line 194:
```suggestion
  if (!ai) {
    # if the chunk content starts with \n, don't add \n; similarly, if it ends with \n, don't append \n
    n1 = n2 = '\n'
    if (grepl('^\\s*\n', x)) n1 = ''
    if (grepl('\n\\s*$', x)) n2 = ''
    x = sprintf('\\begin{knitrout}%s%s%s%s\\end{knitrout}', size, n1, x, n2)
  }
```

yihui (commented):
How about the hack I suggested? Personally I'd prefer providing a chunk option like the one you proposed, but I feel we have had too many chunk options. Perhaps the hack I suggested could solve the most common problems. If it can, we won't need to consider adding a new option or document it.

Thanks!

R/hooks-latex.R
krivit: My reason for using the option in the first place was that there may be Rnw documents out there that rely on the old behaviour, and I didn't want to presume to break backwards compatibility. If you think it's OK in this case, then I don't mind.

jpritikin:
Horray!

PR #1960, opened 2021-02-05
Support inheriting chunk options from ref.label (closes #1959)

### LINKED ISSUE #1959: Carry over fig.alt label when referencing chunks with ref.label
This is a feature request to carry over the `fig.alt` option when referencing chunks with `ref.label`:

When earlier chunks are referenced with `ref.label`, chunk options don't carry along to the next chunk. It would be nice if, at a minimum, the `fig.alt` text was carried over to avoid the need for copying and pasting the alt text (which can sometimes be extensive). In the following example, the same figure is shown twice, once without and once with the code. In the rendered document alt text only appears for the first plot, not the second.

```
---
output: html_document
---

```{r pressure, echo = FALSE, fig.alt = "This is alt text for pressure plot."}
plot(pressure)
` ``


```{r ref.label = "pressure"}
` ``
```

---

atusy:
By this PR, referencing label like `ref.label(knitr::with_opts("label"))` can inherit chunk options.
If the current chunk specifies the same options, they have the priority.

## Example

``` r
cat(knitr::knit(
  text = paste(
    '```{r label, foo="foo", bar="bar"}',
    'str(knitr::opts_current$get(c("foo", "bar")))',
    '```',
    '',
    '```{r ref.label=knitr::with_opts("label"), foo = "FOO"}',
    '```',
    sep = '\n'
  ),
  quiet = TRUE
))
#> 
#> ```r
#> str(knitr::opts_current$get(c("foo", "bar")))
#> #> List of 2
#> #>  $ foo: chr "foo"
#> #>  $ bar: chr "bar"
#> ```
#> 
#> 
#> ```r
#> str(knitr::opts_current$get(c("foo", "bar")))
#> #> List of 2
#> #>  $ foo: chr "FOO"
#> #>  $ bar: chr "bar"
#> ```
```

<sup>Created on 2021-02-05 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


R/parser.R
```diff
@@ -159,6 +159,14 @@ parse_params = function(params) {
   if (!is.character(res$label))
     res$label = gsub(' ', '', as.character(as.expression(res$label)))
   if (identical(res$label, '')) res$label = unnamed_chunk()
+
+  ref.label = eval(res$ref.label)
+  if (inherits(ref.label, 'with_opts')) {
+    template = attr(knit_code$get(ref.label), 'chunk_opts')
+    targets = setdiff(names(template), names(res))
+    res[targets] = template[targets]
```
cderv, line 167: Why `merge_list()` is not working here ?

tests/testit/test-parser.R
```diff
@@ -25,7 +25,18 @@ assert(
...
+    knit_code$set(
+      ref = structure('', chunk_opts = list(foo = 'foo', bar = 'bar'))
+    )
+    res = identical(
+      parse_block(NULL, '', 'r, label, ref.label=with_opts("ref"), foo="FOO"')$params,
+      alist(label = 'label', ref.label = with_opts('ref'), foo = 'FOO', bar = 'bar')
+    )
+    knit_code$delete('ref')
+    res
+  })
```
cderv, line 39: I think I would write that in another `assert()` as this one concerns specifically `"parse_params() parses the language engine from ```{lang}'"`. But maybe that is just me.

Regarding the use of `local()`, I always forgot that`testit::assert()` does not evaluate in a local context (as `testthat::test_that()` does) making this required (or removing the object). Nice to be reminded.

R/utils.R
```diff
@@ -1017,5 +1017,16 @@ make_unique = function(x) {
 #' }
 image_uri = function(f) xfun::base64_uri(f)
 
+#' Reference label with chunk options
+#'
+#' When the `ref.label` chunk option is specified with this function,
+#' the chunk inherits chunk options from the reference chunk.
+#'
+#' @param ref.label A string to be passed to the `ref.label` chunk option.
+#' @export
```
cderv, line 1026: We don't use markdown syntax in **knitr** yet. You should use `\code{ref.label}` instead.

R/utils.R
```diff
@@ -1017,5 +1017,16 @@ make_unique = function(x) {
 #' }
 image_uri = function(f) xfun::base64_uri(f)
 
+#' Reference label with chunk options
+#'
+#' When the `ref.label` chunk option is specified with this function,
+#' the chunk inherits chunk options from the reference chunk.
```
cderv, line 1023: I would add in documentation that inheritance means that unset chunk option in the current chunk locally will be set to the value of the reference chunk. 

This is it, right ?  No overwrite of local chunk option by the referenced chunk.

cderv (changes requested):
Thanks @atusy ! 

I like this feature. That is a nice way to do it.

I did a first review but @yihui will be the one to decide if it will be merged as is.

R/parser.R
atusy: Sorry, I thought `alist` cannot be merged, which was wrong.
I got CI error because order of elements are wrong.
I again introduced `merge_list` by 0712b1a

tests/testit/test-parser.R
atusy: Done via f16d4dc.
I was just lazy to split tests.

R/utils.R
atusy: Oops, how many times do I make that mistake!
Thank you.
Fixed via fce3f3b.

R/utils.R
atusy: You are right! Done via 57a14de .

atusy:
Thanks @cderv for the neat review!!
I fixed all of them.

Note: Windows environment does not pass CI due to `'"pdflatex"' not found`.

cderv:
Thanks. 

> Note: Windows environment does not pass CI due to '"pdflatex"' not found.

This an issue with the download of TinyTeX which is not correctly handled. I tried to fixed that already but it seems not enough. The error is different than before though. THanks.

cderv (approved):

yihui (commented):
Thanks! Instead of a new function `with_chunk()`, I was thinking of using something simpler like `ref.label = I('foo')` to indicate that chunk options of the chunk `foo` should be inherited. Does that sound like a good idea?

cderv:
pinging @mine-cetinkaya-rundel as author of the feature request in the first place. 

* 1st proposal: 
````markdown
---
output: html_document
---

```{r pressure, echo = FALSE, fig.alt = "This is alt text for pressure plot."}
plot(pressure)
```


```{r ref.label = knitr::with_opts("pressure")}
```
````

* 2nd proposal
````markdown
---
output: html_document
---

```{r pressure, echo = FALSE, fig.alt = "This is alt text for pressure plot."}
plot(pressure)
```


```{r ref.label = I("pressure")}
```
````

I think `I()` is less explicit than `knitr::with_opts()` but I see the simplicity of using it once you know how this works. It is also shorter to write which could be seen as an advantage for code chunk options.

**Comment by mine-cetinkaya-rundel** [comment]
While brevity is good for chunk options especially, I would favour explicitness over succinctness here, especially since the chunk with this option would likely not have many other options listed since it's inheriting (at least some of) them. Therefore I like the first option better. 

I've also not used `I()` outside of a formula setting much (e.g. defining a model) so I'd find it quite surprising to see it used here, at least the first time.

**Comment by atusy** [comment]
I think the both has something good that the other do not have.

1. `I` is nicely short.
2. `with_opts` appears in the `help(package="knitr")`, it means users can easily find what they need from the document.

Although I love the document comes with the package, I think `I` is acceptable and enough if we update [the chunk options documentation](https://yihui.org/knitr/options/) because no one will even understand what `ref.label` is without reading the page.

**Review by yihui — COMMENTED** [review]
Technically, `I()` and `knitr::with_opts()` are essentially the same idea. Both of them simply add a special class to the value passed to them (`AsIs` or `with_opts`), and the class indicates some kind of specialness. All we need is an indicator. This indicator could also be a chunk option, e.g.,

````
---
output: html_document
---

```{r pressure, echo = FALSE, fig.alt = "This is alt text for pressure plot."}
plot(pressure)
```

```{r ref.label = "pressure", opts.label = TRUE}
```
````

`opts.label = TRUE` means inheriting chunk options from the label in `ref.label`.

Personally I'd use `I()` by myself, but I'll be happy to support `opts.label = TRUE` if you don't mind typing more characters (of which I totally understand the advantage).

BTW, `opts.label` is an existing chunk option that has a different meaning at the moment (but doing a similar task of including chunk options from another label). I'm proposing to add another usage of this option.

**Comment by atusy** [comment]
@yihui Thanks for your thoughts. I love the `opts.label = TRUE` idea. In this way, we can also do `opts_chunk$set(opts.label = TRUE)` to tweak the default behavior. I'll follow your suggestions :)

**Comment by mine-cetinkaya-rundel** [comment]
I think `opts.label = TRUE` could work as well. 

Would there ever be a situation where one might want to use both the options from the chunk they're referencing with `ref.label` and also the options from a template (original use case for `opts.label`)? If so, would this look like `opts.label = (TRUE, myfigures)`? Or perhaps `opts.label = "myfigures"` would be used in the original chunk and the second one would inherit all options from the first chunk with `opts.label = TRUE`?

**Comment by atusy** [comment]
I understand @yihui's suggestion is to support `ref.label=I('chunk-label')` and `opts.label=TRUE`.
If so, @mine-cetinkaya-rundel's case can be approached by `ref.label=I('chunk-label'), opts.label='template-name'`.
However, I won't encourage it because of the difficulty in predicting which chunk options have the priority.

**Comment by cderv** [comment]
> Or perhaps opts.label = "myfigures" would be used in the original chunk and the second one would inherit all options from the first chunk with opts.label = TRUE?

I think I would promote this: 

* defines all chunk options in the main chunk then... 

* use [same label to reuse a chunk content](https://bookdown.org/yihui/rmarkdown-cookbook/reuse-chunks.html#same-label)
  * set chunk option as any other chunks
  ````r
  ---
  output: html_document
  ---
   
  ```{r pressure, echo = FALSE, fig.alt = "This is alt text for pressure plot."}
  plot(pressure)
  ```
    
  ```{r, pressure, echo = TRUE, fig.alt = "This is alt text for pressure plot."}
  ```
  ````
  * add `opts.label = TRUE` to get all the chunk options from the referenced chunk
  ````r
  ---
  output: html_document
  ---
   
  ```{r pressure, echo = FALSE, fig.alt = "This is alt text for pressure plot."}
  plot(pressure)
  ```
    
  ```{r, pressure, opts.label = TRUE}
  ```
  ````
  * use options template if you want to share some chunk options but modify some (`opts.label="shared"` + other options)
  ````markdown
  ---
  output: html_document
  ---
  
  Define alt text

  ```{r pressure-alt, echo = FALSE}
  knitr::opts_template$set(
    pressure = list(fig.alt = "This is alt text for pressure plot.", echo = FALSE)
  )
  ```
  
  and use it 
  
  ```{r pressure, opts.label = "pressure"}
  plot(pressure)
  ```
  
  ```{r, pressure, opts.label = "pressure", echo = TRUE}
  ```
  ````
  * Use an object to store common value
  ````r
  ---
  output: html_document
  ---
   
  ```{r pressure, echo = FALSE}
  pressure_alt = "This is alt text for pressure plot."
  plot(pressure)
  ```
    
  ```{r, pressure, fig.alt = pressure_alt, echo = TRUE}
  ```
  ````
* use [`ref.label` when you want to combine several chunk into one](https://bookdown.org/yihui/rmarkdown-cookbook/reuse-chunks.html#ref-label)  
  * set chunk options as any other chunk
  * `opts.label = TRUE` would be possible only if `length('ref.label') == 1`, because otherwise how to set the priority ? 
    * as example in https://github.com/yihui/knitr/pull/1960#pullrequestreview-634173043
  * Use options template 
  * Use common object
  ````markdown
  ---
  output: html_document
  ---
  
  Define alt text
	  
  ```{r pressure-alt, echo = TRUE}
  pressure_alt = "This is alt text for pressure plot."
  ```
	  
  and use it 
	   
  ```{r pressure, echo = FALSE, fig.alt=pressure_alt}
  plot(pressure)
  ```
	
  In another chunk where we should the two chunks in one
	  
  ```{r, ref.label = c("pressure-alt", "pressure"), fig.alt = pressure_alt, echo = TRUE}
  ```
  ````

Trying to cover all the cases here for this kind of feature. `ref.label` can take a vector of labels to reference and combine several chunks so this case must be taken into account.

File: R/parser.R [hunk 5]
```diff
@@ -159,6 +159,21 @@ parse_params = function(params) {
   if (!is.character(res$label))
     res$label = gsub(' ', '', as.character(as.expression(res$label)))
   if (identical(res$label, '')) res$label = unnamed_chunk()
+
+  ref.label = eval(res$ref.label)
```
yihui, line 163: The only thing that I'm not sure about is whether `eval()` should take place here, or inside `call_block()`: https://github.com/yihui/knitr/blob/18ce1a0b61295fdc6fa995b6ae3fe52ddf637bb5/R/block.R#L26-L30

The advantage of the latter is that `ref.label` could be an R expression that uses variables in the knitting environment, e.g.,

````
```{r, foo}
1 + 1
```

```{r}
lab = "foo"
```

```{r, ref.label = lab}
```
````

I understand that this may be a rare use case.

yihui (approved):
Looks okay to me now. Just one comment. Thanks!

R/parser.R
cderv: I think it would be fairly rare. But is there any downside of allowing this ?

yihui:
@mine-cetinkaya-rundel Good question. Here is my proposal for `opts.label`:

- When it is a character value:
    -  We first look into `opts_template`. If the label is found there, we inherit the chunk options from `opts_template`.
    - Then we continue to search for the label in other code chunks. If a code chunk has the label specified by `opts.label`, we use these chunk options to partially override the ones from the first step. Here *partially* means only overriding options of the same names, e.g., for `opts_template$set(foo = list(a = 1, b = 2))` and ```` ```{r, foo, b = 3}````, `opts.label = 'foo'` means `{a = 1, b = 3}`. This may be a small breaking change, but I really love this idea and have wanted this feature myself many times. On the other hand, I guess [not many people are aware of `opts_template`](https://twitter.com/jaredlander/status/1348040075276410880), so it's probably not too late to make this change.
- When it is `TRUE`, we treat it as a shorthand of `opts.label = ref.label`.

For `ref.label`, `I()` is a shorthand of `opts.label = TRUE`.

R/parser.R
yihui: No, I don't think there is a downside. It's quite flexible and I actually like it: https://bookdown.org/yihui/rmarkdown-cookbook/chunk-variable.html This is a great power that I guess Jupyter will never have (every time someone claims that Jupyter is an implementation of "Literate Programming," I smile in silence---it doesn't support _programming_ in chunk options).

So I tend not to put the constraint on `ref.label` that it must be a literal character vector. In other words, I'm in favor of moving this part to `call_block()`.

cderv: That confirms my thought, thanks. I don't see why not do it then !

yihui (commented):
@atusy If you don't mind, I can carry on from here and finish the PR. It's midnight in Japan now and I don't want you to stay up too late. This is the last issue I want to resolve for the next **knitr** release, so I'm hoping to finish it and start running reverse dependency checks. Thanks!

yihui:
I've added documentation at https://yihui.org/knitr/options/ and an example doc at https://github.com/yihui/knitr-examples/blob/master/121-ref-label.Rmd (further improvements are welcome).

atusy:
Thank you, @yihui, for improvements and finalizing!

PR #1963, opened 2021-02-10
created a `eng_r` function which exposes the R engine

abhsarma:
I created a separate engine for handling R code blocks, and exported it to make it accessible to users. This was useful in the implementation of our engine --- we are creating a package which allows end-users to declare multiverse analyses (multiple reasonable alternative statistical analysis pipelines) in a flexible and concise manner. Thus, our implementation resulted in a syntax which is not strictly R, but is processed into (multiple sets of) valid R syntax. We want to execute these generated R syntax using knitr to create rendered documents, which is not possible as current implementation does not expose this to the developer.

It can be useful to extend the R syntax --- developers can create custom engines which allows end-users to write commands/syntax which is not strictly in R but can then be processed into R output if necessary.

### Example:
```r
r_code <- process_code(options$code) # processes code which is not R to valid R syntax

eng_r(r_code) # outputs which have knitr formatting
```


yihui (commented):
It sounds good to me to factor out the R engine. I just want to make a slight modification, i.e., register this engine via `knit_engine$set(R = eng_r)` in `engine.R`. Then you don't need to export the function `eng_r`, but can access it via `knit_engine$get('R')`. Would that be okay to you?

BTW, if you use RStudio, please use `Ctrl + Shift + B` to generate the Rd files. Or just leave out changes in Rd. Thanks!

abhsarma:
Thanks for bringing that up. I've changed things so that `eng_r` is no longer exported and it works like you said. I also reset the head and tried to leave out changes in the Rd files, but I am not sure why there are still conflicts. Please let me know if I can do anything else

yihui (approved):
Looks perfect now. Thanks!

abhsarma:
Awesome thank you!

cderv:
That is awesome @abhsarma !
it was something I had in mind for a long time ! IMO it simplifies the reading / understanding of code and will allow to do more extension of R engine behavior ! 

Thanks you for this !

PR #1752, opened 2019-08-28
Allow for ids in figures

muschellij2:
This PR allows for Ids for `<div class="figure"`, which solves half of the problem in https://github.com/rstudio/bookdown/issues/766, which will allow for id references in bookdown for figures in `epub` documents.  An additional adaptation needs to be made in https://github.com/rstudio/bookdown/blob/master/R/html.R#L566, which I will make momentarily.  

Otherwise, this `reprex` should show that adding the `id` causes fewer `epub` errors:

# Clone the Repo

``` r
library(git2r)
library(bookdown)
library(knitr)
local_path = "bookdown-demo"
git2r::clone("https://github.com/rstudio/bookdown-demo.git",
             local_path = local_path)
#> cloning into 'bookdown-demo'...
#> Receiving objects:   1% (6/530),    8 kb
#> Receiving objects:  11% (59/530),   16 kb
#> Receiving objects:  21% (112/530),  120 kb
#> Receiving objects:  31% (165/530),  321 kb
#> Receiving objects:  41% (218/530),  409 kb
#> Receiving objects:  51% (271/530),  472 kb
#> Receiving objects:  61% (324/530),  544 kb
#> Receiving objects:  71% (377/530),  576 kb
#> Receiving objects:  81% (430/530),  584 kb
#> Receiving objects:  91% (483/530),  592 kb
#> Receiving objects: 100% (530/530),  723 kb, done.
#> Local:    master /private/var/folders/1s/wrtqcpxn685_zk570bnx9_rr0000gr/T/RtmpgLdL4z/reprexac002cc7d5c9/bookdown-demo
#> Remote:   master @ origin (https://github.com/rstudio/bookdown-demo.git)
#> Head:     [4e34630] 2018-10-22: Add now.json and Dockerfile for building HTML book and deploy to now.sh (#36)
setwd(local_path)


epub_file = bookdown::render_book(
  "index.Rmd",
  bookdown::epub_book(pandoc_args = "--reference-links"),
  clean = FALSE)
#> processing file: bookdown-demo.Rmd
#> output file: bookdown-demo.knit.md
#> /usr/local/bin/pandoc +RTS -K512m -RTS bookdown-demo.utf8.md --to epub3 --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash --output bookdown-demo.epub --reference-links --number-sections --filter /usr/local/bin/pandoc-citeproc
#> 
#> Output created: _book/bookdown-demo.epub
[...]

PR #1979, opened 2021-03-30
Allow for Image Ids

muschellij2:
This is an old PR that was killed about 1.5 years ago.  Here is the branch I had to fix https://github.com/yihui/knitr/pull/1752.  I'm not sure it sill is going to fix rstudio/bookdown#766 and it definitely needs to be squashed, but this is what I could do on a short turnaround @cderv


cderv:
Thanks!

muschellij2:
Not sure about the knitr_examples failure.

cderv:
Don't worry, we'll look into it.

lastthyme (approved):

cderv:
This has been added a bit differently in #2169 so I'll close this. 

We will see how we can leverage the new `fig.id` option to fix **bookdown** issue maybe. 

Thanks again

PR #1980, opened 2021-04-01
Add a bslib engine

cpsievert:
Adds a `{bslib}` engine, which provide a more convenient way to write Sass code that wants to leverage the [Bootstrap Sass](https://getbootstrap.com/docs/4.6/getting-started/theming/) behind `html_document_base`'s `theme` parameter. 

This approach works because `html_document_base()` sets the `theme` globally pre-knit when `theme` represents a `bslib::bs_theme()` object. When no global theme is set (i.e., `theme` isn't a `bs_theme()` object, or this code is running in some other document), then `bslib::bs_global_get()` return `NULL`, making the `sass::sass_partial(input, NULL)` equivalent to `sass::sass(input)`)

cc @cderv

Here's an example

````
---
title: "Say hello to the bslib theming engine"
output: 
  html_document:
    theme:
      version: 4
---

```{bslib, echo = FALSE}
p.my-class {color: $danger !important; }
```

<p class="my-class">
If this text appears red, then the `{bslib}` knitr engine is working!
</p>

The `{bslib}` knitr engine is a special version of `{scss}`/`{sass}` engine (it also expects Sass code), but has the added ability to reference `{bslib}`'s [theming options](https://rstudio.github.io/bslib/articles/bs4-variables.html) (aka Bootstrap Sass variables) when `theme` describes a [`bslib::bs_theme()`](https://rstudio.github.io/bslib/reference/bs_theme.html) object. 
````

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -742,14 +742,14 @@ eng_sxss = function(options) {
 
 eng_bslib = function(options) {
   if (!loadable("bslib")) {
-    stop("The {bslib} package must be installed in order for the {bslib} knitr engine to work.")
+    stop2("The 'bslib' package must be installed in order for the knitr engine 'bslib' to work.")
   }
   if (!is.null(options$engine.opts$sass_fun)) {
-    stop("The {bslib} knitr engine does not allow for customization of the Sass compilation function.", call. = FALSE)
+    stop2("The 'bslib' knitr engine does not allow for customization of the Sass compilation function.")
   }
-  func <- sass::sass_partial
-  formals(func)$bundle <- quote(bslib::bs_global_get())
-  options$engine.opts$sass_fun <- func
+  func = sass::sass_partial
+  formals(func)$bundle = quote(bslib::bs_global_get())
+  options$engine.opts$sass_fun = func
   eng_sxss(options)
 }
 
--- a/R/engine.R
+++ b/R/engine.R
@@ -738,7 +738,19 @@ eng_sxss = function(options) {
   }
 
   engine_output(options, options$code, final_out)
+}
 
+eng_bslib = function(options) {
+  if (!loadable("bslib")) {
+    stop("The {bslib} package must be installed in order for the {bslib} knitr engine to work.")
+  }
+  if (!is.null(options$engine.opts$sass_fun)) {
+    stop("The {bslib} knitr engine does not allow for customization of the Sass compilation function.", call. = FALSE)
+  }
+  func <- sass::sass_partial
+  formals(func)$bundle <- quote(bslib::bs_global_get())
+  options$engine.opts$sass_fun <- func
+  eng_sxss(options)
 }
 
 # set engines for interpreted languages
@@ -756,7 +768,8 @@ knit_engines$set(
   c = eng_shlib, cc = eng_shlib, fortran = eng_shlib, fortran95 = eng_shlib, asy = eng_dot,
   cat = eng_cat, asis = eng_asis, stan = eng_stan, block = eng_block,
   block2 = eng_block2, js = eng_js, css = eng_css, sql = eng_sql, go = eng_go,
-  python = eng_python, julia = eng_julia, sass = eng_sxss, scss = eng_sxss, R = eng_r
+  python = eng_python, julia = eng_julia, sass = eng_sxss, scss = eng_sxss, R = eng_r,
+  bslib = eng_bslib
 )
 
 cache_engines$set(python = cache_eng_python)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
I did not have a close look yet but as a general thoughts, would such engine be better in its related package ? Here **bslib** package. 

I know we had the sass engine used the sass package already but just throwing the idea. 

`bslib` engine sounds like something that could live in **bslib** directly. 

I believe the registering engine mechanism allows that. 

For users it could be also clearer to find docs of such engine in **bslib**.  

Just sharing thoughts here.

**Review by yihui — APPROVED** [review]
@cpsievert Looks good. Please add a news item. Thanks!

> would such engine be better in its related package ?

@cderv This question has been on my mind for a couple of years, and I run into it from time to time when people contribute new engines. I don't have a good answer to it. My criterion has been that if the engine is relatively simple, I'll just merge the PR. If it is contributed by someone that I'm familiar with or from RStudio, I can also merge it. In other cases, I may suggest registering these engines in other packages, such as https://github.com/yihui/knitr/pull/866.

For this case, I'm happy to merge it, although in hindsight, it's probably better to host the `sass` engine in the **sass** package (like `python` in **reticulate**). The other one that I wish could live outside **knitr** is the `sql` engine, since the original contributor has left RStudio, but it doesn't really burn me.

**Comment by cpsievert** [comment]
We can put this in `{bslib}` if you prefer, but I figured knitr would be a sensible place since the `{sass}` engine lives here and the implementation depends on that (I'd be happy to take a PR if you want to move the sass engine to the `{sass}` package). 

BTW, in addition to adding NEWS, I also added a check that `{bslib}` is `loadable()`

File: R/engine.R [hunk 1]
```diff
@@ -738,7 +738,19 @@ eng_sxss = function(options) {
...
+  if (!loadable("bslib")) {
+    stop("The {bslib} package must be installed in order for the {bslib} knitr engine to work.")
+  }
+  if (!is.null(options$engine.opts$sass_fun)) {
+    stop("The {bslib} knitr engine does not allow for customization of the Sass compilation function.", call. = FALSE)
+  }
+  func <- sass::sass_partial
+  formals(func)$bundle <- quote(bslib::bs_global_get())
+  options$engine.opts$sass_fun <- func
+  eng_sxss(options)
```
cderv, line 753: I think style is `=` instead of `<-` from other engine
```suggestion
  func = sass::sass_partial
  formals(func)$bundle = quote(bslib::bs_global_get())
  options$engine.opts$sass_fun = func
  eng_sxss(options)
```

cderv (approved):
> BTW, in addition to adding NEWS, I also added a check that {bslib} is loadable()

If this stays in **knitr**, it requires to add **bslib** to Suggest also. (as the **sass** package)

I don't have a strong opinion on where to have those engine. I was just wondering what the best solution is but both are working. I don't know if each engine are sensible to an update of their related package - if so it would be easier to update the engine if it was part of the package than coordinate a new knitr release. Apart from that, it is simple enough to have it here. 

@yihui if you want me to move the saas engine and bslib engine, I can look at that.

yihui:
> @yihui if you want me to move the saas engine and bslib engine, I can look at that.

Let's keep them here for now. It won't be a breaking change even if we move them out to **sass** and **bslib** in the future. Thanks!

R/engine.R
```diff
@@ -738,7 +738,19 @@ eng_sxss = function(options) {
...
 
   engine_output(options, options$code, final_out)
+}
 
+eng_bslib = function(options) {
+  if (!loadable("bslib")) {
+    stop("The {bslib} package must be installed in order for the {bslib} knitr engine to work.")
+  }
+  if (!is.null(options$engine.opts$sass_fun)) {
+    stop("The {bslib} knitr engine does not allow for customization of the Sass compilation function.", call. = FALSE)
```
yihui, line 748:
```suggestion
    stop2("The 'bslib' knitr engine does not allow for customization of the Sass compilation function.")
```

R/engine.R
```diff
@@ -738,7 +738,19 @@ eng_sxss = function(options) {
   }
 
   engine_output(options, options$code, final_out)
+}
 
+eng_bslib = function(options) {
+  if (!loadable("bslib")) {
+    stop("The {bslib} package must be installed in order for the {bslib} knitr engine to work.")
```
yihui, line 745:
```suggestion
    stop2("The 'bslib' package must be installed in order for the knitr engine 'bslib' to work.")
```

yihui (approved):

Issue #1919, opened 2020-11-15
Suggestion:  knitr::markdown vignette engine should check that `rmarkdown` is a dependency


dmurdoch:
When writing an R package containing a vignette using `knitr::rmarkdown`, it is a requirement to have `rmarkdown` as a suggested package.  Currently the `knitr::rmarkdown` engine detects if `rmarkdown` is not present and gives a warning or error, but it would be better if it could warn about the problem before it occurs:

 - when processing the vignette, find the base directory of the package it's in.  (I don't think there's any reason anyone would call the `knitr::rmarkdown` engine except when producing vignettes in a package, but maybe there is, so this may fail:  in which case don't worry about the second step.)

 - once you know the base directory of the package, examine the `DESCRIPTION` file to see if `rmarkdown` is listed as a dependency.  If not, issue a warning, whether or not `rmarkdown` is found on the current machine.

This would make it easy to find the error of omitting `rmarkdown` before submitting the package to CRAN.

The Writing R Extensions manual for R-devel currently contains some bogus advice about putting `rmarkdown` and `formatR` in the `VignetteBuilder` field.  I don't know the conditions under which `formatR` is needed, but this test could also apply to them.

yihui:
> when processing the vignette, find the base directory of the package it's in

I agree it would be better to notify the package author earlier, but I'm not sure about this "when" or how I could detect the absence of `rmarkdown` in DESCRIPTION _before_ the `knitr::rmarkdown` engine is actually run. Could you elaborate or perhaps send a PR if the implementation is simple enough? Thanks!

PR #1982, opened 2021-04-06
Package build will fail if rmarkdown is not declared as a dependency

dmurdoch:
... when there's an rmarkdown vignette.

This is too strong:  a warning would be better, but warnings appear to be suppressed in the package build code.  So at a minimum there needs to be a way to suppress this check.  I haven't added one, because I don't know your conventions for doing that.  Or maybe the doc warning is enough, or this check belongs in the RStudio template for an RMarkdown vignette.

This is what I was asking for in issue #1919 , but I'm not sure you'll want it...

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -107,6 +107,7 @@ Description: Provides a general-purpose tool for dynamic report generation in R
 Depends:
     R (>= 3.2.3)
 Imports:
+    desc,
     evaluate (>= 0.10),
     highr,
     methods,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -108,7 +108,6 @@ Description: Provides a general-purpose tool for dynamic report generation in R
 Depends:
     R (>= 3.2.3)
 Imports:
-    desc,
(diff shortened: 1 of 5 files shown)


yihui (commented):
Thanks! I see what you mean now (signal the error during `R CMD build` instead of `R CMD check`). Let me think more about it.

The documentation has been there for several years, but I doubt if anyone would find or read it: https://github.com/yihui/knitr/commit/77f94072f9a9752e0d60b3e12be995fe3938bc1b.

cderv:
FWIW, `usethis::use_vignette()` will take care of adding if missing **knitr** and **rmarkdown** as Suggests in DESCRIPTION. 

Using this function when building package is a great way to not forget to do it.
We could also communicate and advice that more to use `usethis::use_vignette()` so that all the requirement are correctly set without the user needing to worry about it. IMO this is one advantage of this package. 

> or this check belongs in the RStudio template for an RMarkdown vignette.

We could definitely add a comment in the template too. This would definitely push the information to the user.

yihui (approved):

PR #1987, opened 2021-04-20
Use parameterised query option in SQL chunk

### LINKED ISSUE #1867: SQL Chunk: Sending parameters through engine API
- [x ] I have fully read the issue guide at https://yihui.org/issue/.
- [x ] I have provided the necessary information about my issue.
- [ x] I have learned the Github Markdown syntax, and formatted my issue correctly.

 Stack overflow question: 
https://stackoverflow.com/questions/62466760/using-sql-chunks-in-rmarkdown-with-query-parameters

This is also raised agains [bigrquery](https://github.com/r-dbi/bigrquery/issues/391) package, but after a bit of research it seems this belongs to knitr. 

---

I am trying to work out if there is a way to pass parameters to the `sql` chunks connected to Google's [BigQuery](https://cloud.google.com/bigquery):

```
```{r setup}
bq_auth(path = "access_token.json")
db <- dbConnect(
  bigquery(), 
  dataset = 'my_data', 
  project = 'my-project', 
  use_legacy_sql = FALSE
)
parameter_value = c(10L, 20L)
```

```
```{sql, echo=FALSE, connection=db, output.var="x"}
SELECT UNNEST(@parameter_value)
```

```
```{r output}
print(x)
# I want to see 10, 20 here.
```

I understand that current implementation replaces `?value` placeholders with values from variables, but I think `sql` chunks should be able to pass parameters to the engine API and leverage platforms capabilities, instead of doing string magic inside this package:

* BigQuery: https://cloud.google.com/bigquery/docs/parameterized-queries
* Oracle & Postgres: I believe you can use `:` to add query parameters, see https://stackoverflow.com/questions/7389416/postgresql-how-to-pass-parameters-from-command-line
* MySQL: https://stackoverflow.com/questions/775296/mysql-parameterized-queries

byapparov:
closes #1867


byapparov:
@cderv I have tried this on my version of `bigrquery` package that supports `params` and it works as expected. 

It seems like quite a small change, but number of `if` statements that sql engine has is slightly dangerous, unless covered by unit tests. I don't think  `eng_sql()` function has any tests right now.

I would also recommend to deprecate DBI interpolation of parameters gradually for databased that support parameterised queries.

yihui (approved):
Thanks!

> but number of `if` statements that sql engine has is slightly dangerous

That's true :)

> I would also recommend to deprecate DBI interpolation of parameters gradually for databases that support parameterised queries.

That sounds like a good plan to me (but again, I don't think I have enough expertise to make the decision).

byapparov:
I have checked [sqlite](https://sqlite.org/lang_expr.html) documentation and it supports `@`, `:` and `$` named parameters: 

e.g.:
> An "at" sign works exactly like a colon, except that the name of the parameter created is @AAAA.

I will later update examples file with the params options, so we have some test coverage for that. 

Not sure how this two repositories sync, possibly I should wait till this is merged?

cderv:
> Not sure how these two repositories sync, possibly I should wait till this is merged?

Yeah, in fact they are not really sync between PR but are on master.  I need to improve that and implement an integration workflow for such cases. 

But you can still open a PR in knitr-example if you want to share / save your example file.

I'll have a look soon at this. You example file could help me for the review. Thanks.
I'll think also about the deprecation you mentioned

byapparov:
@cderv I have added example to `knitr-examples`. 

I am not sure how you generate `.md` files from `.Rmd`, so you might want to generate it your self.

It would be also helpful to update README with the method used to generate md for future contributors.

yihui (approved):
I think this PR can be merged after you add an item to NEWS.md (unless @cderv has any other feedback). Also feel free to add your name to `DESCRIPTION` as `ctb`. Thanks!

> It would be also helpful to update README with the method used to generate md for future contributors.

There is an ugly shell script `k` in the repo, and the examples are knitted with this shell script, e.g., `./k 115-engine-sql.Rmd`. Before documenting it in README, we'll need to make it less ugly :)

cderv (approved):

cderv:
That is ok with me. 

We should also update the documentation for the SQL engine in different places now so that this is documented. 

>  Before documenting it in README, we'll need to make it less ugly :)

Yes this workflow of adding test in knitr-example should be improved - and documented / automated at the right level. Currently it is not easy to setup - I had a hard time doing it if I remember correctly :) 
We'll improve that.

byapparov:
@yihui I have added a line to NEWS and added myself to DESCRIPTION.

Thank you for helping with this, looking forward to using this in my projects )

Issue #1935, opened 2020-12-19
FR: Allow collapse and strip.white


kbvernon:
I would like to request that the chunk options `collapse` and `strip.white` be allowed to work together. Currently, this is not possible. I'm told this is a result of a line in `wrap.source()` ([Line 481](https://github.com/yihui/knitr/blob/046ae373e674f2e47e4d83a19d61fe46ab9ca4b5/R/output.R#L481)). 

I asked this question on the RStudio Community forum [here](https://community.rstudio.com/t/code-chunk-collapse-true-adds-unwanted-line-to-code-block-output/91060/2). 

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.org/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('knitr')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('yihui/knitr')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
- [x] I have learned the Github Markdown syntax, and formatted my issue correctly.

I understand that my issue may be closed if I don't fulfill my promises.

yihui:
Could you remove the condition `!options$collapse && ` and submit a PR? I don't remember why I had that condition now. If it can passes the checks without the condition, I can probably get rid of it. Thanks! https://github.com/yihui/knitr/blob/046ae373e674f2e47e4d83a19d61fe46ab9ca4b5/R/output.R#L481

PR #2011, opened 2021-06-03
allow collapse and strip white as per https://github.com/yihui/knitr/issues/1935

kbvernon:
removed `!options$collapse &&` from `sew.source()`

I read the contributing guidelines. Thanks for including that. Not sure this rises above "non-trivial," so I didn't mess with anything else. Hopefully, it passes your checks.

Cheers,


yihui (approved):
Looks good (all tests have passed). Thank you!

I'll merge this for now, but I will have to run reverse dependency checks when I prepare the next CRAN release. I don't know if I'll discover problems at that time. I'll try my best to keep this change even if this happens to be a breaking change to other packages.

PR #2015, opened 2021-06-10
`include_*`

### LINKED ISSUE #2012: include_url results in enc2utf8(x): argument is not a character vector
When calling knitr::include_url("https://www.yihui.org") in a Rmd notebook i get the following error:

```
Error in enc2utf8(x) : argument is not a character vector 

Traceback:
3. writeLines(enc2utf8(x), con = mdfile, useBytes = TRUE) 
2. print.knit_asis(x) 
1. (function (x, ...) 
     UseMethod("print"))(x) 
```

I'm running:

| Environment|Version |
| ---:|:---|
knitr | 1.33
platform  | x86_64-apple-darwin17.0                          
version.string  | R version 4.0.5 (2021-03-31)

---

cderv:
This close #2012 and contribute to resolve https://github.com/rstudio/rstudio/issues/6476

The IDE use `print` in interactive mode to output a result. `print.knit_asis` is one of the method overriden. This creates problem for `include_url()` related function that are not coercible to character. So it creates an error. 

This PR use a new special `knit_asis_url` class for these function and the IDE use a default `print` for them. 

* Do we want a special print ? 

* Also are those function supposed to be used inlined ? There is a special handling for this in 
https://github.com/yihui/knitr/blob/89a1f6cc19e060c0f45b80fe1c4905dfab441193/R/block.R#L530
Maybe we knit to check for the new class too ? At least to keep the same behavior.

I left this PR as draft for now until we decide.


R/output.R
```diff
@@ -699,6 +699,9 @@ knit_print.default = function(x, ..., inline = FALSE) {
 #' @export
 knit_print.knit_asis = function(x, ...) x
```
yihui, line 700: Can roxygen2 recognize two function names if we write the code this way?

```suggestion
knit_print.knit_asis = knit_print.knit_asis_url = function(x, ...) x
```

If not, adding a new function on a separate line is fine to me.

yihui (approved):
> * Do we want a special print ?

I don't know. Since we don't have `print.knit_asis`, perhaps we don't need `print.knit_asis_url`, either.

> * Also are those function supposed to be used inlined ?

Although I guess this use case should be rare, it seems we should support inlined URLs, too, i.e., `inherits(res, c('knit_asis', 'knit_asis_url'))`.

Thanks!

R/output.R
cderv: It seems like roxygen2 does not support it. The S3 method is not exported in NAMESPACE in this case.

cderv:
> I don't know. Since we don't have print.knit_asis, perhaps we don't need print.knit_asis_url, either.

Yes you're right. I have thought one time about adding a `print.knit_asis` but if we do, then this is not for this PR.

cderv:
> Although I guess this use case should be rare, it seems we should support inlined URLs,

Yes I think so. This is now done and allows to write like this 

````markdown

See the app below

`r knitr::include_app("https://yihui.shinyapps.io/miniUI/")`

This is an example app you can play with
````

And it works ok.

PR #2016, opened 2021-06-11
Use `conditionMessage()` to render conditions

davidchall:
When following the [error constructor design pattern](https://design.tidyverse.org/err-constructor.html), we might have condition objects with an empty message field. Instead, their message is constructed via the `conditionMessage()` S3 method. Currently, such condition messages are omitted from the {knitr} output.

`sew.error()` doesn't need fixing because `as.character.error()` internally calls `conditionMessage()` ([see here](https://github.com/wch/r-source/blob/77b3c6f6c8f7615af5bf0dda5e6ec898a54ba087/src/library/base/R/conditions.R#L148-L155)).

This the same problem fixed by https://github.com/r-lib/downlit/pull/100.

### DIFF
--- a/R/output.R
+++ b/R/output.R
@@ -538,7 +538,7 @@ sew.warning = function(x, options, ...) {
 
 #' @export
 sew.message = function(x, options, ...) {
-  msg_wrap(paste(x$message, collapse = ''), 'message', options)
+  msg_wrap(paste(conditionMessage(x), collapse = ''), 'message', options)
 }
 
 #' @export
--- a/R/output.R
+++ b/R/output.R
@@ -533,7 +533,7 @@ sew.warning = function(x, options, ...) {
     call = deparse(x$call)[1]
     if (call == 'eval(expr, envir, enclos)') '' else paste(' in', call)
   }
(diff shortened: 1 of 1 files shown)


yihui (approved):
Looks good. Thanks!

davidchall:
@yihui - thanks for the ✅ !

I'm trying to debug a whitespace issue caused by this change. In the past, the following code:

```r
my_func <- function() {
  message("hello")
  TRUE
}
my_func()

# i did a message
```

would output:
```r
my_func <- function() {
  message("hello")
  TRUE
}
my_func()
#> hello
#> [1] TRUE

# i did a message
```

But now it outputs:
```r
my_func <- function() {
  message("hello")
  TRUE
}
my_func()
#> hello
#> [1] TRUE
# i did a message
```

davidchall:
Ah please ignore that last message. It turns out it was an issue with my local installation. Even if I remove the changes of this PR, the whitespace is gobbled up. It not possible this is caused by `conditionMessage()` because it simply returns `x$message` for warning and message objects.

Issue #2032, opened 2021-07-29
Setting message=FALSE chunk opts suppresses fig.cap for PDF output


dchiu911:
I'm not sure when this started, but I've tried with the devel version of `knitr` and the issue persists: setting `knitr::opts_chunk$set(message = FALSE)` suppresses figure captions when the output is PDF.

````r
---
title: "Untitled"
date: "7/29/2021"
output: pdf_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = FALSE, message = FALSE)
```

```{r test, fig.cap=caps}
library(ggplot2)
caps <- c("another caption", "and yet an other")
qplot(1:5)
qplot(6:10)
```
````

And it is important for me to have `message=FALSE` to suppress messages such as
``` `stat_bin()` using `bins = 30`. Pick better value with `binwidth`.```

**Comment by atusy** [comment]
Reproduced.

With `keep_md: true`, I see it happens because images are aligned inline like below.
They should be separated by blank lines.
I'll take a look.

```
![another caption](example_files/figure-latex/test-1.pdf) ![and yet an other](example_files/figure-latex/test-2.pdf) 
```

**Comment by atusy** [comment]
I think this is the minimal reproducible example.

````
---
output: pdf_document
---

```{r test, fig.cap=caps, echo=FALSE}
caps <- c("another caption", "and yet an other")
plot(1:5)
plot(6:10)
```
````

**Comment by atusy** [comment]
I also find that `html_document(keep_md = TRUE)` has the same problem, which is inconsistent to `keep_md = FALSE`.

````
---
output:
  html_document:
    keep_md: true
---

```{r test, fig.cap=caps, echo=FALSE}
caps <- c("another caption", "and yet an other")
plot(1:5)
plot(6:10)
```
````

atusy:
and also in `md_document`

cderv:
> I also find that html_document(keep_md = TRUE) has the same problem, which is inconsistent to keep_md = FALSE.

@atusy, I am curious about this. Did you found why we get different result if `keep_md` is set to `TRUE` or `FALSE` ?

atusy:
That is because the `w` below is `NULL` when `keep_md = TRUE`, whereas `w` becomes 672 when `keep_md = FALSE`.
That results in `keep_md = TRUE` to generate markdown output and `keep_md = FALSE` to generate HTML output.
The value 672 is equal to 7 inch times 92 dpi, and comes from the default values of `html_document`.

[...]

PR #2033, opened 2021-08-05
add linebreaks before and after figures

atusy:
to solve #2032

### DIFF
--- a/R/hooks-md.R
+++ b/R/hooks-md.R
@@ -63,7 +63,7 @@ hook_plot_md_base = function(x, options) {
     res = sprintf('![%s](%s%s)', cap, base, .upload.url(x))
     if (!is.null(lnk) && !is.na(lnk)) res = sprintf('[%s](%s)', res, lnk)
     res = paste0(res, if (nocap) '<!-- -->' else '', if (is_latex_output()) ' ' else '')
-    return(res)
+    return(add_linebreak(res, options))
   }
   add_link = function(x) {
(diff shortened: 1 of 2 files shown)


atusy:
I think we need 2 linebreaks before and after the plots so that they become individual blocks.
But this certainly breaks the test below.
Is it okay to update the test? @yihui

```
Error: (img_output("a.png") %==% "![](a.png)") is not TRUE
```

https://github.com/yihui/knitr/pull/2033/checks?check_run_id=3251862194#step:11:137

yihui:
A breaking change is sometimes okay. I'm totally fine with breaking tests in this package. The much trickier and harder problem is to deal with is when the change breaks tests in other people's packages, over which I don't have control.

We can try and see how bad the breaking change is. Thanks!

cderv:
I think this could break some outputs if this is no more possible to generate inline plots. The HTML output will be different (using inline `img` vs using figure environment) and this could break some specific styling maybe. As often with rmarkdown changes, I don't think this will concern R developer that have tests, but more users with their outputs. 

Also, I am trying to understand why we are finding this now. It feels to me like there is a regression somewhere. The current code here seems to assume that when `nocap` and inline image should be created. 
https://github.com/atusy/knitr/blob/46a48b1333e56802eab49851d861ca67b2929bdf/R/hooks-md.R#L58-L62

I did look deeply through it yet but this gets me thinking 😅 

Thanks for looking into it @atusy !

atusy:
Closing as this PR is the duplicate of #1760

PR #2018, opened 2021-06-17
utils-vignettes: Don't disable vtangle if .Rout.save exists

lentinj:
When inside CMD check, vtangle() returns an empty file to avoid
unnecessary code execution (that already ran as part of the weave), and
prevent vignettes that don't tangle correctly breaking check output.
However, this output is needed if you're trying to compare tangled
output to an ".Rout.save".

This whole code block is nearly redundant, CMD check will perform
similar checks and not tangle, but only if
_R_CHECK_VIGNETTES_SKIP_RUN_MAYBE_ is enabled, which it currently isn't
by default.

Fixes #2017

Issue #1344, opened 2017-01-09
purl does not output inline code


kforner:
Hi,

I noticed that purl() does not seem to extract the inline R expressions. 
It would be extremely useful, at least for me, to have access to all R code included in a document, at least as knitr recognize it. The use case is code analysis for instance. 

With documentation == 2, the inline code is found, but commented out, so it is difficult to distinguish it from other comments.

From what I understood from the code, it seems that it could be done by modifying a bit this function https://github.com/yihui/knitr/blob/115e994b332c2d36a598a6ea12ddb639478c266b/R/block.R#L462 

```
process_tangle.inline = function(x) {
  if (opts_knit$get('documentation') == 2L) {
    return(paste(line_prompt(x$input.src, "#' ", "#' "), collapse = '\n'))
  }
  code = x$code
  if (length(code) == 0L || !any(idx <- grepl('knit_child\\(.+\\)', code)))
    return('')
  paste(c(sapply(code[idx], function(z) eval(parse_only(z))), ''), collapse = '\n')
}
```

and more particularly this line: 
```
  if (length(code) == 0L || !any(idx <- grepl('knit_child\\(.+\\)', code)))
    return('')
```
Changing it to something like "`return(as.character(code))`"  should probably do the trick. 
Would you accept a Pull Request ? 

Thanks.

yihui:
I don't remember why I excluded inline code, but if I were to make this change, I will have to do a reverse dependency check first to make sure no CRAN packages will be broken. This is a rather tedious job, and I die a little inside every time I do it.

In general, inline code is supposed to be a single variable or simple expression writing out a single value. I'm not sure how important it is to your code analysis.

kforner:
> I don't remember why I excluded inline code, but if I were to make this change, I will have to do a > reverse dependency check first to make sure no CRAN packages will be broken. This is a rather tedious job, and I die a little inside every time I do it.

I perfectly understand the problem. Could we implement the new behaviour via an option ? Either adding a value to documentation (but that is a little hacky), or adding a new option/param in purl ? 

> In general, inline code is supposed to be a single variable or simple expression writing out a single value. I'm not sure how important it is to your code analysis.

The problem is that you can have calls to knit_child(), that are critical to know that the document has children, and you can also have calls to functions in other packages, that one could need to track.

 I could work-around it, but I suppose that if I need this, it could also benefit to other people. It is nice to be able to reuse the exact knitr parsing engine.

yihui:
[...]

Issue #1093, opened 2015-08-21
Do not skip tangle in future versions of R


yihui:
i.e. https://github.com/yihui/knitr/blob/01ba6758d6409912aee579e28f627ec1aff5dff4/R/utils-vignettes.R#L46-L50

CRAN said they would skip the tangle check via `R CMD check --as-cran` (unless there is .R.out.save).

Issue #784, opened 2014-05-29
Inline code chunks, tangle, and R CMD check errors


cboettig:
It seems that R CMD check has decided to tangle knitr vignettes and source the resulting `.R` script as way of checking the R code even when `--no-build-vignettes` is selected.  

This can create a problem since `tangle` ignores the inline chunks, and yet inline chunks can modify the R code.  As a consequence it is possible to have a vignette that compiles perfectly fine when knit directly, but does not execute successfully when `tangle`'d and then `source`'d. 

 (Yes, I realize they are primarily designed for just echoing values, not setting them, but for instance, in `knitcitations` my inline calls to things like `citep` change an environmental variable).  

Perhaps tangle should not ignore inline blocks?

PR #2036, opened 2021-08-19
drop obsolete vtangle hack for R CMD check

### LINKED ISSUE #2017: .Rout.save files empty when used with knitr vignettes
If I try and create an ``.Rout.save`` file to go with a knitr vignette, ``R CMD check`` will always compare against an empty file, for example:

```
# NB: Non-empty vignette
$ ls -l vignettes/faq.Rmd
-rw-r--r-- 1 lentinj lentinj 4060 Jun 16 15:47 vignettes/faq.Rmd
echo '[Dummy Rout.save]' > vignettes/faq.Rout.save
$ R CMD check . . .
     . . .
* checking running R code from vignettes ...
  ‘faq.Rmd’ using ‘UTF-8’... NOTE
differences from ‘faq.Rout.save’
1,2c1
<
<  *** Run successfully completed ***
---
> [Dummy Rout.save]
```

This is due to the following test, that disables tangling when inside CMD check, which was done AIUI since [not all documents will successfully tangle](https://r.789695.n4.nabble.com/R-CMD-check-for-the-R-code-from-vignettes-td4691457.html):

https://github.com/yihui/knitr/blob/11eeb4326b6ed1afd6808b29ef0fb3ed78762ea6/R/utils-vignettes.R#L41-L48

As I see it, options are:

1. Remove the ``if (is_R_CMD_check()) {  ... }`` block entirely. According to #1093, this test can be removed once CMD check only tangles if an ``.Rout.save``. Which seems to be implemented [with this commit](https://github.com/wch/r-source/commit/39e369fe5a808fe9451ece97c24d9149c1a72f62?w=1).
2. Be more conservative, change ``vtangle``() to also notice the ``.Rout.save`` and continue tangling in that case. For example:-

```
--- a/R/utils-vignettes.R
+++ b/R/utils-vignettes.R
@@ -39,7 +39,9 @@ vweave = function(file, driver, syntax, encoding = 'UTF-8', quiet = FALSE, ...)
 }

 vtangle = function(file, ..., encoding = 'UTF-8', quiet = FALSE) {
-  if (is_R_CMD_check()) {
+  if (is_R_CMD_check() && !file.exists(with_ext(file, 'Rout.save'))) {
+    # Inside CMD check, and no corresponding .Rout.save. No point doing tangle,
+    # since code was already run as part of weave operation.
     file = with_ext(file, 'R')
     file.create(file)
     return(file)
```

Happy to send a pull request implementing either, if it's useful!

Many thanks,

---

bastistician:
Following up on #2018, this corresponds to "Option 1" to fix #2017 (and #1093).
I think tampering with R CMD check in `vtangle()` has long been obsolete, because knitr provides

- chunk option `purl=FALSE` following #519
- *_notangle vignette engines following #784
- option "knitr.purl.inline" following #1344

to address potential (hopefully rare) issues with R code extracted from vignette sources.

### DIFF
--- a/R/utils-vignettes.R
+++ b/R/utils-vignettes.R
@@ -39,15 +39,6 @@ vweave = function(file, driver, syntax, encoding = 'UTF-8', quiet = FALSE, ...) 
 }
 
 vtangle = function(file, ..., encoding = 'UTF-8', quiet = FALSE) {
-  if (is_R_CMD_check() && !file.exists(with_ext(file, 'Rout.save'))) {
-    # Inside CMD check, and no .Rout.save file to compare with, so return an
-    # empty script to avoid errors from vignettes that won't tangle correctly.
-    # NB: This is ~equivalent to setting _R_CHECK_VIGNETTES_SKIP_RUN_MAYBE_=TRUE
-    # and if that becomes the default this is redundant.
-    file = with_ext(file, 'R')
-    file.create(file)
-    return(file)
-  }
   purl(file, encoding = encoding, quiet = quiet, ...)
 }
 


yihui (approved):
`_R_CHECK_VIGNETTES_SKIP_RUN_MAYBE_ ` is still `FALSE` by default: https://github.com/wch/r-source/blob/6a086aea7afa3365173c180574d065288f462772/src/library/tools/R/check.R#L6272-L6274 unless the `--as-cran` flag is turned on: https://github.com/wch/r-source/blob/6a086aea7afa3365173c180574d065288f462772/src/library/tools/R/check.R#L6351

That means a normal `R CMD check` can still cause problems. I'm not sure how many people run `R CMD check` without the `--as-cran` flag. Maybe you are right that this is rare.

I'll consider merging this in a future version of **knitr**. Thanks!

yihui:
FYI I had to revert this change due to problems reported in https://github.com/yihui/knitr/issues/2052#issuecomment-929646480.

bastistician:
I think the purpose of

> * checking running R code from vignettes ...

is to check that `R CMD Stangle` produces an R script that runs without error. In principle, this is a useful test.

In the example provided in #2052

````
```{r, include = FALSE}
knitr::opts_chunk$set(eval = FALSE)
```
```{r}
stop("I shouldn't run!")
```
````

this test fails because knitr's tangle function, `purl()`, cannot know anything about dynamically modified chunk options. I'm sure there are more examples where `purl` won't produce a useful R script because of that.

However, in my opinion, **knitr** should not generally torpedo this check (as it did before 1.35), but allow package developers to have their tangled vignette code run during R CMD check and get notified when `purl()` produces a broken R script.

If I don't want to run vignette code in a specific environment, I can always check with `--no-vignettes`. To check vignette building but skip running the R script produced by the tangle function, I could set `_R_CHECK_VIGNETTES_SKIP_RUN_MAYBE_=TRUE` as is done by `--as-cran` (although I see that this environment variable is actually undocumented).

If a vignette cannot be fixed to produce an executable R script via tangle (for which I have yet to see a convincing example), shouldn't it then explicitly use a *_notangle engine? It might then also be useful if the tangle functions of these notangle-engines produced R scripts with (only) a comment line explaining that this is a "notangle" vignette whose code can only be run via `knit()`.

bastistician:
> check that `R CMD Stangle` produces an R script that runs without error. In principle, this is a useful test.

It might actually be less useful in the case of knitr-based package vignettes, as the installed R script (the one eventually linked from `browseVignettes()`) will not come from "tangle" but is generated during "weave" when the package is built:
https://github.com/yihui/knitr/blob/563565bb5ee7fae3986f78334c854a7d9d6dcaae/R/output.R#L173-L178
I'm wondering if it might make more sense for the "checking running R code from vignettes" step to run the R script provided to the user via inst/doc rather than a newly generated tangle output. Such a change requires some thought (especially because .Rout.save stuff is involved). As it stands, this step checks R CMD Stangle compatibility of the vignette source.

yihui:
Thanks for your feedback! There is a history behind it:

> I'm wondering if it might make more sense for the "checking running R code from vignettes" step to run the R script provided to the user via inst/doc rather than a newly generated tangle output.

That's exactly my wish. If only you could convince R core of running the R script provided by package author, instead of tangling a new one, happy tears would be rolling down my face. I pointed it out to them that tangle might not be reliable and weave is much more trustworthy in 2014, but no one was with me at that time:
https://github.com/yihui/knitr/blob/04d23d2e189d4ebcc2f7c3f37ae673c9733d82ae/R/output.R#L101-L110

> If I don't want to run vignette code in a specific environment, I can always check with `--no-vignettes`. To check vignette building but skip running the R script produced by the tangle function, I could set `_R_CHECK_VIGNETTES_SKIP_RUN_MAYBE_=TRUE` as is done by `--as-cran` (although I see that this environment variable is actually undocumented).
> 
> If a vignette cannot be fixed to produce an executable R script via tangle (for which I have yet to see a convincing example), shouldn't it then explicitly use a *_notangle engine? It might then also be useful if the tangle functions of these notangle-engines produced R scripts with (only) a comment line explaining that this is a "notangle" vignette whose code can only be run via `knit()`.

Everything you said makes perfect sense. The only problem is that these are all extra work for package authors, and they may not know how to do it, or may not be willing to do it. Every surprised author may come to me and ask why the script is broken (only during `R CMD check`). This is going to be a forever confusion. IMHO, hiding the problem from `R CMD check` is the only way to peace for everyone.

bastistician:
Interesting. I was happy to read your opinion:

> Please also note I mean this is _optional_, i.e. package authors can _choose_ whether they want to disable tangle.

in https://stat.ethz.ch/pipermail/r-devel/2014-June/069143.html as this supports mine: ;)

> However, in my opinion, knitr should not generally torpedo this check (as it did before 1.35), but allow package developers to have their tangled vignette code run during R CMD check and get notified when purl() produces a broken R script.

PR #2047, opened 2021-09-21
Lengthen fence characters when needed (fix #1621)

### LINKED ISSUE #1621: knit() cannot handle variable numbers of backticks
(Originally reported here: https://github.com/rstudio/rmarkdown/issues/1444#issuecomment-428504984)

`spin()` now uses various number of backticks (#1611), but currently `knit()` cannot handle them.

````` r
tmp_rmd <- tempfile(fileext = ".Rmd")
tmp_md <- tempfile(fileext = ".md")

writeLines(
  c('````{r}',
    'x <- "',
    '```',
    '"',
    '````'),
  con = tmp_rmd
)

cat(readLines(tmp_rmd), sep = "\n")
#> ````{r}
#> x <- "
#> ```
#> "
#> ````

res <- knitr::knit(tmp_rmd, output = tmp_md, quiet = TRUE)
cat(readLines(tmp_md), sep = "\n")
#> 
#> ```r
#> x <- "
#> #> Error: <text>:1:6: unexpected INCOMPLETE_STRING
#> #> 1: x <- "
#> #>          ^
#> ```
#> "
#> ````
`````

<sup>Created on 2018-10-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

To determine how many backticks are used to represent the end of a chunk, we need to look at the beginning of the chunk (in actual, the rule is [a bit more complex](https://spec.commonmark.org/0.28/#example-93)). But, in the current implementation of `split_file()`, begin chunks and end chunks are matched separately, so it seems difficult to fix without introducing a big change to the current implementation.

https://github.com/yihui/knitr/blob/3abb642649d186ba0a42d7e489ea3babd913b023/R/parser.R#L15-L16

### PR DESCRIPTION
**Pull request by atusy**
Fixes #1621.

- [x] Update source
- [x] Add tests
- [x] Update NEWS
- [x] pass CI

## Reprex

``` r
rmd = tempfile(fileext = ".Rmd")
c(
  "````{r}",
  "invisible('",
  "```",
  "')",
  "````"
) |> writeLines(rmd)

knitr::knit(rmd, tempfile(), quiet = TRUE) |>
  readLines() |>
  cat(sep = "\n")
```

``` #>
#> ````r
#> invisible('
#> ```
#> ')
#> ````
```

<sup>Created on 2021-09-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/parser.R
+++ b/R/parser.R
@@ -400,38 +400,17 @@ parse_chunk = function(x, rc = knit_patterns$get('ref.chunk')) {
 
 # filter chunk.end lines that don't actually end a chunk
 filter_chunk_end = function(chunk.begin, chunk.end, lines = NULL, patterns = NULL) {
-  keys = c('chunk.begin', 'chunk.end')
-  if (identical(patterns[keys], all_patterns[['md']][keys])) {
-    filter_chunk_end_md(chunk.begin, chunk.end, lines)
-  } else {
-    filter_chunk_end_general(chunk.begin, chunk.end)
-  }
-}
-
-filter_chunk_end_general = function(chunk.begin, chunk.end) {
   in.chunk = FALSE
-  fun = function(is.begin, is.end) {
-    if (in.chunk && is.end) {
-      in.chunk <<- FALSE
-      return(TRUE)
-    }
-    if (!in.chunk && is.begin) in.chunk <<- TRUE
-    FALSE
-  }
-  mapply(fun, chunk.begin, chunk.end)
-}
-
-filter_chunk_end_md = function(chunk.begin, chunk.end, lines) {
-  in.chunk = FALSE
-  expected_end = NA_character_
+  is.md = identical(patterns, all_patterns[['md']])
+  pattern.end = NA_character_
   fun = function(is.begin, is.end, line) {
-    if (in.chunk && is.end && grepl(expected_end, line)) {
+    if (in.chunk && is.end && (is.na(pattern.end) || grepl(pattern.end, line))) {
       in.chunk <<- FALSE
       return(TRUE)
     }
     if (!in.chunk && is.begin) {
       in.chunk <<- TRUE
-      expected_end <<- paste0(sub('(^[\t >]*```+).*', '\\1', line), "\\s*$")
+      if (is.md) pattern.end <<- paste0(sub('(^[\t >]*```+).*', '\\1', line), "\\s*$")
     }
     FALSE
   }
--- a/R/parser.R
+++ b/R/parser.R
@@ -505,10 +505,10 @@ parse_chunk = function(x, rc = knit_patterns$get('ref.chunk')) {
 }
 
 # filter chunk.end lines that don't actually end a chunk
-filter_chunk_end = function(chunk.begin, chunk.end, lines = NA_character_, patterns = NA_character_) {
+filter_chunk_end = function(chunk.begin, chunk.end, lines = NA, patterns = list()) {
   in.chunk = FALSE
   is.md = identical(patterns, all_patterns[['md']])
-  pattern.end = NA_character_
+  pattern.end = NA
   fun = function(is.begin, is.end, line) {
     if (in.chunk && is.end && (is.na(pattern.end) || grepl(pattern.end, line))) {
       in.chunk <<- FALSE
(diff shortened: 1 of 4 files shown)


atusy:
For some reason following output occurs in https://github.com/yihui/knitr-examples/blob/master/115-engine-sql.Rmd

`````
````r
readrPackage
``` 
````

```
## Error: attempt to use zero-length variable name
```
`````

I do not understand why, but the error disappears if I trim following lines
https://github.com/yihui/knitr-examples/blob/08cf670e3296b5febffb454bad7b5acf5f1f6295/115-engine-sql.Rmd#L71-L77

I have to investigate more.

**Comment by atusy** [comment]
It seems the error happens when duplicated chunks exist.

``` r
rmd = tempfile(fileext = ".Rmd")
c(
  "```{r}", "x=1", "```",
  "",
  "```{r}", "x=1", "``` ",
  "",
  #"```{r}", "1", "```",
  NULL
) |> writeLines(rmd)
cat(readLines(knitr::knit(rmd, tempfile())), sep = "\n")
#> processing file: /tmp/RtmppEyr1p/file37db7e54a1b1.Rmd
#> output file: /tmp/RtmppEyr1p/file37db661c6a48
#> 
#> ```r
#> x=1
#> ```
#> 
#> 
#> ````r
#> x=1
#> ``` 
#> ````
#> 
#> ```
#> #> Error: attempt to use zero-length variable name
#> ```
```

<sup>Created on 2021-09-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

atusy:
The previous post is wrong.
The problem happens if chunk is ending with backticks followed by space(s).

``` r
rmd = tempfile(fileext = ".Rmd")
c(
  "```{r}", "x = 1", "``` ", ""
) |> writeLines(rmd)
cat(readLines(knitr::knit(rmd, tempfile())), sep = "\n")
#> processing file: /tmp/RtmpdRpzv0/file6f376822dca7.Rmd
#> output file: /tmp/RtmpdRpzv0/file6f374be0aa19
#> 
#> ````r
#> x = 1
#> ``` 
#> ````
#> 
#> ```
#> #> Error: attempt to use zero-length variable name
#> ```
```

<sup>Created on 2021-09-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

atusy:
Yes, CI is green! Now, I'm ready for the review!!

R/parser.R
```diff
@@ -399,7 +399,16 @@ parse_chunk = function(x, rc = knit_patterns$get('ref.chunk')) {
 }
 
 # filter chunk.end lines that don't actually end a chunk
-filter_chunk_end = function(chunk.begin, chunk.end) {
+filter_chunk_end = function(chunk.begin, chunk.end, lines = NULL, patterns = NULL) {
+  keys = c('chunk.begin', 'chunk.end')
+  if (identical(patterns[keys], all_patterns[['md']][keys])) {
```
yihui, line 404: In this case, I hope to update `chunk.end` and turn its elements to `FALSE` if the matched lines are not really the end of code chunks (i.e., if they do not match the number of backticks in `chunk.begin`).

After `chunk.end` truly represents chunk endings, we can use the original `filter_chunk_end()` instead of calling `filter_chunk_end_md()`, because the latter contains duplicated code from `filter_chunk_end()`, which I hope to avoid.

Could you try that?

yihui (approved):
This implementation is okay. I just wish to avoid the repetition of code in `filter_chunk_end_general()` and `filter_chunk_end_md()` if possible. Thank you!

R/parser.R
atusy: I hope my simplification is fine which employs an approach different from yours.

If I understand right, updating `chunk.end` is the exactly what is done by `filter_chunk_end*()`.
It means calling `filter_chunk_end` after updating the `chunk.end` will attempts unnecessary filtering. 
This is why I chose the different approach.

https://github.com/yihui/knitr/pull/2047/files#diff-92c9f6892b4000a9291d262dbbdb4519edfc0437c89e7eeb686d84c1ee13119eL402-R417

atusy:
@yihui Thanks for the review! I updated the codes based on it.

R/parser.R
yihui: Perfect! c302685 achieved exactly what I wanted, although it was a different method. Thank you!

PR #2048, opened 2021-09-22
Set default `strip.white` so that we modify later conditionally

### LINKED ISSUE #2046: Change default `strip.white` value when `collapse = TRUE`
Following #2011, a breaking change was reported in https://github.com/rstudio/rmarkdown/issues/2220

Basically, when `collapse = TRUE`, blank lines are stripped by default because `strip.white = TRUE` is the default, and it applies now on those chunk. 

We probably want to make `strip.white = FALSE` the default when `collapse = TRUE` if possible.

cderv:
This close #2046 and fix the regression mentioned in rstudio/rmarkdown#2220

* Options is set to `NA` by default (i.e `strip.white = NA`)
* Option default is then "fixed" in `fix_options()` where we adjust according to other chunk option, global or per chunk. 

This is the first option where we set a default value that will be conditional to another one, defined by user. 

I added a unit test for this, and I guess we'll need a knitr-example test for this to have snapshot test of the output. It would have prevented the breakage in the first place. However, not as easy as with testthat snaphot test to add a new test in knitr-example synced with this PR - I need to improve the system there. 

Anyway, I think this will work ok now and don't think this change will impact other stuff.

### DIFF
--- a/R/defaults.R
+++ b/R/defaults.R
@@ -67,10 +67,12 @@ new_defaults = function(value = list()) {
 #' @export
 #' @examples opts_chunk$get('prompt'); opts_chunk$get('fig.keep')
 opts_chunk = new_defaults(list(
-
   eval = TRUE, echo = TRUE, results = 'markup', tidy = FALSE, tidy.opts = NULL,
   collapse = FALSE, prompt = FALSE, comment = '##', highlight = TRUE,
-  strip.white = TRUE, size = 'normalsize', background = '#F7F7F7',
+  size = 'normalsize', background = '#F7F7F7',
+
+  # NA value means will set the default option later (e.g in fix_options)
+  strip.white = NA,
 
   cache = FALSE, cache.path = 'cache/', cache.vars = NULL, cache.lazy = TRUE,
   dependson = NULL, autodep = FALSE, cache.rebuild = FALSE,
--- a/R/defaults.R
+++ b/R/defaults.R
@@ -71,8 +71,9 @@ opts_chunk = new_defaults(list(
   collapse = FALSE, prompt = FALSE, comment = '##', highlight = TRUE,
   size = 'normalsize', background = '#F7F7F7',
 
-  # NA value means will set the default option later (e.g in fix_options)
-  strip.white = NA,
+  # value wrapped in I() means a change in default option later
+  # (e.g in fix_options, conditionally to other argument)
+  strip.white = I(TRUE),
 
   cache = FALSE, cache.path = 'cache/', cache.vars = NULL, cache.lazy = TRUE,
   dependson = NULL, autodep = FALSE, cache.rebuild = FALSE,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # CHANGES IN knitr VERSION 1.35
 
+- Fix a regression in 1.34: Blank lines are now more stripped by default when `collapse = TRUE`, only when `collapse = FALSE`. If you liked the new behavior, set `strip.white = TRUE` globally or on per chunk basis (thanks, @IndrajeetPatil, rstudio/rmarkdown#2220, #2046).
+
 - In `knitr::combine_words()`, when `words` is length 2 and `and = ""`, `sep` will now be used (thanks, @eitsupi, #2044).
 
 # CHANGES IN knitr VERSION 1.34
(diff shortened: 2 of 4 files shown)


R/defaults.R
```diff
@@ -67,10 +67,12 @@ new_defaults = function(value = list()) {
...
 #' @examples opts_chunk$get('prompt'); opts_chunk$get('fig.keep')
 opts_chunk = new_defaults(list(
-
   eval = TRUE, echo = TRUE, results = 'markup', tidy = FALSE, tidy.opts = NULL,
   collapse = FALSE, prompt = FALSE, comment = '##', highlight = TRUE,
-  strip.white = TRUE, size = 'normalsize', background = '#F7F7F7',
+  size = 'normalsize', background = '#F7F7F7',
+
+  # NA value means will set the default option later (e.g in fix_options)
+  strip.white = NA,
```
cderv, line 75: Is this safe to move the order of the option here ? 

I am not sure but I find it clearer to isolate the option we set using NA. 

but the minimal change would work the same

````diff
opts_chunk = new_defaults(list(
-
+ # NA value means will set the default option later (e.g in fix_options)
  eval = TRUE, echo = TRUE, results = 'markup', tidy = FALSE, tidy.opts = NULL,
  collapse = FALSE, prompt = FALSE, comment = '##', highlight = TRUE,
- strip.white = TRUE, size = 'normalsize', background = '#F7F7F7',
+ strip.white = NA, size = 'normalsize', background = '#F7F7F7',
  size = 'normalsize', background = '#F7F7F7',
````

**Comment by cderv** [comment]
Ok it seems `fix_options()` could not be the right place to make this change...

What is happening: 

* `engine_output()` will call `sew()`
* `sew.source()` will check with a `if()`, `options$strip.white` which will be `NA` and not `TRUE/FALSE`

````r
> # expert use only
> engine_output(opts_chunk$merge(list(engine = "python")),
+     out = list(structure(list(src = "1 + 1"), class = "source"),
+         "2"))
Error in if (options$strip.white) src = strip_white(src) : 
  missing value where TRUE/FALSE needed
Calls: engine_output ... unlist -> sew -> sew.list -> lapply -> FUN -> sew.source
Execution halted
````

So we need to use another way of setting default option conditionally, or maybe `engine_output()` or `sew.source()` need a fix 🤔 
It happens with the example 
```r
engine_output(opts_chunk$merge(list(engine = "python")), out = list(structure(list(src = "1 + 1"), class = "source"), "2"))
```
which is quite specific, and indeed in this case, `opts_chunk` will not be processed by `fix_options()`. 

Another idea, (bc I like this trick too): Could we use set the default to 
```r
strip.white = I(TRUE)
```

So that it is a valid logical value, but we can check if it has been modified ? 

``` r
opts = knitr::opts_chunk$get(default = TRUE)
opts$strip.white = I(TRUE)
str(opts$strip.white)
#>  'AsIs' logi TRUE
if (inherits(opts$strip.white, "AsIs")) opts$strip.white = !opts$collapse
str(opts$strip.white)
#>  logi TRUE
```

What do you think ?

cderv:
Regarding **knitr-examples** checks, following #2049 we get now the correct PR branch checked. 

I think it errors because of some end line difference maybe... I may need to tweak the git config so that WINDOWS vs LINUX / MAC does not bite me. 

This is one of the main difficulty with this snapshot tests. I can look into that, but it could be as easy if you knit the Rmd to md on your computer. I tried from Linux, but oddly I get the same file - no git diff...

yihui (approved):
I'm okay with the `I(TRUE)` trick. Alternatively we can also run `fix_options()` inside `engine_output()` as you mentioned. Either way is fine. The `I()` trick might be easier and safer.

> Error is here: https://github.com/yihui/knitr/runs/3674545808#step:11:1615

That is not an EOL problem, but there are two spaces after `1 + 1`, which are removed because of `tidy = TRUE`: https://github.com/yihui/knitr-examples/blob/master/k The source files are knitted using this script, e.g.,

```bash
./k 122-strip-white.Rmd
```

cderv:
I forgot about this script `./k`

I don't think I can use it from Windows directly. I need to switch to Linux and install the package before rendering. I did `devtools::load_all()` then `knit()` 😅 

I'll adapt the workflow.

cderv:
I'll recheck this process on my Windows machine. Possibly having an R function for this is better and a very good idea. This would allow to document also as an internal help page. I like that.

PR #2053, opened 2021-09-29
Extend class= regex in rocco to allow hyphens

MichaelChirico:
As seen e.g. here:

https://github.com/cran/startup/blob/6c67c07bab3e464f2f7f10e51fb12d5d93972c25/inst/doc/startup-intro.html#L95

I'm not sure exactly what causes `"language-r"` to be generated (which makes adding a test hard, too), but I do see it in a few places.

It seems like a pretty straightforward fix, but I am treading lightly in case there are other risks I may not be aware of.


yihui (approved):
Sounds good to me. Thanks!

Issue #2040, opened 2021-08-31
Allow to write verbatim code chunks using ```{{ }} 


cderv:
This would follow first feature in **rmarkdown** (https://github.com/rstudio/rmarkdown/issues/2206) which would allow to write this: 

`````markdown
````{.markdown}
```{{r setup, echo=TRUE}}
library(tidyverse)
```
````
`````
or 

`````markdown
````markdown
```{{r setup, echo=TRUE}}
library(tidyverse)
```
````
`````

Per @yihui's idea, we could also do it in **knitr** directly to simplify by writing only this 
`````markdown
```{{r setup, echo=TRUE}}
library(tidyverse)
```
`````

* **knitr** would parse the `{{ }}` 
* replace to `{ }`
* Put the content between the fenced backtick with the language class for Pandoc to process. 

This is only a simplification to write less: 
* Creating a verbatim code chunk would be only to add `{{ }}` instead of `{ }`

cc @apreshill

Related to https://github.com/yihui/knitr/issues/1688 and others linked

**Comment by apreshill** [comment]
Updated syntax is a bit different:
https://quarto.org/docs/computations/running-code.html#fenced-echo

````
```{python}
#| echo: fenced
#| output: false
#| code-overflow: wrap
1 + 1
```
````

**Comment by cderv** [comment]
Oh good to know. I'll have a look at the way it is done in quarto. 

Maybe @yihui already know as I think he is looking through the code for the new syntax for chunk option.

**Comment by apreshill** [comment]
Note that the syntax with double curly brackets is also possible as a non-executable block:
https://quarto.org/docs/computations/running-code.html#fenced-echo

````
```{{python}}
1 + 1
```
````

**Comment by cderv** [comment]
I understand the `{{ }}` syntax is equivalent to 

```{r}
#| echo: fenced
#| eval: FALSE
1 + 1
```

So maybe we just need to support with **knitr** `echo = "fenced"` 🤔

cderv:
We have now introduced a verbatim engine for this purpose

`````
````{verbatim}
```{r}
1 + 1
```
````
`````

will output the content of the verbatim chunk in a fenced code block 

`````
````default
```{r}
1 + 1
```
````
`````

This serve the purpose of this feature request

PR #2060, opened 2021-10-21
Add a `verbatim` engine to output verbatim content

yihui:
This is one of possible implementations of #2040.


yihui (commented):
@cderv I'm only documenting the `lang` option in the NEWS for the sake of simplicity. I guess that should be enough for most users.

If you don't have other thoughts, I think this PR can be merged now. Thanks!

cderv:
I think it is good! 

We can document further in the R Markdown cookbook or another place once new version is on CRAN. 

Tests needs to added for this one. I'll that this week.

cderv (approved):

Issue #1411, opened 2017-06-23
include_graphics fails with '~' home directory shorthand


RoyalTS:
`include_graphics` apparently fails on Mac when I specify a path using the `~` shorthand for the user home directory.

    ```{r}
    knitr::include_graphics('~/myProject/out/figures/2017-06-23/sample_size_all.pdf')
    ```

gives

    [pandoc warning] Could not find image `~/myProject/out/figures/2017-06-23/sample_size_all.pdf', skipping...

upon knitting. The following works, in contrast:

    ```{r}
    knitr::include_graphics('/Users/myUsername/myProject/out/figures/2017-06-23/sample_size_all.pdf')
    ```

yihui:
Absolute paths like this is almost always a very bad idea. Please use relative paths instead for the sake of portability.

RoyalTS:
I am well aware that it's usually a good idea to use paths relative to the location of the document. In this case, however, precisely for reasons of portability I've specified all paths relative to project root using the [`here`](https://cran.r-project.org/web/packages/here/index.html) package, which spits out absolute paths. The above is just a demonstration that _some_ absolute paths don't seem to work for `include_graphics`.

yihui:
I guess it is not the responsibility of `knitr::include_graphics()` to process the path. This function takes whatever paths you pass to it. It is Pandoc that failed to recognize `~`. If this issue has to be fixed, I'd recommend the **here** package to do either of the following things:

- `normalizePath()` or `path.expand()`;

- or return a relative path (at least provide an option to return a relative path).

I have my reserved opinions about **here**. I see its point, but I don't think this nasty problem of file paths really has a perfect solution.

jzadra:
THe odd thing is that paths with '~' work for ioslides, but not html/doc documents, and both use pandoc.  Couldn't include_graphics() just expand the path itself to avoid this error?

yihui:
It could. It is just that I prefer fixing the disease instead of the symptoms. In this case, it is not the fault of `include_graphics()`. Either the user or the **here** package could do `path.expand()`. If I do it in `include_graphics()`, I have no idea whether it is going to have other undesired effects.

ioslides is different with all other output formats. It is based on a [custom Lua writer](http://pandoc.org/MANUAL.html#custom-writers). That is probably why Pandoc wouldn't complain about the path.

jzadra:
Gotcha, thanks for the response and for the awesome package!

PR #2063, opened 2021-10-23
Give a warning if include_graphics is given a pathname starting a tilde

kcarnold:
The error message from Pandoc in this situation has confused several of my students.

I think `knitr` should just run `path.expand` itself, but the discussion at https://github.com/yihui/knitr/issues/1411 seems to indicate otherwise.

At the very least, it seems odd that there's one kind of error if the file doesn't exist (the `stop` in `include_graphics` and a different kind of error if the file does exist but its name is given in the "wrong" (but Unix-standard) format. `file.exists` unconditionally runs `path.expand`; perhaps `include_graphics` should too.


yihui (approved):
> Is a warning message too far?

I'm not sure. I'll merge it for now, but may revert the warning if it breaks anything or anyone complains about it. Thanks!

PR #2070, opened 2021-11-09
Possible issue with is_cran()

cderv:
I think this could be an issue here. 

We want the parenthesis here right ? 

Saw that in some log 
````
Warning in any(tolower(Sys.getenv(c("CI", "NOT_CRAN")))) :
    coercing argument of type 'character' to logical
````


yihui (approved):
That's right. Thanks!

PR #2076, opened 2021-11-30
Provide an `embed` engine to embed external file verbatim

### LINKED ISSUE #1275: include_file
For writing about Rmarkdown, it would be useful to have `knitr::include_file()` which would just insert the contents of the file into the text. It would also be useful if you could set the syntax highlighting.

### LINKED ISSUE #1931: Offer a way to insert file content as a verbatim code with highlighting
This is a recurring topic covering several use case

* inserting verbatim code chunk in a Rmd file
* inserting 'asis' code output generated from R which language to highlight

We've documented this in the cookbook for verbatim code chunk
https://bookdown.org/yihui/rmarkdown-cookbook/verbatim-code-chunks.html
and in a way with asis output
https://bookdown.org/yihui/rmarkdown-cookbook/results-asis.html

and used a [trick in the source code](https://github.com/yihui/rmarkdown-cookbook/blob/9931f7644fb8a7becde55925da4aaf224321f392/index.Rmd#L59-L67) for insterting Rmd content asis
```r
import_example <- function(file, lang = xfun::file_ext(file)) {
  x = xfun::read_utf8(file.path("examples", file))
  lang = tolower(lang)
  if (nchar(lang) > 1) {
    lang = sub('^r', '', lang)
    if (lang == 'nw') lang = 'tex'
  }
  knitr::asis_output(paste(c(sprintf("````%s", lang), x, "````"), collapse = '\n'))
}
```
that we used mainly with [inline chunk](https://github.com/yihui/rmarkdown-cookbook/blob/d26e5fb65376d7f58bedb0e42370b995bff38d1a/12-output-hooks.Rmd#L99)
```r
`r import_example('hook-secret.Rmd')`
```

This kind of function are often used. Other example in [Plumber documentation](https://github.com/rstudio/apis-plumber/blob/8474a1019aca578b3d5ccbde9c1961505f68cab3/index.Rmd#L20-L29)
```r
code_chunk <- function(output, language=""){
  cat(paste0("```",language,"\n"))
  cat(output)
  cat("\n```\n")
}

# Include an external R file with syntax highlighting in the doc
include_r <- function(file){
  code_chunk(paste0(readLines(file), collapse="\n"), language="r")
}
```
Used in several ways, like [inserted JSON output](https://github.com/rstudio/apis-plumber/blob/ded991c95628ccab9d26e9570c54689e6e25f4f3/04-rendering-output.Rmd#L67-L71) 
````r
```{r, echo=FALSE, results='asis'}
pr <- plumber::plumb("files/apis/04-03-letters.R")
e <- pr$endpoints[[1]][[1]]
code_chunk(json_serialize(e$exec(letter="U")), "json")
```
````
or [inserting R code](https://github.com/rstudio/apis-plumber/blob/ded991c95628ccab9d26e9570c54689e6e25f4f3/04-rendering-output.Rmd#L33-L35)
````r
```{r, echo=FALSE, results='asis'}
include_r("files/apis/04-01-response.R")
```
````

I believe having a way to do such things (maybe by just having one of these functions above) would be very interesting. Could be sum up by having a way to "include as-is as language-XXX”. (cc @aronatkins)

Just putting this here so that we can discuss.

### PR DESCRIPTION
**Pull request by yihui**
close #1931, and also close #1275

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -76,7 +76,7 @@
   
   This engine also works for other types of documents (e.g., `Rnw`) but it will not allow for nested code chunks within the `verbatim` engine.
 
-- Added a new engine named `embed` to embed external plain-text files. It is essentially a simple wrapper based on the `verbatim` engine, with the chunk content read from an external file. That is,
+- Added a new engine named `embed` to embed external plain-text files. It is essentially a simple wrapper based on the `verbatim` engine, with the chunk content read from an external file and default language guessed from file extension. That is,
 
   ````
   ```{embed, file="foo.R"}
@@ -86,11 +86,11 @@
   is equivalent to
   
   ````
-  ```{verbatim, code=xfun::read_utf8("foo.R")}
+  ```{verbatim, file = "foo.R", lang = "r"}
   ```
   ````
   
-  If you provide the chunk option `file` to the `embed` engine, it will read the file and show its content verbatim in the output document. Alternatively, you can specify the file path in the chunk body, e.g.,
+ If you provide the chunk option `file` to the `embed` engine, it will read the file and show its content verbatim in the output document. Alternatively, you can specify the file path in the chunk body, e.g.,
   
   ````
   ```{embed}
@@ -100,7 +100,7 @@
   
   The quotes are optional but can be helpful for editors (e.g., RStudio IDE) to autocomplete the file paths.
   
-  The syntax highlighting language name is from the filename extension by default, and you can override it with the chunk option `lang` (e.g., `file = "foo.sh", lang = "bash"`).
+  The syntax highlighting language name is from the filename extension by default, and you can override it with the chunk option `lang` (e.g., `file = "foo.sh", lang = "bash"`) which is then identical to the `verbatim` engine.
 
 ## BUG FIXES
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -74,6 +74,32 @@
   
   This engine also works for other types of documents (e.g., `Rnw`) but it will not allow for nested code chunks within the `verbatim` engine.
 
+- Added a new engine named `embed` to embed external plain-text files. It is essentially a simple wrapper based on the `verbatim` engine, with the chunk content read from an external file. That is,
+
+  ````
+  ```{embed, file="foo.R"}
+  ```
+  ````
+  
+  is equivalent to
+  
+  ````
+  ```{verbatim, code=xfun::read_utf8("foo.R")}
+  ```
+  ````
+  
+  If you provide the chunk option `file` to the `embed` engine, it will read the file and show its content verbatim in the output document. Alternatively, you can specify the file path in the chunk body, e.g.,
+  
+  ````
+  ```{embed}
+  "foo.txt"
+  ```
+  ````
+  
+  The quotes are optional but can be helpful for editors (e.g., RStudio IDE) to autocomplete the file paths.
+  
+  The syntax highlighting language name is from the filename extension by default, and you can override it with the chunk option `lang` (e.g., `file = "foo.sh", lang = "bash"`).
+
 ## BUG FIXES
 
 - The chunk option `child` also respects the package option `root.dir` now (thanks, @salim-b, https://community.rstudio.com/t/117563).
--- a/R/engine.R
+++ b/R/engine.R
@@ -792,6 +792,23 @@ eng_verbatim = function(options) {
   eng_cat(options)
 }
 
+# embed a file verbatim
+eng_embed = function(options) {
+  # if `file` is empty, use `code` as the list of files
+  f = options$file %n% options$code
+  f = gsub('^["\']|["\']$', '', f)  # in case paths are quoted
+  if (length(f) == 0) return()
+  # TODO: use xfun::read_all() so we can read multiple files at once
+  options$code = xfun::read_utf8(f)
+  # use the filename extension as the default language name
+  if (nchar(lang <- file_ext(f[1])) > 1) {
+    lang = sub('^R', '', lang)  # Rmd -> md, Rhtml -> html, etc.
+    if (lang == 'nw') lang = 'tex'
+  }
+  options$class.source = options$class.source %n% tolower(lang)
+  eng_verbatim(options)
+}
+
 # set engines for interpreted languages
 local({
   for (i in c(
@@ -814,6 +831,7 @@ knit_engines$set(
   comment = eng_comment,
   css = eng_css,
   dot = eng_dot,
+  embed = eng_embed,
   fortran = eng_shlib,
   fortran95 = eng_shlib,
   go = eng_go,
--- a/R/engine.R
+++ b/R/engine.R
@@ -783,29 +783,33 @@ eng_comment = function(options) {}
 eng_verbatim = function(options) {
   # change default for the cat engine
   options$eval = FALSE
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: R/engine.R [hunk 1]
```diff
@@ -792,6 +792,23 @@ eng_verbatim = function(options) {
...
+  f = gsub('^["\']|["\']$', '', f)  # in case paths are quoted
+  if (length(f) == 0) return()
+  # TODO: use xfun::read_all() so we can read multiple files at once
+  options$code = xfun::read_utf8(f)
+  # use the filename extension as the default language name
+  if (nchar(lang <- file_ext(f[1])) > 1) {
+    lang = sub('^R', '', lang)  # Rmd -> md, Rhtml -> html, etc.
+    if (lang == 'nw') lang = 'tex'
+  }
+  options$class.source = options$class.source %n% tolower(lang)
```
cderv, line 808: I was asking myself the question so putting this here for reference. Maybe we could add a comment to not forget

`eng_embed` uses `eng_verbatim` , order of options: 
1. `options$engine.opts$lang`
2. `options$lang`
3. `options$language`
4. `options$class.source`

So I would use one of the first because `class.source` can be used to add a class in addition to language. and language will be removed later in `eng_cat`. Currently it can cause issues likes this : 

````markdown
```{cat, engine.opts = list(file = "style.css"), echo = FALSE}
h1 {
  color: red;
}
```
# Specifying language

```{embed, lang = "md"}
"style.css"
```
````

will return 
````markdown
# Specifying language


```{.md .css}
h1 {
  color: red;
}
```
````

because we pass the file extension as a `class.source` without setting the other attributes used in `eng_verbatim`. 
What we want is to just change the default value for `eng_verbatim` (which is `default` string)

We could 

* Add a helper to check attributes and set default otherwise 
  ```
  set_lang_attr = function(options, default) {
    options$engine.opts$lang = options$engine.opts$lang %n%
        unlist(options[c('lang', 'language')])[1] %n%
        options$class.source %n% default
    options
  }
  ```
  to use in both `eng_embed` (`options = set_lang_attr(options, lang)` and `eng_verbatim` (`options = set_lang_attr(options, "default")`

* Add an internal `options$default.lang` in `eng_verbatim` than we can use in `eng_embed`
  ````diff
  diff --git a/R/engine.R b/R/engine.R
  index ce2e58af..110927cd 100644
  --- a/R/engine.R
  +++ b/R/engine.R
  @@ -787,7 +787,7 @@ eng_verbatim = function(options) {
    # or class.source; if all are empty, use 'default'
    options$engine.opts$lang = options$engine.opts$lang %n%
      unlist(options[c('lang', 'language')])[1] %n%
  -    options$class.source %n% 'default'
  +    options$class.source %n% options$default.lang %n% 'default'
  
    eng_cat(options)
  }
  @@ -805,7 +805,7 @@ eng_embed = function(options) {
    lang = sub('^R', '', lang)  # Rmd -> md, Rhtml -> html, etc.
    if (lang == 'nw') lang = 'tex'
  }
  -  options$class.source = options$class.source %n% tolower(lang)
  +  options$default.lang = tolower(lang)
  eng_verbatim(options)
  }

  ````

**Review by cderv — CHANGES_REQUESTED** [review]
This looks ok. Only one drawback while testing in the way we are setting the default language. 
I let you have a look

**Comment by cderv** [comment]
Also something came to mind. 

What about : 
````mardkown
```{verbatim, file = "style.css"}
```
````

This would avoid a specific engine, but we would loose the hability to provide file as content of the code chunk - only through option. Basically, you can provide `file` to `verbatim` engine, and in that case, chunk content is ignored and replaced by the file content. Default language is set to file extension. 

Have you thought of that and already decided on a new `embed` engine instead ?

File: R/engine.R (hunk 1 above)
Comment (yihui) [reply R/engine.R]: I applied the helper function you provided. Thanks!

**Comment by yihui** [comment]
> ```` ```{verbatim, file = "style.css"}````

This is supported now with https://github.com/yihui/knitr/commit/545aed32274955be2109150ce1ef4e24fe3732de. Note that the `file` option is not specific to the `verbatim` engine but any engine. This is a feature that I have wanted to add for long.

Now the only benefit of the `embed` engine is that you can specify the file path inside the code chunk, which makes the syntax a little terser.

File: NEWS.md [hunk 2]
```diff
@@ -76,6 +76,32 @@
...
+  ````
+  ```{embed}
+  "foo.txt"
+  ```
+  ````
+  
+  The quotes are optional but can be helpful for editors (e.g., RStudio IDE) to autocomplete the file paths.
+  
+  The syntax highlighting language name is from the filename extension by default, and you can override it with the chunk option `lang` (e.g., `file = "foo.sh", lang = "bash"`).
+
```
cderv, line 104: We should change with this new file option, shouldn't we ? 

`````suggestion
- Added a new engine named `embed` to embed external plain-text files. It is essentially a simple wrapper based on the `verbatim` engine, with the chunk content read from an external file and default language guessed from file extension. That is,

  ````
  ```{embed, file="foo.R"}
  ```
  ````
  
  is equivalent to
  
  ````
  ```{verbatim, file = "foo.R", lang = "r"}
  ```
  ````
  
 If you provide the chunk option `file` to the `embed` engine, it will read the file and show its content verbatim in the output document. Alternatively, you can specify the file path in the chunk body, e.g.,
  
  ````
  ```{embed}
  "foo.txt"
  ```
  ````
  
  The quotes are optional but can be helpful for editors (e.g., RStudio IDE) to autocomplete the file paths.
  
  The syntax highlighting language name is from the filename extension by default, and you can override it with the chunk option `lang` (e.g., `file = "foo.sh", lang = "bash"`) which is then identical to the `verbatim` engine.

`````

File: R/engine.R [hunk 3]
```diff
@@ -783,13 +783,34 @@ eng_comment = function(options) {}
...
-    options$class.source %n% 'default'
+    options$class.source %n% default
+  options
+}
 
-  eng_cat(options)
+# embed a file verbatim
+eng_embed = function(options) {
+  # if `file` is empty, use `code` as the list of files
+  if (is.null(options$file)) {
```
cderv, line 802: I think we need to assign the file option to `f` and just modify it if it is NULL
```suggestion
  if (is.null(f <- options$file)) {
```

Otherwise, this will error as f not define 
```
if (nchar(lang <- file_ext(f[1])) > 1) {
```

cderv (changes requested):
> This is supported now with [545aed3](https://github.com/yihui/knitr/commit/545aed32274955be2109150ce1ef4e24fe3732de). Note that the `file` option is not specific to the `verbatim` engine but any engine. This is a feature that I have wanted to add for long.

I was getting used to using `code = xfun::read_utf8("test.R")` 😄 But this will be easier for sure. That is a great addition ! 

> Now the only benefit of the `embed` engine is that you can specify the file path inside the code chunk, which makes the syntax a little terser.

The other benefit is that you have the the language set by default to the file extension. The `verbatim` engine won't do that right ? 

* `verbatim`
````
```{verbatim, file = "foo.R", lang = "r"}
```
````

* `embed`
````
```{embed, file="foo.R"}
```
````

That is a bit shorter. 

There is one small issues found by testing (I'll add to knitr-example afterwards), Otherwise looks good! 

Thanks!

yihui (commented):
Good suggestions. Thanks!

cderv:
I have added example for this in **knitr-example**

yihui:
Great. Thanks!

PR #2078, opened 2021-12-02
Add `knit_cnd_format()` generic

lionel-:
rlang is now in charge of fully printing error messages, including the error prefix. This makes it possible to support more features, such as custom display of condition calls and chained errors.

```r
mtcars |> dplyr::summarise(cyl = 1 + "")
#> Error in `dplyr::summarise()`:
#>   Problem while computing `cyl = 1 + ""`.
#> Caused by error in `+`:
#>   non-numeric argument to binary operator
```

Ideally, the condition messages in knitted documents should match the messages in the REPL. This is currently achieved for errors via an `as.character.rlang_error()` method that is called by `sew.error()`. However, methods for other conditions do not call `as.character()`. Also that customisation point is not very explicit and is probably to be considered off-label usage from the point of view of knitr.

To solve these issues, this PR proposes to add a new generic for the full formatting of condition messages, including the prefix. Example usage can be found in https://github.com/r-lib/rlang/pull/1323/files.

cc @hadley


R/output.R
```diff
@@ -529,23 +529,60 @@ msg_sanitize = function(message, type) {
...
+  msg_wrap(knit_cnd_format(x), 'warning', options)
+}
+is_eval_call = function(x) {
+  is.call(x) && identical(x[1], quote(eval()))
 }
 
 #' @export
 sew.message = function(x, options, ...) {
-  msg_wrap(paste(conditionMessage(x), collapse = ''), 'message', options)
+  msg_wrap(knit_cnd_format(x), 'message', options)
```
cderv, line 543: I think the `paste()` here was required so that this works

> A message in `message(..., appendLF = FALSE)` will be merged with the next adjacent message:
>
> ```{r}
>f3 = function() {
>  message('Hello ', appendLF = FALSE)
>  message('World!')
>}
>f3()
>```

The change here currently make the test fails on this. The knitr-examples checks is showing the difference between previous and new behavior for example (https://github.com/yihui/knitr-examples/blob/master/117-messages.Rmd)
https://github.com/yihui/knitr/runs/4396175038?check_suite_focus=true#step:11:1669

Was it necessary to remove it in `knit_cnd_format.message` ?

cderv (commented):
Just a small feedback as I observe a change in the test. 

Otherwise, this looks good. I let @yihui add his review which will be more relevant for **knitr** internals. Thanks Lionel !

R/output.R
lionel-: Thanks for the pointer Christophe!

Since it's not "in-API" to return a character vector from `conditionMessage()`, I overlooked that part. Now fixed, with unit test and comments. However this means that `knit_cnd_format()` methods for messages need to know about that message merging behaviour and the `message` field potentially being a character vector. That's a bit problematic.

By the way, `merge_class()` manipulates `message` fields instead of the result of `conditionMessage()`. This could cause issues in the future because conditions might have empty `""` or dummy strings as message fields, with the actual message generated with a `conditionMessage()` method. It's not a big deal though because lazy generation of error messages mostly occurs with errors not messages. However it's worth knowing that logic isn't 100% correct.

I guess ideally `knit_cnd_format()` would be called inside the `merge_class()` routine so that it generates messages properly. A new object holding the merged messages would need to be created instead of modifying the existing condition, which is not hygienic because it may break class expectations. And then we don't need the collapsing in `knit_cnd_format.message()` or any methods for subclasses of messages. 

That said, we don't need to fix this now (or maybe ever) since message subclasses are not that important.

lionel-: By the way, is there a way to run the knitr-examples check locally?

cderv: To run them locally, you need to clone the repository and then knit the files. 
Usually, we automate the testing using CI and only run locally when something seems off. 

Otherwise, to reproduce the checks without trying to recreate the full environnemnt (not that easy, that is why I rely on CI as I struggled bc on Windows mainly), the script to run in `knitr-examples/` is `./knitall` after having installed dev version. 
The full process is `make integration` in knitr folder, which should work if you have knitr-example at the right place inside the knitr folder. (what happens in the GHA workflow). 

To run one file, the script is `./k`

Anyway, relying on the CI to run those tests is fine also.

yihui:
Thanks for this, @lionel-, and sorry for the long delay.

The motivation here was to give rlang an explicit extension point so knitted condition messages could match the REPL (call prefixes, chained errors, etc.), with the consumer side in r-lib/rlang#1323. That rlang PR was closed unmerged in Feb 2023 with your own note, "Not sure this is still needed," and rlang currently has no reference to `knit_cnd_format()`. So merging this would export a new generic (plus an `rlang` Suggests dependency) with nothing dispatching on it.

The original problem also appears to be handled already on the current codebase. With rlang 1.3.0, an rlang error renders through `sew.error()` → `as.character(x)` as its full REPL-style message:

```
Error in `foo()`:
! boom
```

so the `as.character()` path you described as "off-label" is in fact the working mechanism today, and errors match the REPL without this generic. Warnings and messages don't get rlang's richer formatting, but that's precisely what the closed rlang PR would have supplied.

Given there's no consumer and the error case already works, I'll close this to keep the exported API surface minimal. If rlang revisits this and wants the extension point again, I'm happy to reopen — the refactor itself is clean. Thanks again.

PR #2079, opened 2021-12-02
Pass user-supplied calling handlers to `evaluate()`

lionel-:
Depends on r-lib/evaluate#105.

New `calling.handlers` option to set calling handlers within `evaluate()`. The goal is to `entrace()` errors, as explained in the evaluate PR.


yihui (approved):
Thanks! We can merge this after the next version of **evaluate** is released to CRAN.

PR #2083, opened 2021-12-10
export partition_chunk

zkamvar:
I would like to export the `partition_chunk()` function so that it can be used in {tinkr} to add the chunk options to the XML representation (see: https://github.com/ropensci/tinkr/issues/55). 

I've given some internal documentation for it as well.


yihui (approved):
Thanks!

PR #2092, opened 2022-01-05
Engine for ditaa text diagrams

kondziu:
Ditaa (https://github.com/stathissideris/ditaa) is a small command-line utility that converts ASCII art diagrams into bitmaps. It can be useful for making rough-and-ready images accompanying technical text. The pull request adds an engine to knitr to perform the conversion in place while knitting. 

Example ditaa diagram in ASCII form:

```
    +--------+--------+--------+--------+--------+--------+--=--+--------+--------+--------+--------+
    | cCCC   | cCCC   | c99C   | c99C   | cCCC   | cCCC   |     | cCCC   | cCCC   | c99C   | c99C   |
----+ R a[0] | R b[0] | W a[0] | W b[0] | R a[1] | R b[1] | ... | R a[9] | R b[9] | W a[9] | W b[9] +----
    |        |        |        |        |        |        |     |        |        |        |        |
    +--------+--------+--------+--------+--------+--------+--=--+--------+--------+--------+--------+

                                        +--------+--------+--------+--------+--------+--------+--=--+--------+--------+--------+--------+
                                        | cCCC   | cCCC   | c99C   | c99C   | cCCC   | cCCC   |     | cCCC   | cCCC   | c99C   | c99C   |
----------------------------------------+ R b[0] | R c[0] | W b[0] | W c[0] | R b[1] | R c[1] | ... | R b[9] | R c[9] | W b[9] | W c[9] +----
                                        |        |        |        |        |        |        |     |        |        |        |        |
                                        +--------+--------+--------+--------+--------+--------+--=--+--------+--------+--------+--------+
```

And an image generated from it:

lets_have_some_fun3

The extension is based on the Dot/Asymptote engine, extended with command line options for configuring the conversion.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
-Version: 1.37.5
+Version: 1.37.6
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("Abhraneel", "Sarma", role = "ctb"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,8 @@
   
   We'd like to thank @TianyiShi2001 for the inspiration (#1829 #1823 #1833).
 
+- Added a new engine `ditaa` based on the `exec` engine to convert ASCII art diagrams to bitmaps (thanks, @kondziu, #2092).
+
 - Added two new chunk options, `class.chunk` and `attr.chunk`, for R Markdown output. These options can enclose the whole chunk output (source and results) in a fenced Div. Their syntax follows other chunk options with the similar names `class.*` and `attr.*` (e.g., `class.source` and `attr.source`). For example, `class.chunk = "foo"` would wrap the chunk output inside `::: {.foo}` and `:::` (thanks, @atusy, #2099).
 
 - Added a new chunk option `lang` to set the language name of a code chunk. By default, the language name is the engine name. This is primarily useful for syntax highlighting the source chunks in Markdown-based output. Previously the `lang` option was only available to a few engines such as `verbatim` and `embed`. Now it is available to all engines.
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,7 +36,7 @@
   
   We'd like to thank @TianyiShi2001 for the inspiration (#1829 #1823 #1833).
 
-- Added a new engine `ditaa` based on the `exec` engine to convert ASCII art diagrams to bitmaps (thanks, @kondziu, #2092).
+- Added a new engine `ditaa` based on the `exec` engine to convert ASCII art diagrams to bitmaps via [the `ditaa` command](https://github.com/stathissideris/ditaa) (thanks, @kondziu, #2092).
 
 - Added two new chunk options, `class.chunk` and `attr.chunk`, for R Markdown output. These options can enclose the whole chunk output (source and results) in a fenced Div. Their syntax follows other chunk options with the similar names `class.*` and `attr.*` (e.g., `class.source` and `attr.source`). For example, `class.chunk = "foo"` would wrap the chunk output inside `::: {.foo}` and `:::` (thanks, @atusy, #2099).
 
--- a/R/engine.R
+++ b/R/engine.R
@@ -365,6 +365,52 @@ eng_dot = function(options) {
   engine_output(options, code, '', extra)
 }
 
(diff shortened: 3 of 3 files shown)


yihui (approved):
Thanks for the contribution! I just simplified your implementation at the cost of not providing chunk options to represent `ditaa` command-line arguments. If users need to use any `ditaa` arguments, they can use the chunk option `engine.opts$args1`, e.g., 

````
```{ditaa, engine.opts=list(args1 = c('-r', '-s', '2'))}
    +--------+   +-------+    +-------+
    |        | --+ ditaa +--> |       |
    |  Text  |   +-------+    |diagram|
    |Document|   |!magic!|    |       |
    |     {d}|   |       |    |       |
    +---+----+   +-------+    +-------+
        :                         ^
        |       Lots of work      |
        +-------------------------+
```
````

Are you okay with that?

**Comment by kondziu** [comment]
I think `engine.opts=list(args1 = c('-r', '-s', '2'))` is relatively difficult to type out and subsequently read, but I guess it's a reasonable simplification and you're the boss, boss.

**Comment by yihui** [comment]
I agree with you (I said it was a "cost"). To make the arguments a little easier to read, we can use the full names like `--round-corners --scale 2`. The main problem with using chunk options to represent command-line arguments is that we will need to document all of them, and keep the documentation up-to-date in future. That would be a maintenance burden, although I guess it shouldn't be too much burden.

Alternatively, I can provide a general mechanism to turn an R list into command-line arguments, e.g., `list(round_corners = TRUE, scale = 2)` -> `--round-corners --scale 2`. Then I don't need to document all options for every single command-based engine. Would you be happy with that?

**Comment by kondziu** [comment]
I think from the usage point of view `engine.opts=list(args1 = c('-r', '-s', '2'))` is relatively difficult to type out but it's better than nothing, and you're the boss, boss.

**Comment by kondziu** [comment]
Sorry for leaving you hanging for so long. Thank you for considering my concern. I think providing a list that converts into command line arguments is fine.

However, the point I am really trying to make is that I would be careful of boilerplate. ditaa's defaults are not necessarily the best, so a typical user typically would need to provide at least some of those commandline arguments for every block. It'll probably get annoying fast if they have to write this obscure magical formula for every block. In other words, it's not the `c('-r', '-s', '2')` part that worried me, it's the long and opaque `engine.opts=list(args1 = ...)` that I think might end up being unwieldy in practice. 

So I'd advocate for a solution that removes as much of that boilerplate as practical. A solution that lets users pass arguments directly to ditaa without interpreting them, such as the one you suggested, is fine. I agree that it is a more maintainable way of doing things, and easier to document, and that this is good and necessary. But I think it's important to make passing in these arguments simpler. E.g. instead of `engine.opts=list(args1 = c(...))` just `ditaa_args=...` or at least `engine.opts(ditaa_args=...)`.

Having said all that, if it ends up taking the form of `engine.opts=list(args1 = c('-r', '-s', '2'))`,  that's still better than nothing and the users will just get used to it.

**Comment by cderv** [comment]
> It'll probably get annoying fast if they have to write this obscure magical formula for every block

Usually, `engine.opts` can be defined globally per engine. Something like this  This should work when set in a setup chunk. 

````r
knitr::opts_chunk$set(engine.opts = list(ditaa= list(args1 = c('-r', '-s', '2'))))
````

This would apply this default `args1` to all chunks using the `ditaa` engine.

Would that help ? 

I don't think it is working here yet, but it could IMO.

**Comment by kondziu** [comment]
That's a good suggestion. I think it addresses most of my misgivings.

**Review by yihui — APPROVED** [review]
@kondziu I have applied the default arguments from your commit 2cc85d2d8e8eaae9079ffae103e95919136d62c1 to a861e53e727ae204331fa1bc3f442c0944218575. Hopefully you won't need to tweak them much. If you do want different defaults, the method @cderv mentioned won't work, but here is one way that works, using option hooks to change chunk options (https://yihui.org/knitr/hooks/#option-hooks):

````
```{r, include=FALSE}
knitr::opts_hooks$set(engine = function(options) {
  if (options$engine == 'ditaa') {
    options$engine.opts$args1 = c('-r', '-s', '3')
  }
  options
})
```
````

Or change it only if users didn't provide the chunk option `engine.opts$args1` (i.e., when it's `NULL`), otherwise just use the `args1` that users provided:

```r
knitr::opts_hooks$set(engine = function(options) {
  if (options$engine == 'ditaa' && is.null(options$engine.opts$args1)) {
    options$engine.opts$args1 = c('-r', '-s', '3')
  }
  options
})
```

@kondziu Are you happy with this PR now?

yihui:
I'm merging this PR for now, but I'm open to making further changes if you have any suggestions. Thanks again!

PR #2096, opened 2022-01-10
Add a newline on empty string in `comment_out(newline = TRUE)`

cderv:
This fixes rstudio/rmarkdown#1849

`comment_out(newline = TRUE)` also now works with an empty string. 

This PR adds some tests for `comment_out()` behavior.

Seems like a change ok to do that will fix the empty message case reported in **rmarkdown**.  Is there anything I may have missed ? 

Also, should we add NEWS here for this internal function fix ? Or a NEWS item in **rmarkdown** as the issue was reported there ?


yihui (approved):
Looks good!

I don't have an opinion on whether this is worth a NEWS item. It's up to you. Please feel free to merge with or without a NEWS item. Thanks!

PR #2097, opened 2022-01-10
bug fix in group_indices()

chroetz:
`knitr::Sweave2knitr(text = c("<<>>=", "0", "@"))` wrongly removes the "@". This is due to a bug in `knitr:::group_indices()`. Try `knitr:::group_indices(c(F,T,F,F,F),c(F,F,F,T,F))`. This results in `1 0 1 1 1` instead of `1 2 2 2 3`. The reason for the bug is the last line of `group_indices()`:
```
mapply(fun, chunk.begin, chunk.end, lines, seq_along(lines))
```
`lines` is an argument of `group_indices()` with default value `NA`. Thus, in the default case, `fun()` is always called with the 4th argument equal to 1, which indicates that the processed line is the first line of a document. Hence, the function always returns a vector consisting only of 1s and 0s. Changing this line to 
```
mapply(fun, chunk.begin, chunk.end, lines, seq_along(chunk.begin))
```
solves the problem.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 - Chunk options defined in the `#|` style are not recognized when the code chunk is indented or quoted (thanks, @mine-cetinkaya-rundel, #2086).
 
+- Fixed a bug in `Sweave2knitr()` #2097 (thanks, @chroetz).
+
 # CHANGES IN knitr VERSION 1.37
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
That sounds correct. I'll double check and merge this fix later today. Thanks a lot!

PR #2099, opened 2022-01-18
Implement class.chunk and attr.chunk option to wrap entire chunk in Pandoc's fenced div (closes #2000)

### LINKED ISSUE #2000: Option to wrap entire chunk in classed div
To support certain types of formatting it would be very useful to wrap the entire chunk in a classed div. Currently I'm doing that with a chunk knit_hook:

```R
indent <- function(x, indent) {
  if (is.null(indent)) {
    return(x)
  }
  paste0(indent, gsub("\n", paste0("\n", indent), x))
}

knitr::knit_hooks$set(chunk = function(x, options, ...) {
  if (!options$echo){
    return(x)
  }
  x <- gsub("\033", "\u2029", x, fixed = TRUE)
  x <- paste0("<div class='sourceCode'>", x, "</div>")
  indent(x, options$indent)
})
```

But it would be very useful if this was a top-level chunk option, e.g.

```R
knitr::opts_chunk$set(class.chunk = "sourceCode")
```

---

atusy:
```` r
library(knitr)

cat(knitr::knit(quiet = TRUE, text = c(
  '```{r, class.chunk="foo"}',
  '"classed div"',
  "```",
  "",
  '```{r, class.chunk=""}',
  '"classless div"',
  "```",
  "",
  '```{r}',
  '"no div"',
  "```"
)))
#> <div class="foo">
#> 
#> ```r
#> "classed div"
#> #> [1] "classed div"
#> ```
#> 
#> </div>
#> 
#> <div>
#> 
#> ```r
#> "classless div"
#> #> [1] "classless div"
#> ```
#> 
#> </div>
#> 
#> 
#> ```r
#> "no div"
#> #> [1] "no div"
#> ```
````

<sup>Created on 2022-01-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/hooks-md.R
+++ b/R/hooks-md.R
@@ -229,13 +229,13 @@ pandoc_div = function(x, .attr = NULL, .class = NULL) {
   div = paste(rep(':', n), collapse = '')
   paste0(
     paste(div, block_attr(.attr, .class), sep = " "),
-    '\n\n',
+    '\n',
     x,
-    '\n\n',
+    '\n',
     div
   )
 }
-                     
+
 # convert some engine names to language names
 eng2lang = function(x) {
   d = c(
--- a/R/hooks-md.R
+++ b/R/hooks-md.R
@@ -205,11 +205,35 @@ hooks_markdown = function(strict = FALSE, fence_char = '`') {
         r = sprintf('\n([%s]{3,})\n+\\1((\\{[.])?%s[^\n]*)?\n', fence_char, tolower(options$engine))
         x = gsub(r, '\n', x)
       }
+      x = pandoc_div(x, options[['attr.chunk']], options[['class.chunk']])
       if (is.null(s <- options$indent)) return(x)
       line_prompt(x, prompt = s, continue = s)
     },
     output = hook.o('output'), warning = hook.o('warning'),
     error = hook.o('error'), message = hook.o('message')
+  )
+}
+
+pandoc_div = function(x, .attr = NULL, .class = NULL) {
+  if (is.null(.attr) && is.null(.class)) return(x)
+  if (!is.character(x) || length(x) != 1L) {
+    warning('x is unprocessed because x is not a character vector of length 1.')
+    return(x)
+  }
+  n = 1L + max(
+    2L,
+    vapply(
+      gregexpr('^:{3,}', strsplit(x, '\n')[[1L]]),
+      attr, NA_integer_, 'match.length'
+    )
+  )
+  div = paste(rep(':', n), collapse = '')
+  paste0(
+    paste(div, block_attr(.attr, .class), sep = " "),
+    '\n\n',
+    x,
+    '\n\n',
+    div
   )
 }
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,8 @@
   
   We'd like to thank @TianyiShi2001 for the inspiration (#1829 #1823 #1833).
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by yihui — APPROVED** [review]
Thanks! Two comments:

1. Since chunk options such as `class.source` are supported in `hooks-md.R`, would it be better to support `class.chunk` there, too? https://github.com/yihui/knitr/blob/36e6a30b4c8eb29d9d9d0e87a669b83c697234a7/R/hooks-md.R#L200-L209
2. Would it be slightly cleaner to use the fenced Div syntax `::: .chunk-class` instead of the HTML syntax? It doesn't make any difference to Pandoc (both are parsed as divs). If we use `:::`, we also need to detect if `:::` is present inside the chunk output and increase the number of colons accordingly just like we did for backticks: https://github.com/yihui/knitr/blob/36e6a30b4c8eb29d9d9d0e87a669b83c697234a7/R/hooks-md.R#L155-L163

**Comment by atusy** [comment]
Thanks for the comment!

1. Implementation in `hooks-md.R` is the good idea! I'll do that.
2. I love the `:::` syntax, but did not employ it as a proof of concept because there are several thing we need to take care
    a. we conditionally have to extend the length as you mentioned
    b. **knitr** is often used in combination with Pandoc via **rmarkdown**, however, there are some exceptions like **xaringan**.
  
Maybe 2-b is ignorable as this is a sort of advanced feature.

**Comment by yihui** [comment]
Yes, please feel free to ignore 2-b for now.

**Comment by atusy** [comment]
Sounds good.

```` r
library(knitr)
cat(knitr::knit(quiet = TRUE, text = c(
  '```{r, class.chunk="foo"}',
  '"classed div"',
  "```",
  "",
  '```{r, class.chunk=""}',
  '"classless div"',
  "```",
  "",
  '```{r, class.chunk=""}',
  'knitr::asis_output("lengthen div\n:::")',
  "```",
  "",
  '```{r}',
  '"no div"',
  "```"
)))
#> ::: {.foo}
#> 
#> 
#> ```r
#> "classed div"
#> #> [1] "classed div"
#> ```
#> 
#> :::
#> 
#> ::: 
#> 
#> 
#> ```r
#> "classless div"
#> #> [1] "classless div"
#> ```
#> 
#> :::
#> 
#> :::: 
#> 
#> 
#> ```r
#> knitr::asis_output("lengthen div
#> :::")
#> ```
#> 
#> lengthen div
#> :::
#> 
#> ::::
#> 
#> 
#> ```r
#> "no div"
#> #> [1] "no div"
#> ```
````

<sup>Created on 2022-01-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

File: R/hooks-md.R [hunk 1]
```diff
@@ -213,6 +214,29 @@ hooks_markdown = function(strict = FALSE, fence_char = '`') {
...
+pandoc_div = function(x, .attr = NULL, .class = NULL) {
+  if (is.null(.attr) && is.null(.class)) return(x)
+  if (!is.character(x) || length(x) != 1L) {
+    warning('x is unprocessed because x is not a character vector of length 1.')
+    return(x)
+  }
+  n = 1L + max(
+    2L,
+    vapply(
+      gregexpr('^:{3,}', strsplit(x, '\n')[[1L]]),
```
atusy, line 226: Here I simply look for all lines starting with `:::`.
This is because we should take into account start of fenced div as well.

An example below encloses a paragraph by two `:::`, however, the first one is considered as a paragraph, and the internal `:::: .bar` is paired with the second `:::`.

``` bash
echo '::: .foo\n\n:::: .bar\n\n:\n\n:::' | pandoc
<p>::: .foo</p>
<div class=".bar">
<p>:</p>
</div>
```

yihui: That sounds good.

yihui (approved):
Thank you!

PR #2106, opened 2022-02-09
fix bug in Sweave2knitr when removing unnecessary chunk ends

chroetz:
`Sweave2knitr(text = c("<<>>=", "@"))` wrongly removed the line "@". 

Now it correctly yields `c("<<>>=", "@")`. 
Furthermore, chunk ends (`"@"`) that are not part of a chunk are removed:
```
Sweave2knitr(text = c("@", "a", "@", "<<>>=", "@", "@", "a", "@"))
#> [1] "a"     "<<>>=" "@"     "a"
```


yihui (approved):
I admit that I'm unable to quickly understand my original code now, but your new code seems to be both more reasonable and easier to understand. Thanks!

PR #2110, opened 2022-03-07
Update GHA workflow

cderv:
* Update to r-lib/actions v2
* test 4 older R versions as with **rmarkdown** to prevent bug found after release.


yihui (approved):
Looks good. Thanks!

PR #2115, opened 2022-03-19
Add rnw2pdf(), an opinionated wrapper to convert Rnw to PDF

### LINKED ISSUE #2113: questions on knit2pdf (output only allows ".tex" file extension,  can't produce to different folder, no clean arg)
Hi, 

I've been using `knitr::knit2pdf()` for years but there're some questions on designing that I don't quite understand since the very beginning:

1. The `output` argument must be with ".tex" file extension. It just produces a pdf file with the same name. Since the name is `knit2pdf()`, shouldn't it allow a ".pdf" file extension?
2. There's no way of removing the "tex" and "log" files, automatically.
3. The value provided to the `output` argument must be in the same folder of the Rnw file. Why can't it allow generating the output into another folder?

Is it because of historical reasons or am I using the wrong function?

shrektan:
Closes #2113 

The new (opinionated) `rnw2pdf()` is very useful (for me) in ways that:

1. Sets the default compiler to "xelatex". This is often desired for non-ASCII users and a natural replacement (I think) for "pdflatex".
2. Checks if `output` file is locked before knitting, which is very useful on Windows.
3. Stops (by default) when the knitting content throws any error. This is often desired for automated reports.
4. Allows to clean the intermediate files, e.g., ".tex"
5. Allows any output path. In contrast, `knit2pdf()` only allows the output to be in the same folder as input.
6. The output file is "xxx.pdf" but in `knit2pdf()` it's "xxx.tex". 

I'm not sure if a new function is desired in this package, as it's already very complicated, so feel free to decline this PR. But `knit2pdf()` supports both Rnw and rst files and has been used by many users already, I don't want to take the risk to break their code.


yihui (approved):
I think this is a good idea. I'll tweak this PR and merge it later. Thanks!

R/utils-conversion.R
```diff
@@ -83,6 +83,50 @@ knit2pdf = function(
...
+#' @return The \code{output} file path.
+#' @export
+rnw2pdf = function(
+  input, output = with_ext(input, 'pdf'), compiler = 'xelatex',
+  envir = parent.frame(), quiet = FALSE, clean = TRUE, error = FALSE, ...
+) {
+  # On Windows, when tweaking the content, users may forget to close the PDF
+  # file (thus can't be written). Since knitting may take quite some time, it's
+  # better to check the write permission of the output file in advance.
+  if (xfun::file_exists(output) && !file.access(output, 2)) {
```
yihui, line 118: Here I'm using `file.access(mode = 2)` to test if the output file is writable, instead of trying to delete the file. The former should be safer, although it probably doesn't make a practical difference. Is that okay?

R/utils-conversion.R
```diff
@@ -83,6 +83,50 @@ knit2pdf = function(
...
+    stop("The file '", output, "' is not writable (may be locked by a PDF reader).")
+  }
+  old = opts_chunk$set(error = error)
+  on.exit(opts_chunk$set(old), add = TRUE)
+  file_tex = knit(input, envir = envir, quiet = quiet)
```
yihui, line 123: Here I'm letting `knit()` decide the output `.tex` file path, instead of using `with_ext(input, 'tex')`.

R/utils-conversion.R
```diff
@@ -83,6 +83,50 @@ knit2pdf = function(
...
+  if (clean) on.exit(file.remove(file_tex), add = TRUE)
+  file_pdf = tinytex::latexmk(file_tex, engine = compiler, clean = clean, ...)
```
yihui, line 125: Similarly, here I'm letting `latexmk()` decide the `.pdf` file path.

R/utils-conversion.R
```diff
@@ -83,6 +83,50 @@ knit2pdf = function(
...
+  if (!xfun::same_path(output, file_pdf)) file.rename(file_pdf, output)
```
yihui, line 126: The function `same_path()` in **xfun** does exactly what you did in your implementation.

yihui (approved):
I've finished tweaking this PR. Please check if the changes are okay to you. Thanks!

R/utils-conversion.R
shrektan: Does `file.access(output, 2)` work on Windows? It always reports `0` no matter what in my test.

shrektan: All other changes look good to me, thanks!

yihui: Okay, I'll switch back to `file.remove()` then.

cderv: @shrektan I can confirm that `file.access(mode = 2)` works on Windows

````r
# Create a new file
> file.create("test.txt")
[1] TRUE
# By default use has all right on it, so return is 0 (success for writable)
> file.access("test.txt", mode = 2)
test.txt 
       0 

# In Windows Explorer, Properties, change access right for user to read-only
# it now returns -1 has the file is not writable
> file.access("test.txt", mode = 2)
test.txt 
      -1 
````

yihui: @cderv You can check what `file.access(mode = 2)` returns when a file is locked or occupied by an application or process, e.g., a PDF file opened by Acrobat Reader, or a `.dll` file loaded by an R package.

cderv: It seems that don't work the same: 

* `test2.pdf` is open in a PDF reader
* and I get this 
````r
> file.access("test2.pdf", mode = 2)
test2.pdf 
        0 
````

Also for package 
````r
> library(rlang)
> find.package("rlang")
[1] "C:/Users/chris/Documents/R/win-library/4.1/rlang"
> file.access("C:/Users/chris/Documents/R/win-library/4.1/rlang/libs/x64/rlang.dll")
C:/Users/chris/Documents/R/win-library/4.1/rlang/libs/x64/rlang.dll 
                                                                  0
````

whereas the dll is indeed locked in several paces
````r
> pak:::get_locked_libs(.libPaths()[1], "rlang")
# A tibble: 9 x 7
  dll       path           pid name  username ps_handle  package
  <chr>     <chr>        <int> <chr> <chr>    <I<list>>  <chr>  
1 rlang.dll "C:\\Users~  27576 Rter~ "DESKTO~ <ps_handl> rlang  
2 rlang.dll "C:\\Users~  45012 Rter~ "DESKTO~ <ps_handl> rlang  
3 rlang.dll "C:\\Users~  52540 Rter~ "DESKTO~ <ps_handl> rlang  
4 rlang.dll "C:\\Users~  62080 rses~ "DESKTO~ <ps_handl> rlang  
5 rlang.dll "C:\\Users~  69648 Rter~ "DESKTO~ <ps_handl> rlang  
6 rlang.dll "C:\\Users~  70624 rses~ "DESKTO~ <ps_handl> rlang  
7 rlang.dll "C:\\Users~  85752 Rter~ "DESKTO~ <ps_handl> rlang  
8 rlang.dll "C:\\Users~  88624 Rter~ "DESKTO~ <ps_handl> rlang  
9 rlang.dll "C:\\Users~ 102396 Rter~ "DESKTO~ <ps_handl> rlan
````

shrektan: Yes, we can't rely on `file.access()` to decide if a file is locked or not. I'm not a fan of using `file.remove()`, either. If there's another way to detect the "unremovable" status, I'd love to hear!

PR #2117, opened 2022-03-29
fix: file.remove() raises warning when the file is not there

shrektan:
A patch to #2115.

image

`file.remove()` should be only performed when the file exists, or it raises an annoying warning.

I confirm that it returns `FALSE` when the file is locked on Windows and `TRUE` if it's removable.


yihui (approved):
Thank you!

PR #2120, opened 2022-04-06
defer evaluation of yaml cell options

jjallaire:
This PR makes my motivating example work correctly:

````markdown
---
title: "test"
output: html_document
---

```{r}
x <- "Air"
```

```{r}
#| label: fig-test
#| fig-cap: !expr paste(x, "Quality")
plot(mtcars)
```
````

@yihui It's definitely possible that I haven't thought of everything required here as I don't have a full picture in my mind of the parsing/options/execution pipeline.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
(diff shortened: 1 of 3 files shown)


yihui (approved):
I think it should be good. Thanks!

PR #2127, opened 2022-05-12
check file and code chunk option when checking label duplication

### LINKED ISSUE #2126: When `file=` chunk option, duplication of chunk label is not checked.
Reported by @mine-cetinkaya-rundel

Reprex: 

* Will error 

````markdown

```{r abcd}
1 + 1
```


```{r abcd}
2 + 2
```
`````
 
```r
> knitr::knit("test.Rmd")


processing file: test.Rmd
Error in parse_block(g[-1], g[1], params.src, markdown_mode) : 
  Duplicate chunk label 'abcd', which has been used for the chunk:
1 + 1
````

* Will not error - no check duplication ?

````markdown
```{cat, engine.opts = list(file = "_common.R")}
1 + 1
```

```{r abcd, code = xfun::read_utf8("_common.R")}
```


```{r abcd}
2 + 2
```

```{r}
unlink("_common.R")
```
````

```r
> knitr::knit("test.Rmd")


processing file: test.Rmd
  |...........                                                                                      |  11%
  ordinary text without R code

  |......................                                                                           |  22%
label: unnamed-chunk-1 (with options) 
List of 2
 $ engine.opts:List of 1
  ..$ file: chr "_common.R"
 $ engine     : chr "cat"

  |................................                                                                 |  33%
  ordinary text without R code

  |...........................................                                                      |  44%
label: abcd (with options) 
List of 1
 $ code: chr "1 + 1"

  |......................................................                                           |  56%
  ordinary text without R code

  |.................................................................                                |  67%
label: abcd
  |...........................................................................                      |  78%
  ordinary text without R code

  |......................................................................................           |  89%
label: unnamed-chunk-2
  |.................................................................................................| 100%
  ordinary text without R code


output file: test.md

[1] "test.md"
```

Same with 

````markdown
```{r abcd, file = "_common.R"}
```
````

We probably want the duplication chunk to still be done, right ?

### PR DESCRIPTION
**Pull request by cderv**
This PR aim to fix #2126

### DIFF
--- a/R/parser.R
+++ b/R/parser.R
@@ -114,8 +114,8 @@ parse_block = function(code, header, params.src, markdown_mode = out_format('mar
         one_string(knit_code$get(label))
       )
     }
-    if (length(code))
-      knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
+    code = as.character(code)
+    knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
   }
 
   # store dependencies
--- a/R/parser.R
+++ b/R/parser.R
@@ -114,7 +114,8 @@ parse_block = function(code, header, params.src, markdown_mode = out_format('mar
         one_string(knit_code$get(label))
       )
     }
-    knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
+    if (length(code))
+      knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
   }
 
   # store dependencies
--- a/R/parser.R
+++ b/R/parser.R
@@ -105,7 +105,7 @@ parse_block = function(code, header, params.src, markdown_mode = out_format('mar
   code = parts$code
 
   label = params$label; .knitEnv$labels = c(.knitEnv$labels, label)
-  if (length(code)) {
+  if (length(code) || length(params$file) || length(params$code)) {
     if (label %in% names(knit_code$get())) {
       if (identical(getOption('knitr.duplicate.label'), 'allow')) {
         params$label = label = unnamed_chunk(label)
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - The internal function `knitr:::wrap()` has been removed from this package. If you rely on this function, you will have to use the exported function `knitr::sew()` instead.
 
+- Duplicate chunk label error will now be thrown with code chunks using `code` or `file` chunk options (thanks, @mine-cetinkaya-rundel, #2126).
+
 # CHANGES IN knitr VERSION 1.39
 
 ## MAJOR CHANGES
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by cderv** [comment]
@yihui small thing I encountered that need your insight: 

Initially the check for duplication happens only for `code` meaning when chunk content is not empty. And then `knit_code$set` is used to save the chunk. 

When `file` or `code` chunks option are used, we can do the duplication check, but `code` content is NULL and we can't create the `structure()` to save in `knit_code` without resolving the chunk option as we are doing later. This is because `structure(NULL)` does not work. 

Related code: 

* Check duplication and saving
  https://github.com/yihui/knitr/blob/925b9b345528ba9689b4a4144ba35ab6d952e55b/R/parser.R#L108-L115
  
* Resolving the content using the option
   https://github.com/yihui/knitr/blob/925b9b345528ba9689b4a4144ba35ab6d952e55b/R/block.R#L33-L40
   
To solve this, I don't save the content to `knit_code` when an option is used.  But this means this PR fix the initial issue only if the empty chunk with same label appears after another one with the same name. If the order is different, then it still won't error

* Still don't error
````markdown
```{cat, engine.opts = list(file = "_common.R")}
1 + 1
```

```{r abcd, code = xfun::read_utf8("_common.R")}
```

```{r abcd}
2 + 2
```

```{r}
unlink("_common.R")
```
````

* Will error
````markdown
```{cat, engine.opts = list(file = "_common.R")}
1 + 1
```

```{r abcd}
2 + 2
```

```{r abcd, code = xfun::read_utf8("_common.R")}
```

```{r}
unlink("_common.R")
```
````

So this is a bit more complicated that discussed it seems.

File: R/parser.R [hunk 1]
```diff
@@ -114,7 +114,8 @@ parse_block = function(code, header, params.src, markdown_mode = out_format('mar
         one_string(knit_code$get(label))
       )
     }
-    knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
+    if (length(code))
+      knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
```
cderv, line 118: I did that to keep the current behavior as we have really nothing to save in `knit_code`. 

Would it make sense to save an empty code in this case ? Maybe that would mess up some part of the code 🤔 

Something like 
````diff
-knit_code$set(setNames(list(structure(code, chunk_opts = params)), label))
+knit_code$set(setNames(list(structure(code %n% '', chunk_opts = params)), label))
```

cderv: I noticed also that using `code` or `file` won't make the code available in `purl()` output currently. 
Using `''` for `knit_code` for those chunk will still keep that.

cderv:
Thanks @yihui your change indeed fix both order. 

I think we can merge this to fix #2126 and maybe improve the logic (by registering the code content from `file` or `code`) in a later state ? 

If that is useful....

yihui (approved):
> maybe improve the logic (by registering the code content from `file` or `code`) in a later state ?

I just added that to this PR so we don't need to remember to do it later. Thanks!

PR #2128, opened 2022-05-18
sql engine: add sql.args option to pass extra arguments to DBI query functions

### LINKED ISSUE #2241: [FR] Run SQL chunk with DBI option `immediate = TRUE` ?
I understand from the [documentation](https://bookdown.org/yihui/rmarkdown/language-engines.html#sql) that DBI is used in backend to run SQL chunk.

I try to reproduce the behaviour of running of R instruction similar to:
```r
DBI::dbGetQuery(con, statement = "SELECT * FROM mytable LIMIT 10", immediate = TRUE)
```
but I could not find any option to pass in the sql chunk header, such as: 

````
```{sql, output.var = "data", connection = "con", immediate = TRUE}
SELECT * FROM mytable LIMIT 10
```
````
I tried many things such as:
- `immediate = "TRUE/True"`
- `params = parameters` with `parameters <- list(immediate = TRUE)` ...

but could not success. 

Is there a way to achieve this I have not thought about ?
Thank you !

### PR DESCRIPTION
**Pull request by AWKruijt**
As discussed here: fixes https://github.com/rstudio/rmarkdown/issues/2241

I haven't added test files as these changes require access to an sql server with permission to create temporay tables. Haven't found that in an open database. 

I'm excited about this! :)

### DIFF
--- a/R/engine.R
+++ b/R/engine.R
@@ -560,13 +560,14 @@ is_sql_update_query = function(query) {
   grepl('^\\s*(INSERT|UPDATE|DELETE|CREATE|DROP).*', query, ignore.case = TRUE)
 }
 
+
 # sql engine
 eng_sql = function(options) {
   # return chunk before interpolation eagerly to avoid connection option check
   if (isFALSE(options$eval) && !isTRUE(options$sql.show_interpolated)) {
     return(engine_output(options, options$code, ''))
   }
-
+  
   # Return char vector of sql interpolation param names
   varnames_from_sql = function(conn, sql) {
     varPos = DBI::sqlParseVariables(conn, sql)
@@ -575,12 +576,12 @@ eng_sql = function(options) {
       sub('^\\?', '', varNames)
     }
   }
-
+  
   # Vectorized version of exists
   mexists = function(x, env = knit_global(), inherits = TRUE) {
     vapply(x, exists, logical(1), where = env, inherits = inherits)
   }
-
+  
   # Interpolate a sql query based on the variables in an environment
   interpolate_from_env = function(conn, sql, env = knit_global(), inherits = TRUE) {
     names = unique(varnames_from_sql(conn, sql))
@@ -588,14 +589,14 @@ eng_sql = function(options) {
     if (length(names_missing) > 0) {
       stop("Object(s) not found: ", paste('"', names_missing, '"', collapse = ", "))
     }
-
+    
     args = if (length(names) > 0) setNames(
       mget(names, envir = env, inherits = inherits), names
     )
-
+    
     do.call(DBI::sqlInterpolate, c(list(conn, sql), args))
   }
-
+  
   # extract options
   conn = options$connection
   if (is.character(conn)) conn = get(conn, envir = knit_global())
@@ -608,13 +609,14 @@ eng_sql = function(options) {
     max.print = -1
   sql = one_string(options$code)
   params = options$params
-
+  imm = options$immediate
+  
   query = interpolate_from_env(conn, sql)
   if (isFALSE(options$eval)) return(engine_output(options, query, ''))
-
+  
   data = tryCatch({
     if (is_sql_update_query(query)) {
-      DBI::dbExecute(conn, query)
+      DBI::dbExecute(conn, query, immediate = imm)
       NULL
     } else if (is.null(varname) && max.print > 0) {
       # execute query -- when we are printing with an enforced max.print we
@@ -623,50 +625,50 @@ eng_sql = function(options) {
       data = DBI::dbFetch(res, n = max.print)
       DBI::dbClearResult(res)
       data
-
+      
     } else {
       if (length(params) == 0) {
-        DBI::dbGetQuery(conn, query)
+        DBI::dbGetQuery(conn, query, immediate = imm)
       } else {
         # If params option is provided, parameters are not interplolated
-        DBI::dbGetQuery(conn, sql, params = params)
+        DBI::dbGetQuery(conn, sql, immediate = imm, params = params)
       }
     }
   }, error = function(e) {
     if (!options$error) stop(e)
     e
   })
-
+  
   if (inherits(data, "error"))
     return(engine_output(options, query, one_string(data)))
-
+  
   # create output if needed (we have data and we aren't assigning it to a variable)
   output = if (length(dim(data)) == 2 && ncol(data) > 0 && is.null(varname)) capture.output({
-
+    
     # apply max.print to data
     display_data = if (max.print == -1) data else head(data, n = max.print)
-
+    
     # get custom sql print function
     sql.print = opts_knit$get('sql.print')
-
+    
     # use kable for markdown
     if (!is.null(sql.print)) {
       options$results = 'asis'
       cat(sql.print(data))
     } else if (out_format('markdown')) {
-
+      
       # we are going to output raw markdown so set results = 'asis'
       options$results = 'asis'
-
+      
       # force left alignment if the first column is an incremental id column
       first_column = display_data[[1]]
       if (is.numeric(first_column) && length(first_column) > 1 && all(diff(first_column) == 1))
         display_data[[1]] = as.character(first_column)
-
+      
       # wrap html output in a div so special styling can be applied
       add_div = is_html_output() && getOption('knitr.sql.html_div', TRUE)
       if (add_div) cat('<div class="knitsql-table">\n')
-
+      
       # determine records caption
       caption = options$tab.cap
       if (is.null(caption)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by AWKruijt** [comment]
reptable =  gsub('(^.*into[[:space:]]+)(#.+)([[:space:]]+from.*$)', '\\2', sql, ignore.case = T)
on line 618 fixes the error handling on attempts to replace non-temporary tables.

**Comment by AWKruijt** [comment]
Just checking (I'm really inexperiences >.<): I'm seeing this super inviting 'Update branch' button (with alternate update with rebase) - do I push that or do I just wait now? :)

**Comment by cderv** [comment]
You can push that to update but that is not necessary. We'll do it ourself, or just merge the branch. Your change target in code part that has not been modified in some time, and tests are passing ok. So you can leave that way, and just wait for us to review. I added that in my TODO list. Thanks !

**Review by yihui — APPROVED** [review]
I don't really know much about SQL, so I only reviewed the coding style. BTW, the changes with pure white spaces in this PR need to be discarded (if you don't know how, we can handle this before we merge).

Thanks!

File: R/engine.R [hunk 1]
```diff
@@ -608,13 +609,26 @@ eng_sql = function(options) {
...
+    immediate= T
+    replace=  F
+    if(isTRUE(options$replace)) {
+        replace= T
+        reptable =  gsub('(^.*into[[:space:]]+)(#.+)([[:space:]]+from.*$)', '\\2', sql, ignore.case = T)
+        if(sub('(.).*.$', '\\1', reptable) != '#') stop2(
+          "To replace a table, the table has to be a temporary table (tablename staring with # or ##)."
+        )
+        sql <- paste0("if object_id('tempdb.dbo.", reptable, "') is not null drop table ", reptable, " ;", sql)
+    }
```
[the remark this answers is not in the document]
AWKruijt: Thanks!  I'll try to get all of it solved in the coming days :)

R/engine.R
```diff
@@ -608,13 +609,26 @@ eng_sql = function(options) {
...
+  }
+  
   query = interpolate_from_env(conn, sql)
   if (isFALSE(options$eval)) return(engine_output(options, query, ''))
-
+  
   data = tryCatch({
     if (is_sql_update_query(query)) {
-      DBI::dbExecute(conn, query)
+      DBI::dbExecute(conn, query, immediate = immediate)
```
cderv, line 631: I think we need to take precaution here. 
`DBI::dbExecute()` is a generic and `immediate` argument, is indeed among the specifications. However, `immediate` is not part of the generic argument to improve compatibility with the different backends. 
https://dbi.r-dbi.org/reference/dbexecute#additional-arguments-1

> The following arguments are not part of the dbExecute() generic (to improve compatibility across backends) but are part of the DBI specification:
> * params (default: NULL)
> * immediate (default: NULL)
> 
> They must be provided as named arguments. See the "Specification" sections for details on their usage

This is valid for other calls too.

R/engine.R
```diff
@@ -608,13 +609,26 @@ eng_sql = function(options) {
     max.print = -1
   sql = one_string(options$code)
   params = options$params
-
+  immediate=  F
```
cderv, line 612: **DBI** seems to default to NULL for `dbExecute` and `dbGetQuery`, which means 
https://dbi.r-dbi.org/reference/dbexecute#specification-for-the-immediate-argument-1
> The default NULL means that the backend should choose whatever API makes the most sense for the database

So setting FALSE, will change behavior for all the DB that backends that will use TRUE by default. 

Maybe we should stick to that, and pass `immediate = TRUE` or `FALSE` when the chunk option is explicitly set ?

cderv, line 612: **DBI** seems to default to NULL for `dbExecute` and `dbGetQuery`, which means 
https://dbi.r-dbi.org/reference/dbexecute#specification-for-the-immediate-argument-1
> The default NULL means that the backend should choose whatever API makes the most sense for the database

So setting FALSE, will change behavior for all the DB that backends that will use TRUE by default. 

Maybe we should stick to that, and pass `immediate = TRUE` or `FALSE` when the chunk option is explicitly set ?

cderv (changes requested):
Thanks. 

I made a few comments to share a few thoughts. 
Currently, with this change : 

* We are passing `immediate` no matter what
* We are setting the default value to `FALSE`  whereas it seems to be NULL in **DBI** which has a different meaning.

I don't really know which backend that would impact or not, but I think it would be safer to prevent this somehow using more conditional logic. I can help tweak the PR if you do not know how to do it. 

Thanks

R/engine.R
```diff
@@ -608,13 +609,26 @@ eng_sql = function(options) {
...
+  if(isTRUE(options$immediate)) {
+    immediate= T
+    replace=  F
+    if(isTRUE(options$replace)) {
+        replace= T
```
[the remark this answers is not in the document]
yihui: Yes, let's use the `sql.` prefix for the new options here.

> I don't know if we can end up with a rule on under which condition which should use `engine.opts`, bare chunk option name or prefixed option name. I think we should try being consistent.

I agree. It has been a mess, which I don't like, but I feel it's tough to decide which is better:

````md
```{sql, engine.opts = list(immediate = TRUE, replace = TRUE)}
````

````md
```{sql, sql.immediate = TRUE, sql.replace = TRUE}
````

> chunk engine is enough to scope the variable (advanced usage of chunk options hooks would require checking the engine, and then checking the option in case of same option name used by two different engine with different meaning - may never happen though).

That's also what I think.

**Comment by AWKruijt** [comment]
Sorry for the radio silence - things where happening. Thanks for all the input! I've implemented as much as I could - see commit [358d552](https://github.com/yihui/knitr/pull/2128/commits/358d55248dc07c6147b4276fea793bf6312cea65).

A big one that is open for debate still is whether/how to make it so that paramater immediate is NOT passed to dbGetQuery when the user did not set it. I suppose that's what we want, given that it's cleanest if we also don't pass the NULL value if sql.immediate is not actively invoked. For option params the code has a split based on length(params) == 0 and separately defined calls to DBI::dbGetQuery() with and without passing of params. We could expand that into four calls (with/without immediate and/or params), but perhaps something along the lines of a do.call with a list of arguments could be preferable? I did try every trick I could think of along the lines of DBI::dbGetQuery(conn, sql, if(exists('immedate'){immedate = immediate}) but no joy :/

**Comment by AWKruijt** [comment]
Alas, I notice that in practice the replace option doesn't always work flawlessly due to the regex to lift the tempary table name from the query also picking up new lines (`/n`) immediately after the #table name. Fixed the regex and made another correction to the getQuery call with params where I had made it use `sql` instead of `query`.

**Review by cderv — COMMENTED** [review]
I am not sure I understand why we need the `exist()` for retrieving the option. I commented about that. 

> We could expand that into four calls (with/without immediate and/or params), but perhaps something along the lines of a do.call with a list of arguments could be preferable? I did try every trick I could think of along the lines of DBI::dbGetQuery(conn, sql, if(exists('immedate'){immedate = immediate}) but no joy :/

After looking closer to DBI, I think the most important thing is not changing the default. So that 
````
dbExecute(con, immediate = options$sql.immediate)
```` 
correctly pass NULL value. I don't think that methods that don't have `immediate` will break because of the `...` - it will just be unused argument in this. So we may be fine passing it anyway. 

Otherwise, the `do.call` alternative seems like a good one, with a list of argument containing `immediate` or not depending on the option. Did you try that already ? 

Also, did you come up with test document for this PR ? Just curious to avoid making a new one. 

Thanks

**Comment by yihui** [comment]
Rebased onto current `master` (on top of the `sql.is_statement` work from #1896) and reworked to address @cderv's review (CHANGES_REQUESTED).

Key change per the review: **the DBI default for `immediate` (`NULL`) is preserved**. `immediate` is not part of the `dbExecute()`/`dbGetQuery()` generics, and `NULL` means "let the backend choose the most suitable API", so we must not force `FALSE`. The new chunk option `sql.immediate` is now passed to DBI (via `do.call()`, the approach you suggested) **only when it is explicitly set** to `TRUE`/`FALSE`; otherwise the argument is omitted entirely. This is the `do.call` alternative discussed in the thread, avoiding the `exists()` trick.

Also uses the `sql.` option prefix (`sql.immediate`) for consistency with `sql.print`/`sql.show_interpolated`, and fixes the `dbGetQuery(conn, sql, params=)` → `query` reference.

Scope note: I dropped the `sql.replace` / temporary-table feature from the original PR. Its table-name regex was acknowledged upthread to be fragile (newlines, `#temp` parsing) and it is SQL-Server-specific, which is hard to test in CI. This PR now focuses on `sql.immediate` alone; `replace` can be revisited in a separate PR if there's demand.

Thanks @AWKruijt for the original contribution and @cderv for the detailed review.

**Comment by yihui** [comment]
This adds a general `sql.args` chunk option — a named list forwarded to the underlying DBI query functions (`dbExecute()`, `dbSendQuery()`, `dbGetQuery()`) — so any backend-specific argument can be passed, e.g. `sql.args = list(immediate = TRUE)`. Extra args are passed only when supplied, so DBI's own defaults are preserved otherwise. Added validation, a NEWS entry with an example, and verified dispatch via mocked DBI. Thanks @AWKruijt!

File: R/engine.R [hunk 5]
```diff
@@ -635,24 +635,33 @@ eng_sql = function(options) {
...
   )
 
+  # extra arguments to be passed to the DBI query functions (dbExecute(),
+  # dbSendQuery(), dbGetQuery()). The chunk option sql.args may be a named list
+  # of any arguments supported by these functions (e.g., the `immediate`
+  # argument). These arguments are passed only when supplied, so DBI's own
+  # defaults are otherwise preserved.
+  extra_args = options$sql.args %n% list()
+  if (!is.list(extra_args) || (length(extra_args) > 0 && is.null(names(extra_args))))
+    stop2("The 'sql.args' chunk option must be a named list.")
```
yihui:
```suggestion
```

yihui (approved):

Issue #2134, opened 2022-06-08
[Feature request] Tabular Environment Option for kable


amarakon:
There is already an existing option in the `kable()` function called `table.envir` which can change the environment as opposed to the default floating environment which is `table`.
But I also want to change the `tabular` environment and also have an option to remove the `table` environment.
So there should be another option called `tabular.envir`.
For example, I want the following to be possible with kable:

```
\begin{xltabular}
\caption{Example table}\\
\label{tab:example-table}\\
\end{xltabular}
```

cderv:
For reference follow-up from https://community.rstudio.com/t/how-can-i-prevent-r-markdown-from-escaping-characters-in-custom-blocks/138890/12

cderv:
Is this different than #2135 ? 🤔 is #2138 covering this too ?

amarakon:
> Is this different than #2135 ? thinking is #2138 covering this too ?

kable currently has support for tabular (by default) and longtable.
For my fork, I added support for tabularx and xltabular too.
But what if someone wants to use neither tabular nor longtable, tabularx, or xltabular?
I think there should be an option to change that environment but for me it's not that important anymore since I submitted a pull request for tabularx and xltabular.

cderv:
It makes sense, thanks for the precision.

amarakon:
I will close this issue since my pull request has been merged.

cderv:
Thanks!

PR #2138, opened 2022-06-09
Support for tabularx and xltabular

amarakon:
So far kable only supports tabular and longtable.
But tabularx and xltabular are popular among LaTeX users.
I added support for them: all you have to do is use `tabularx = TRUE` or `xltabular = TRUE`.
I also added an option called `header`, which allows you to specify things to put before the body of the table.

### DIFF
--- a/R/table.R
+++ b/R/table.R
@@ -274,7 +274,7 @@ knit_print.knitr_kable = function(x, ...) {
 }
 
 kable_latex = function(
-  x, booktabs = FALSE, longtable = FALSE, tabularx = FALSE, xltabular = FALSE, valign = 't', position = '', table.length = '', centering = TRUE,
+  x, booktabs = FALSE, longtable = FALSE, tabularx = FALSE, xltabular = FALSE, valign = 't', position = '', table.length = '', header = '', centering = TRUE,
   vline = getOption('knitr.table.vline', if (booktabs) '' else '|'),
   toprule = getOption('knitr.table.toprule', if (booktabs) '\\toprule' else '\\hline'),
   bottomrule = getOption('knitr.table.bottomrule', if (booktabs) '\\bottomrule' else '\\hline'),
@@ -321,7 +321,9 @@ kable_latex = function(
     sprintf('\n%s', toprule), '\n',
     if (!is.null(cn <- colnames(x))) {
       cn = escape_latex_table(cn, escape, booktabs)
-      paste0(paste(cn, collapse = ' & '), sprintf('\\\\\n%s\n', midrule))
+      paste0(paste(cn, collapse = ' & '),
+      if (header != '') sprintf('\\\\\n%s\n%s\n', header, midrule) else sprintf('\\\\\n%s\n', midrule))
+
     },
     one_string(apply(x, 1, paste, collapse = ' & '), sprintf('\\\\%s', linesep), sep = ''),
     sprintf('\n%s', bottomrule),
--- a/R/table.R
+++ b/R/table.R
@@ -274,7 +274,7 @@ knit_print.knitr_kable = function(x, ...) {
 }
 
 kable_latex = function(
-  x, booktabs = FALSE, longtable = FALSE, valign = 't', position = '', centering = TRUE,
+  x, booktabs = FALSE, longtable = FALSE, tabularx = FALSE, xltabular = FALSE, valign = 't', position = '', table.length = '', centering = TRUE,
   vline = getOption('knitr.table.vline', if (booktabs) '' else '|'),
   toprule = getOption('knitr.table.toprule', if (booktabs) '\\toprule' else '\\hline'),
   bottomrule = getOption('knitr.table.bottomrule', if (booktabs) '\\bottomrule' else '\\hline'),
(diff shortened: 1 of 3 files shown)


amarakon:
> Instead of three boolean arguments, how about using one character argument like `tabular.env`? We can keep the `longtable` argument for backward compatibility.

`tabularx` and `xltabular` require different LaTeX code than `tabular` and `longtable`. This is why I have to create distinct options for them. If you take a `tabular` or `longtable` table and replace the environment with `tabularx` or `xltabular`, it will not work.
However, there are some people that might want to use neither `tabular`, `longtable`, `tabularx`, nor `xltabular`, which is why I openend issue #2134.

amarakon:
> > Instead of three boolean arguments, how about using one character argument like `tabular.env`? We can keep the `longtable` argument for backward compatibility.
> 
> `tabularx` and `xltabular` require different LaTeX code than `tabular` and `longtable`. This is why I have to create distinct options for them. If you take a `tabular` or `longtable` table and replace the environment with `tabularx` or `xltabular`, it will not work. However, there are some people that might want to use neither `tabular`, `longtable`, `tabularx`, nor `xltabular`, which is why I openend issue #2134. See this [comment](https://github.com/yihui/knitr/issues/2134#issuecomment-1152510224).

**Edit:** Or you could  add `tabular.env` but make it act differently for different arguments. It should work with `longtable`, `tabularx`, and `xltabular`.

R/table.R
```diff
@@ -274,7 +274,7 @@ knit_print.knitr_kable = function(x, ...) {
 }
 
 kable_latex = function(
-  x, booktabs = FALSE, longtable = FALSE, valign = 't', position = '', centering = TRUE,
+  x, booktabs = FALSE, longtable = FALSE, tabularx = FALSE, xltabular = FALSE, valign = 't', position = '', table.length = '', header = '', centering = TRUE,
```
yihui, line 277:
```suggestion
  x, booktabs = FALSE, longtable = FALSE, tabular = if (longtable) 'longtable' else 'tabular', valign = 't', position = '', table.length = '', header = '', centering = TRUE,
```

R/table.R
```diff
@@ -310,21 +310,25 @@ kable_latex = function(
   x = escape_latex_table(x, escape, booktabs)
   if (!is.character(toprule)) toprule = NULL
   if (!is.character(bottomrule)) bottomrule = NULL
-  tabular = if (longtable) 'longtable' else 'tabular'
+  tabular = if (longtable) 'longtable' else if (tabularx) 'tabularx' else if (xltabular) 'xltabular' else 'tabular'
```
yihui, line 313:
```suggestion
```

R/table.R
```diff
@@ -310,21 +310,25 @@ kable_latex = function(
...
+  if (table.length == '') table.length = '\\linewidth'
 
   paste(c(
-    if (!longtable) c(env1, cap, centering),
-    sprintf('\n\\begin{%s}%s', tabular, valign), align,
-    if (longtable && cap != '') c(cap, '\\\\'),
+    if (!longtable & !xltabular) c(env1, cap, centering),
```
yihui, line 317:
```suggestion
    if (!tabular %in% c('longtable', 'xltabular')) c(env1, cap, centering),
```

R/table.R
```diff
@@ -310,21 +310,25 @@ kable_latex = function(
...
-      paste0(paste(cn, collapse = ' & '), sprintf('\\\\\n%s\n', midrule))
+      paste0(paste(cn, collapse = ' & '),
+      if (header != '') sprintf('\\\\\n%s\n%s\n', header, midrule) else sprintf('\\\\\n%s\n', midrule))
+
     },
     one_string(apply(x, 1, paste, collapse = ' & '), sprintf('\\\\%s', linesep), sep = ''),
     sprintf('\n%s', bottomrule),
     sprintf('\n\\end{%s}', tabular),
-    if (!longtable) env2
+    if (!longtable & !xltabular) env2
```
yihui, line 331:
```suggestion
    if (!tabular %in% c('longtable', 'xltabular')) env2
```

yihui (commented):
> Or you could add `tabular.env` but make it act differently for different arguments.

Yes, that's what I meant. The current `tabularx` argument would be equivalent to using `tabular.env == 'tabularx'`. In my code suggestions, I was using `tabular` instead of `tabular.env`. Are you okay with the suggested changes? (I didn't test them but just typed them directly on Github)

Thanks!

amarakon:
> > Or you could add `tabular.env` but make it act differently for different arguments.
> 
> Yes, that's what I meant. The current `tabularx` argument would be equivalent to using `tabular.env == 'tabularx'`. In my code suggestions, I was using `tabular` instead of `tabular.env`. Are you okay with the suggested changes? (I didn't test them but just typed them directly on Github)
> 
> Thanks!

I tested your changes and tried to make the following table:

```r
Lorem = c("Ipsum")
Dolor = c("Sit")

knitr::kable(data.frame(Lorem, Dolor), escape = FALSE, booktabs = TRUE, linesep = "\\addlinespace", tabular = "xltabular", caption = "Caption")
```

But it threw this error:

```
Error in paste(c(if (!tabular %in% c("longtable", "xltabular")) c(env1,  :
  object 'xltabular' not found
Calls: <Anonymous> ... eval -> <Anonymous> -> do.call -> kable_latex -> paste
```

R/table.R
```diff
@@ -310,21 +310,25 @@ kable_latex = function(
...
+  if (table.length == '') table.length = '\\linewidth'
 
   paste(c(
-    if (!longtable) c(env1, cap, centering),
-    sprintf('\n\\begin{%s}%s', tabular, valign), align,
-    if (longtable && cap != '') c(cap, '\\\\'),
+    if (!longtable & !xltabular) c(env1, cap, centering),
+    if (tabularx | xltabular) sprintf('\n\\begin{%s}{%s}', tabular, table.length) else sprintf('\n\\begin{%s}%s', tabular, valign),
+    align,
+    if (longtable | xltabular && cap != '') c(cap, '\\\\'),
```
yihui, line 320: I think the error you saw was because of this:

```suggestion
    if ((tabular %in% c('longtable', 'xltabular')) && cap != '') c(cap, '\\\\'),
```

amarakon: Yep, you are right. Now the table works.

amarakon:
I committed the suggestions.

amarakon:
Thanks, your changes are perfect.

1. The `valign` argument is satisfactory so there is no need for `table.length`.
2. Same goes for the `midrule` argument.

Issue #2147, opened 2022-07-08
Cell width of kable when hyperlinks are included


jacobbien:
When we use `kable(tbl, format = "pipe")` and our table has one or more hyperlinks, written as `[label](my-url)`, the width of the column will be based on `nchar("[label](my-url)")` rather than on `nchar("label")`.  I've noticed this can be problematic when `my-url` is extremely long.  Here's an example:

``` r
very_long_url <- paste(c("a", rep("very", 20), "long-url"), collapse = "-")
tbl <- data.frame(
  flavor = sprintf("[Mint chip](%s)", very_long_url),
  description = "Of all the many flavors of ice cream, this is my favorite."
  )
knitr::kable(tbl, format = "pipe")
```
```
| flavor                                                                                                                      | description                                                |
|:----------------------------------------------------------------------------------------------------------------------------|:-----------------------------------------------------------|
| [Mint chip](a-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-long-url) | Of all the many flavors of ice cream, this is my favorite. |
```

The problem is that when one uses `rmarkdown::render()` on a Markdown file with such a table, the resulting html squeezes the second column unnecessarily:

![Screen Shot 2022-07-07 at 11 04 19 PM](https://user-images.githubusercontent.com/2008608/177927298-d3b62dbf-d353-4b6c-8501-6c7269bc0aed.png)

To reproduce the screenshot, one can download [example.zip](https://github.com/yihui/knitr/files/9069373/example.zip), which has `example.Rmd` in it, and then run the following two lines:

``` r
knitr::knit("example.Rmd", output = "example.md")
rmarkdown::render("example.md")
```

```
> xfun::session_info("knitr")
R version 4.2.1 (2022-06-23)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur 11.6.7, RStudio 2022.2.3.492

Locale: en_US.UTF-8 / en_US.UTF-8 / en_US.UTF-8 / C / en_US.UTF-8 / en_US.UTF-8

Package version:
  evaluate_0.15   glue_1.6.2      graphics_4.2.1  grDevices_4.2.1 highr_0.9      
  knitr_1.39.5    magrittr_2.0.3  methods_4.2.1   stats_4.2.1     stringi_1.7.6  
  stringr_1.4.0   tools_4.2.1     utils_4.2.1     xfun_0.31       yaml_2.3.5     
```

[...]

PR #2148, opened 2022-07-08
Fix how kable_mark() measures column width when there are hyperlinks

jacobbien:
This is intended to address #2147.

Recall the code example there:

```r
very_long_url <- paste(c("a", rep("very", 20), "long-url"), collapse = "-")
tbl <- data.frame(
  flavor = sprintf("[Mint chip](%s)", very_long_url),
  description = "Of all the many flavors of ice cream, this is my favorite."
  )
knitr::kable(tbl, format = "pipe")
```

It now generates a different output:

```
|flavor    |description                                                |
|:---------|:----------------------------------------------------------|
|[Mint chip](a-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-long-url)|Of all the many flavors of ice cream, this is my favorite. |
```

And here's what the table in the html file now looks like:

Screen Shot 2022-07-08 at 12 30 14 AM

Recall that before it looked like 

Screen Shot 2022-07-07 at 11 04 19 PM

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1059,3 +1059,9 @@ make_unique = function(x) {
 #' browseURL('logo.html') # you can check its HTML source
 #' }
 image_uri = function(f) xfun::base64_uri(f)
+
+# Change all "[label](url)" to "label"
+remove_urls <- function(x) {
+  # regex adapted from https://dev.to/mattkenefick/regex-convert-markdown-links-to-html-anchors-f7j
+  gsub("\\[([^\\]]+)\\]\\(([^\\)]+)\\)", "\\1", x, perl = TRUE)
+}
--- a/R/table.R
+++ b/R/table.R
@@ -391,13 +391,13 @@ kable_mark = function(x, sep.row = c('=', '=', '='), sep.col = '  ', padding = 0
   if (sep.col == '|') for (j in seq_len(ncol(x))) {
     x[, j] = gsub('\\|', '&#124;', x[, j])
   }
-  l = if (prod(dim(x)) > 0) apply(x, 2, function(z) max(nchar(z, type = 'width'), na.rm = TRUE))
+  l = if (prod(dim(x)) > 0) apply(x, 2, function(z) max(nchar(remove_urls(z), type = 'width'), na.rm = TRUE))
   cn = colnames(x)
   if (length(cn) > 0) {
     cn[is.na(cn)] = "NA"
     if (sep.col == '|') cn = gsub('\\|', '&#124;', cn)
     if (grepl('^\\s*$', cn[1L])) cn[1L] = rownames.name  # no empty cells for reST
-    l = pmax(if (length(l) == 0) 0 else l, nchar(cn, type = 'width'))
+    l = pmax(if (length(l) == 0) 0 else l, nchar(remove_urls(cn), type = 'width'))
   }
   align = attr(x, 'align')
   padding = padding * if (length(align) == 0) 2 else {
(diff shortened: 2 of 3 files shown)


R/utils.R
```diff
@@ -1059,3 +1059,9 @@ make_unique = function(x) {
 #' browseURL('logo.html') # you can check its HTML source
 #' }
 image_uri = function(f) xfun::base64_uri(f)
+
+# Change all "[label](url)" to "label"
+remove_urls <- function(x) {
```
yihui, line 1064:
```suggestion
remove_urls = function(x) {
```

R/utils.R
```diff
@@ -1059,3 +1059,9 @@ make_unique = function(x) {
...
+  # regex adapted from https://dev.to/mattkenefick/regex-convert-markdown-links-to-html-anchors-f7j
+  gsub("\\[([^\\]]+)\\]\\(([^\\)]+)\\)", "\\1", x, perl = TRUE)
```
yihui, line 1066: Could you also use zero-width positive/negative lookbehind/lookahead assertions (see `?regex` if you are not familiar with them) to avoid treating `` `[text](url)` `` as a URL?

yihui (changes requested):
Thanks a lot!

R/utils.R
jacobbien: Below are two examples of using positive lookbehind/lookahead for matching and replacing.  Perhaps I'm missing something, but it seems like the fact that it excludes the brackets and parentheses from the match actually makes it harder to remove them.  Both examples have the same regex, which is to look for `[x](` where `x` is 0 or more non-`]` characters.

1. Matching
```r
stringr::str_match_all("[label1](url1) and [label2](url2).","(?<=\\[)[^\\]]*(?=\\]\\()")
```
```
[[1]]
     [,1]    
[1,] "label1"
[2,] "label2"
```
2. Replacing
```r
stringr::str_replace_all("[label1](url1) and [label2](url2).","(?<=\\[)[^\\]]*(?=\\]\\()", "MATCHED")
```
```
"[MATCHED](url1) and [MATCHED](url2)."
```

But I'm new to zero-width lookbehind/lookahead, so perhaps I'm missing something.  Thanks so much for writing `knitr`, by the way -- it's such a valuable tool!

yihui: Sorry I guess I was not clear last time. I meant we should exclude the case of `[]()` surrounded by backticks (i.e., ``(?<!`)`` and ``(?!`)``).

jacobbien: Ah, I see! Thanks for explaining that.  I have updated the PR accordingly.

R/utils.R
```diff
@@ -1059,3 +1059,9 @@ make_unique = function(x) {
 #' browseURL('logo.html') # you can check its HTML source
 #' }
 image_uri = function(f) xfun::base64_uri(f)
+
+# Change all "[label](url)" to "label" except leave "`[label](url)`" as is
+remove_urls = function(x) {
+  # regex adapted from https://dev.to/mattkenefick/regex-convert-markdown-links-to-html-anchors-f7j
+  gsub("(?<!`)\\[([^\\]]+)\\]\\(([^\\)]+)\\)(?!`)", "\\1", x, perl = TRUE)
```
yihui, line 1066: I think we can have fewer backslashes here:

```suggestion
  gsub("(?<!`)\\[([^]]+)\\]\\(([^)]+)\\)(?!`)", "\\1", x, perl = TRUE)
```

yihui (approved):
Perfect. Thanks!

PR #2151, opened 2022-07-25
Convert chunk option to YAML

### LINKED ISSUE #2082: Feature request: function to convert chunk options to YAML-style
It would be neat to have a function that takes an R Markdown document and converts the chunk options to YAML-style. I'm thinking functionality like the one provided in [Rd2roxygen](https://cran.r-project.org/web/packages/Rd2roxygen/index.html), but for processing the chunk options.

---

cderv:
This close #2082 and follows #2149 

Chunk options are parsed and then written to YAML using `yaml::as.yaml()` with two tweaks to the parsed chunk options

* Logical will be `true` or `false` instead of default `yes` or `no`
* R expression or variables passed in chunk option value will be written to YAML after `!expr` to make the fact they should be evaluted before parsing. Example: 

````markdown
```{r, R.options = list(width = 80)}
names(knitr::knit_engines$get())
```
````
will be converted to 
````markdown
```{r}
#| R.options: !expr list(width = 80)
names(knitr::knit_engines$get())
```
````

Doing 
```yaml
R.options:
    width: 80
```

is a bit more complicated than the current solution, as it implies going through the all tree of expression to only find the leaf is list that are variable or expression to evaluated with `!expr`

This solution is simple enough to produce valid document, and it is easy enough for the user to reformat manually some chunk header. 

Here is how a conversion looks like on all knitr-examples: https://github.com/yihui/knitr-examples/compare/master...cderv:knitr-examples:yaml-chunk-header
Does it seems right to you ? 

It seems there is still an indent question where the chunk header is not indented, whereas the code inside is - do we want to indent the in-chunk option part too. 

Anyway, @yihui I let you review. I wanted to have something ready for the conference, so here it is.

### DIFF
--- a/R/parser.R
+++ b/R/parser.R
@@ -876,8 +876,7 @@ convert_chunk_header = function(
             result <- ifelse(x, "true", "false")
             class(result) <- "verbatim"
             return(result)
-          })), "\n")[[1]]
-
+          }), line.sep = "\n"), "\n")[[1]]
       params3 = paste0(prefix, params3)
     }
 
--- a/R/parser.R
+++ b/R/parser.R
@@ -858,12 +858,11 @@ convert_chunk_header = function(
     } else {
       params3 = parse_params(params2, label = FALSE)
 
-      # fix un-evaluated options for yaml
-      # by transforming to !expr val
+      # fix un-evaluated options for yaml by transforming to !expr val
       params3 = lapply(params3, function(x) {
         if (is.symbol(x) || is.language(x)) {
           x = deparse(x, 500L)
-          attr(x, "tag") <- "!expr"
+          attr(x, 'tag') = '!expr'
         }
         x
       })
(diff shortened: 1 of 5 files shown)


R/parser.R
```diff
@@ -858,7 +859,27 @@ convert_chunk_header = function(
         strwrap(params3, width, prefix = prefix)
       }
     } else {
-      # YAML
+      params3 = parse_params(params2, label = FALSE)
```
cderv, line 862: I duplicated this instead of refactoring the if because it seems clearer to have 

```
if type = wrap 
...
else if type = multiline
...
else if type = yaml
...
end
```

if you prefer 
```
if type = wrap 
...
else 
    res = parsed options
    if  type = multiline
    ...
    else if type = yaml
     ...
     end
end
```

cderv:
Tests are failing  because r-forge seems down... and some examples uses it.

R/parser.R
yihui: That's okay. It's only one line of duplication.

R/parser.R
```diff
@@ -858,7 +859,27 @@ convert_chunk_header = function(
...
+          attr(x, "tag") <- "!expr"
+        }
+        x
+      })
+      # convert to yaml and add prefix
+      params3 = strsplit(yaml::as.yaml(
+        params3, handlers = list(
+          # true / false instead of no
+          logical = function(x) {
+            result <- ifelse(x, "true", "false")
```
yihui: I think simply `tolower(x)` would work. Correct me if I'm wrong.

yihui (approved):
Looks perfect. Thanks!

R/parser.R
cderv: Oh yes you're right it should work woth the automatic logical to character conversion.

PR #2167, opened 2022-09-06
Fix of check in color_def() of col being NA

nielsrhansen:
As of R 4.2.0, "Calling if() with a condition of length greater than one gives an error rather than a warning." The utility function color_def() tests if the argument col is NA by if(is.na(col)), which with R 4.2.0 or greater results in an error when e.g. supplying a user specified background color for chunks in an Rnw-file. The proposed change fixes that by converting is.na(col) to length one.


yihui (approved):
Thank you!

PR #2174, opened 2022-09-22
Replace `str_replace_all()` with base equivalent

rich-iannone:
This is the first of several PRs that should slowly remove knitr's dependency on stringr (and by extension, stringi). This should be the least problematic of the `str_*()` functions that need replacement (it's just gsub()). The util functions that replace the stringr functions will live in the `utils-string_formatting.R` file and all of them will use `perl = TRUE`.

Partially addresses https://github.com/yihui/knitr/issues/1549.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -56,10 +56,12 @@ jobs:
 
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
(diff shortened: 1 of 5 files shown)


yihui (approved):
Thanks! I just tweaked the PR a little bit. I hope it makes sense.

rich-iannone:
Thanks Yihui, that’s much better :)

PR #2177, opened 2022-09-28
Replace use of `str_trim()` with base-R version

rich-iannone:
This replaces both uses of `stringr::str_trim()` with a util function that only uses base R.


yihui (approved):
This looks simple enough, so I'll just merge it, and you can move onto the next challenge :)

PR #2186, opened 2022-10-19
Replace `stringr::str_dup()` with `paste()`/`rep()`

rich-iannone:
This PR replaces two uses of `stringr::str_dup()` with a base R equivalent.


yihui (approved):
Thanks!

PR #2187, opened 2022-10-19
Replace `stringr::str_split()` with `strsplit()`

rich-iannone:
This replaces two instances of `stringr::str_split()` with the base R equivalent: `strsplit()`.


yihui (approved):
Thanks!

PR #2195, opened 2022-11-18
Replace instances of `stringr::str_sub()` with base R equivalent

rich-iannone:
This replaces all instances of `stringr::str_sub()` with the base R equivalent `str_substitute()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -191,6 +191,7 @@ Collate:
     'template.R'
     'utils-conversion.R'
     'utils-rd2html.R'
+    'utils-string.R'
     'utils-sweave.R'
     'utils-upload.R'
     'utils-vignettes.R'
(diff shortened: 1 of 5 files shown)


yihui (approved):
Perhaps you borrowed `str_substitute()` from your own packages, but we don't really need a sophisticated function here. The extract function `stringr::str_sub()` can be directly replaced by `base::substr()`, and the replacement function `str_sub<-` is a little trickier, since `substr<-` is not vectorized and has other quirks. Replacing parts of a string doesn't seem to be a complicated task, so I wrote the `str_replace()` function in `utils-string.R`, which seems to work (has passed the tests).

Thanks!

rich-iannone:
Wow, thank you for taking this in another direction and right to the finish line! I was going to make the `str_substitute<-` function this weekend but now I don't have to. Alright! Onto the next one.

PR #2196, opened 2022-11-19
Add labels to the default progress bar and allow users to provide a custom progress bar

### LINKED ISSUE #1880: Rethink the progress bar

yihui:
Users can customize the progress bar via `options(knitr.progress.fun = function(total, labels) {})`. The function should create a progress bar and return two methods, `update()` and `done()`. `txt_pb()` is the default.

- [ ] Document the `knitr.progress.fun` function (`total` is the total number of chunks; `labels` is the vector of chunk labels---for text chunks, the labels are empty strings).

An example of using the **cli** progress bar:

```r
options(knitr.progress.fun = function(total, labels) {
  id = cli::cli_progress_bar(total = total, .auto_close = FALSE)
  list(
    update = function(i) {
      cli::cli_progress_update(id = id)
    },
    done = function() {
      cli::cli_process_done(id)
    }
  )
})
```

An example of **progressr** (not working---I'm not familiar with this package and couldn't figure out how to make it work):

```r
options(knitr.progress.fun = function(total, labels) {
  p = progressr::progressor(total, auto_finish = FALSE, enable = TRUE)
  list(
    update = function(i) {
      p(message = sprintf('chunk: %s', labels[i]))
    },
    done = function() {
      p(type = 'finish')
    }
  )
})
```

To disable the progress bar:

```r
options(knitr.progress.fun = FALSE)
```

Note that the global option `knitr.progress.fun` has to be set before `knitr::knit()` starts. That usually means this option needs to be set in `.Rprofile`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # CHANGES IN knitr VERSION 1.42
 
+## NEW FEATURES
+
+- Users can specify a custom progress bar for `knit()` now. The default is still a text progress bar created from `utils::txtProgressBar()`. To specify a custom progress bar, set `options(knitr.progress.fun = function(total, labels) {})`. This function should take arguments `total` (the total number of chunks) and `labels` (the vector of chunk labels), create a progress bar, and return a list of two methods: `list(update = function(i) {}, done = function() {})`. The `update()` method takes `i` (index of the current chunk) as the input and updates the progress bar. The `done()` method closes the progress bar. See https://yihui.org/knitr/options/#global-r-options for documentation and examples.
+
+- The default text progress bar is still written to `stdout()` by default, but can also be written to other connections or `stderr()` now. To do so, set `options(knitr.progress.output = )` to a connection or `stderr()`.
+
 ## MAJOR CHANGES
 
 - `knit()` no longer prints out chunk options beneath the text progress bar while knitting a document (thanks, @hadley, #1880).
--- a/R/output.R
+++ b/R/output.R
@@ -289,8 +289,9 @@ process_file = function(text, output) {
     labels = unlist(lapply(groups, function(g) {
       if (is.list(g$params)) g[[c('params', 'label')]] else ''
     }))
-    pb = txtProgressBar(0, n, char = '.', style = 3)
-    on.exit(close(pb), add = TRUE)
+    pb_fun = getOption('knitr.progress.fun', txt_pb)
+    pb = pb_fun(n, labels)
+    on.exit(pb$done(), add = TRUE)
   }
   wd = getwd()
   for (i in 1:n) {
@@ -302,7 +303,7 @@ process_file = function(text, output) {
       }
       break  # must have called knit_exit(), so exit early
     }
-    if (progress) setTxtProgressBar(pb, i)
+    if (progress) pb$update(i)
     group = groups[[i]]
     res[i] = withCallingHandlers(
       if (tangle) process_tangle(group) else process_group(group),
--- a/R/output.R
+++ b/R/output.R
@@ -290,8 +290,8 @@ process_file = function(text, output) {
       if (is.list(g$params)) g[[c('params', 'label')]] else ''
     }))
     pb_fun = getOption('knitr.progress.fun', txt_pb)
-    pb = pb_fun(n, labels)
-    on.exit(pb$done(), add = TRUE)
+    pb = if (is.function(pb_fun)) pb_fun(n, labels)
+    on.exit(if (!is.null(pb)) pb$done(), add = TRUE)
   }
   wd = getwd()
   for (i in 1:n) {
@@ -303,7 +303,7 @@ process_file = function(text, output) {
       }
       break  # must have called knit_exit(), so exit early
     }
-    if (progress) pb$update(i)
+    if (progress && !is.null(pb)) pb$update(i)
     group = groups[[i]]
     res[i] = withCallingHandlers(
       if (tangle) process_tangle(group) else process_group(group),
--- a/R/parser.R
+++ b/R/parser.R
@@ -333,7 +333,6 @@ get_option_comment = function(engine) {
 
 print.block = function(x, ...) {
   params = x$params
-  cat(' chunk:', params$label, '\n')
   if (opts_knit$get('verbose')) {
     code = knit_code$get(params$label)
     if (length(code) && !is_blank(code)) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -1096,15 +1096,18 @@ txt_pb = function(total, labels) {
   # progress bar in a previous step could be completely wiped (by spaces)
   s = paste0(s, strrep(' ', n - w))
   w2 = getOption('width')
-  pb = txtProgressBar(0, total, 0, '.', width = max(w2 - 10 - n, 10), style = 3)
+  con = getOption('knitr.progress.output', '')
+  pb = txtProgressBar(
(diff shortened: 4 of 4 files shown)


rich-iannone (approved):
Looks great!

hadley:
This seems like a nice improvement and it will make a lot of people happy. I think there are two small things that you might consider.

Firstly, I was a little surprised that `labels` was passed in the constructor, rather than being supply to the `update` function, e.g.

```R
options(knitr.progress.fun = function(total) {
  id = cli::cli_progress_bar(total = total, .auto_close = FALSE)
  list(
    update = function(i, label = NULL) {
      cli::cli_progress_update(id = id)
    },
    done = function() {
      cli::cli_process_done(id)
    }
  )
})
```

And I think it would be nice if when the label was empty you automatically generated the standard "unnamed chunk {i}" label, to avoid every progress bar implementer having to do that themselves.

---

Secondly, you might also consider bundling two "default" progress bars with knitr, one that's used when `stdout` has been directed to a file (since `txtProgressBar` documents ugly output in this case). I think it could be something as simple as:
 
```R
batch_pb = function(total, labels) {
   s = ifelse(labels == '', '', sprintf(' (%s)', labels))
   list(
     update = function(i) {
       cat(s[i], "\n")
     },
     done = function() {
     }
   )
 }
 ```
 
I don't know how you tell if `stdout`  has been redirected, but cli presumably has some function that you could inline.

**Comment by HenrikBengtsson** [comment]
Excellent. For **[progressr](https://progressr.futureverse.org/)**, the following works:

```r
options(knitr.progress.fun = function(total, labels) {
  ## When creating a progressor it registers an on.exit() that
  ## shuts it downw automatically when exiting the function.
  ## To prevent this, we need need to use on_exit = FALSE.
  p <- progressr::progressor(total, auto_finish = FALSE, on_exit = FALSE)
  
  list(
    update = function(i) {
      p(message = sprintf('chunk: %s', labels[i]))
    },
    done = function() {
      p(type = 'finish')
    }
  )
})
```

That'll signal `progress` conditions.  For a user to "subscribe" to them, use:

```r
progressr::handlers(global = TRUE)
```

To tweak the progress reporter, to say, **progress**, use:

```r
progressr::handlers("progress")
```

A reporter for **cli** is still on the to-do list (https://github.com/HenrikBengtsson/progressr/issues/123)

I'll see if I can add something like

```r
options(knitr.progress.fun = progressr::make_knitr_progress_fcn())
```

with options to configure the `undo` and the `done` steps (and possible also the "init" step).

to **progressr**.

HenrikBengtsson:
> I don't know how you tell if stdout has been redirected ...

Maybe I misunderstand the comment, but shouldn't all progress output to the terminal just be sent to stderr, e.g. `cat(file = stderr(), ...)`, `pb <- utils::txtProgressBar(..., file = stderr())`, ...   There should be no R functions or packages capturing the stderr (if they do, I think it's a mistake, because it's not safe).

hadley:
@HenrikBengtsson I mean outside of R; i.e. you're not rendering interactively but are running `rmarkdown::render()` or similar from a batch job with output redirected to a log file.

yihui:
Thanks @hadley and @HenrikBengtsson!

> I was a little surprised that `labels` was passed in the constructor, rather than being supply to the `update` function

Passing to the constructor gives users a little more freedom, e.g., they can calculate the maximum width of all labels to pad enough spaces to shorter labels (this is what I did in the default progress function). The `update()` function can easily get the current label from `labels[i]`. In the **cli** example, I was unable to figure out how to print the label. I tried

```r
cli::cli_progress_update(id = id, status = labels[i])
```

but it didn't show the labels. Perhaps I need to use `type = "custom"` and a custom `format` string with `extra` data? I don't know.

> I think it would be nice if when the label was empty you automatically generated the standard "unnamed chunk {i}" label, to avoid every progress bar implementer having to do that themselves

They don't need to do that---`labels` already have `unnamed-chunk-i` for unnamed chunks. The label is empty only for text chunks.

>  one that's used when `stdout` has been directed to a file

I'll see if it's possible to check whether `stdout` has been redirected.

hadley:
@yihui ok, that makes sense. What's a text chunk? The non-code contents of an rmd?

yihui:
Yes (prose/narratives).

yihui:
> since `txtProgressBar` documents ugly output in this case

I'm not entirely sure what they meant by "ugly output". I tried the `txtProgressBar()` on a file connection, and I got multiple lines of progress bars, which is what I expected (`\r` won't work in this case so the single-line progress bar can't be redrawn on the same line).

It doesn't seem to be possible to tell if the `stdout` has been redirected. For **cli**, I think it writes the progress bar to `stderr`. If I try to capture it with `sink(type = 'message')`, I get multiple lines of progress bars in the output file (also as I expected).

For now, I'll just leave this issue behind until someone reports it and requests changes.

yihui:
BTW, `options(knitr.progress.output = stderr())` can send the progress to `stderr`.

PR #2200, opened 2022-11-24
Allow concordances to be embedded in Markdown output 

dmurdoch:
I am working on a package that will translate locations for error messages about HTML output into locations in the original Rmd file (see https://github.com/dmurdoch/RmdConcord).  With current `knitr`, it can only convert locations to the intermediate Markdown file; this patch will allow `knitr` to embed enough information in the Markdown intermediate file so that the message can refer back to the original source.

A possible variation on this patch is to remove the format check within `concord_mode` completely, i.e. to write it as

    # do not consider child mode for concordance
    concord_mode = function() {
      opts_knit$get('concordance') && !child_mode() 
    }

but I think the checks still need to be kept in `concord_gen`, so they may as well be here.


dmurdoch:
This is insufficient to handle Rmd -> pdf conversions.  I'll improve it and resubmit.

dmurdoch:
The test for concordances needed to be generalized to handle PDF or LaTeX output.

yihui (approved):
Looks good to me! Thanks!

> A possible variation on this patch is to remove the format check within `concord_mode` completely, i.e. to write it as

That sounds okay to me. If you want to do it, I'll merge this PR after you are done, otherwise I can also merge it as is.

PR #2202, opened 2022-12-13
Replace several stringr-based function calls with base equivalents

rich-iannone:
This PR replaces several uses of stringr functions with base equivalents. This is being done in order to eventually remove the dependency on the stringr package.


yihui (approved):
Thanks! Now we have four more functions left: `str_locate()`, `str_locate_all()`, `str_match_all()`, and `str_extract_all()`.

PR #2205, opened 2022-12-22
Replace several stringr fns with base equivalents

rich-iannone:
This PR replaces several uses of stringr functions with base equivalents. This is being done to remove the dependency on the stringr package. The stringr functions to be replaced are `str_locate()`, `str_locate_all()`, `str_match_all()`, and `str_extract_all()`.

Fixes: https://github.com/yihui/knitr/issues/1549


yihui (approved):
Thanks! I'll take it from here.

PR #2210, opened 2023-01-24
Stata engine - more specific cleanup of files

Hemken:
I would suggest the changes here, changing a few "." wildcards to "\\." literals dots.  This was code a wrote years ago, when I was new to regex!

An alternative would be to simply drop this Stata output file cleanup altogether, and allow it to be handled in the Statamarkdown package.


cderv:
> An alternative would be to simply drop this Stata output file cleanup altogether, and allow it to be handled in the Statamarkdown package.

Thanks I am happy to merge this as is, and we could then adapt the support of Stata here by the **Statamarkdown**. Do we need to do anything specific for this ?

cderv (approved):

Hemken:
Nothing else need be done.

Issue #2172, opened 2022-09-20
Difficulty in using `class.output` option with `collapse=T`


benwhalley:
This reopens this issue: https://github.com/yihui/knitr/issues/1906

These screenshots illustrate the problem. The only thing that changes is the value for `collapse` chunk option:

### With `collapse = F`

<img width="1081" alt="Screenshot 2022-09-20 at 14 35 28" src="https://user-images.githubusercontent.com/524696/191272423-816bcd78-8ffd-4fae-bd02-3c8dea401f0c.png">

### With `collapse = T`
<img width="1177" alt="Screenshot 2022-09-20 at 14 35 11" src="https://user-images.githubusercontent.com/524696/191272440-1da39543-ba78-43a5-843f-62a936a10d29.png">

-------------------

Minimal example is here: https://gist.github.com/benwhalley/99b2ebf4b58605e0cec35727793190f6

I can see there was some previous discussion about expected behaviour, but the problem problem is that if collapse=T then no distinction is made between the R code being echoed and the resulting output (and no different css is applied, so results can't be styled differently to commands).

I can see that for the second code chunk it's not obvious which classes to output, but at the very least allowing the author to know that something is a result rather than the command that produced it would be very useful. 

In my case, I could just set collapse=F but this results in very fragmented commands and output which is less good for explanatory purposes.

cderv:
The purpose of `collapse = TRUE` is to collapse all the source and output blocks from one code chunk into a single bloc. 
This means on HTML side it will be in one `<pre><code>` block by design. This will also mean that syntax highlighting will apply on the all block usually. 

> the problem problem is that if collapse=T then no distinction is made between the R code being echoed and the resulting output (and no different css is applied, so results can't be styled differently to commands).

Usually when setting `collapse = TRUE` as a global option, one want also to set for example the `comment` option to help differentiate code and result as you explain. 

````r
knitr::opts_chunk$set(
  collapse = TRUE,
  comment = "#>"
)
````
![image](https://user-images.githubusercontent.com/6791940/191460199-3e5ec359-b59e-4ade-b37d-bfafd0c8410a.png)

See for example [R4DS](https://r4ds.had.co.nz/tibbles.html#creating-tibbles) book.

A prompt could be added for R code line but this is not good for highlighting
````
knitr::opts_chunk$set(
  collapse = TRUE,
  prompt = TRUE
)
````
[...]

PR #2211, opened 2023-01-26
Split up chunks with collapse = TRUE if they contain htmlwidgets output

dmurdoch:
This addresses my last comment in #2172, where htmlwidgets output stopped later code from getting proper formatting.


dmurdoch:
This is very slow on some real examples.  I'll see if I can improve it.

PR #2235, opened 2023-02-27
Use option or environment variable for imgur_upload.

### LINKED ISSUE #2233: imgur_upload option?
Suggestion for improvement:

`?knitr::imgur_upload` says

> Please register your own Imgur application to get your client ID; you can certainly use mine, but this ID is in the public domain so everyone has access to all images associated to it.

This would be much more feasible if `imgur_upload()` consulted an environment variable (and/or option) for the key.

Right now, when `imgur_upload()` is used inside `reprex::reprex()`, there's no way to pass a key.

I could PR this if you agree, but I'm not sure about testing nor whether there's already a pattern for this sort of thing in knitr.

jonthegeek:
Closes #2233.

I implemented it like xfun::tinify(), but I'm not sure if it makes sense to involve opts_knit here.


cderv:
Not sure about the CI error, seems like something is happening with **pak**. Maybe transient... 

Regarding `opts_knit`, this would still be valid and it is the current way of providing a key. **reprex** could have leveraged that for example. See in https://github.com/tidyverse/reprex/blob/77ec1f612ea3d94534d6b37336a45f2d2957f3b2/R/reprex_document.R#L73-L77

**reprex** define the use of `imgur` and could have offer a way to configure this using reprex options and the `key` argument. 

However, I find it a good idea to get option or/and env var from the knitr function directly. 

I let @yihui do the review for his thoughts.

yihui (approved):
Thanks!

PR #2249, opened 2023-04-11
When caching sql engine, default to character(0) when no output.var

### LINKED ISSUE #1842: Caching sql chunks throwing error when output.var not used
I'm having trouble knitting an Rmd when I set `cache=TRUE` for sql code chunks. 

There is some discussion of a related error https://github.com/yihui/knitr/pull/1544 and https://github.com/rstudio/rmarkdown/issues/914, but those examples seem to be when `output.var` is set in the sql chunk and the bug was fixed a couple years ago.

The following Rmd will knit the first time, but the second time when it tries to read from the cache, I get this error:

```
Quitting from lines 12-13 (test-sql-cache-rmd.Rmd) 
Error in fun(environment()) : invalid first argument
Calls: <Anonymous> ... call_block -> <Anonymous> -> lazyLoad -> lazyLoadDBexec -> fun
Execution halted
```

When `cache=FALSE`, the error goes away. 

````
---
title: "sql chunk"
output: html_document
---

```{r setup, include=FALSE}
con <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
knitr::opts_chunk$set(echo = TRUE, connection = "con")
```

```{sql select_1, cache=TRUE}
SELECT 1
```
````
This issue was previously posed on [RStudio Community](https://community.rstudio.com/t/caching-sql-chunks-using-knitr-r-markdown-throwing-error/63942).
```
> xfun::session_info("knitr")
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17134), RStudio 1.3.878

Locale:
  LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
  LC_NUMERIC=C                           LC_TIME=English_United States.1252    

Package version:
  evaluate_0.14   glue_1.4.0      graphics_3.6.1  grDevices_3.6.1 highr_0.8       knitr_1.28.5    magrittr_1.5    markdown_1.1    methods_3.6.1  
  mime_0.9        stats_3.6.1     stringi_1.4.6   stringr_1.4.0   tools_3.6.1     utils_3.6.1     xfun_0.13       yaml_2.2.1     
```

---

### PR DESCRIPTION
**Pull request by cderv**
This fixes #1842

Small fix as it seems this is `NULL` which is causing with the lazy load db. 

If we want broader change, we can make sure that `cache_save` change `NULL` in `keys` to `character(0)` or do nothing if only `NULL` in `keys`

https://github.com/yihui/knitr/blob/039b6cfa45dea4ece31f2bcb65ed3b34b1330daa/R/cache.R#L17-L35

Or something different if you have idea. 

Though this PR fix is easy enough and should not create more issues.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -159,7 +159,8 @@ block_exec = function(options) {
     cache.exists = cache$exists(options$hash, options$cache.lazy)
     if (options$cache.rebuild || !cache.exists) block_cache(options, output, switch(
       options$engine,
-      'stan' = options$output.var, 'sql' = options$output.var, character(0)
+      'stan' = , 'sql' = options$output.var %n% character(0),
+      character(0)
     ))
   }
   if (options$include) output else ''

### CODE REVIEW

**Comment by yihui** [comment]
FYI the knitr-examples issue should be fixed now: https://github.com/yihui/knitr-examples/commit/d76fe69b917a3458c430515945816c16de1fe73b

**Comment by cderv** [comment]
> Let's apply a broader change instead and cache the objects only when length(keys) > 0. Thanks!

One "tricky" thing in our cache management is that we rely on the assumption that if `cache.lazy = TRUE` (the default), then we expect a lazyload db to exist (in addition to `.RData` file). We check that for `cache_exists()` . If we do not create the lazydb files when `length(keys) > 0` in `cache_save`, then we need to modify `cache_exists` to handle this assumption. Especially here
https://github.com/yihui/knitr/blob/039b6cfa45dea4ece31f2bcb65ed3b34b1330daa/R/block.R#L96-L104
or 
https://github.com/yihui/knitr/blob/d90b0161ddc298b49ec183ac481f4101ad46d8ce/R/block.R#L250-L256

We still need to consider the cache to exists when `.RData` is here, so that `cache$output()` is correctly applied. 

We could adapt `cache_exists` to be less strict about `rdb` or `rdx` file existance but this a broad change for sure. I wanted to confirm before going this road (I started to look at how to adapt though). If you want to screenshare to discuss, ping me.

**Comment by yihui** [comment]
Okay, let's coerce `keys` to character then, i.e., use `as.character(keys)`. Thanks!

**Comment by cderv** [comment]
@yihui this was indeed a very good fix. I wasn't sure what `keys` could be really, but now it makes sense, and this is a better fix than mine. 

Thanks

File: R/cache.R [hunk 1]
```diff
@@ -31,7 +31,7 @@ new_cache = function() {
     if (!lazy) return()  # everything has been saved; no need to make lazy db
     # random seed is always load()ed
     keys = setdiff(keys, '.Random.seed')
```
yihui, line 33:
```suggestion
    keys = as.character(setdiff(keys, '.Random.seed'))
```

R/cache.R
```diff
@@ -31,7 +31,7 @@ new_cache = function() {
...
-    getFromNamespace('makeLazyLoadDB', 'tools')(knit_global(), path, variables = keys)
+    getFromNamespace('makeLazyLoadDB', 'tools')(knit_global(), path, variables = as.character(keys))
```
yihui, line 34:
```suggestion
    getFromNamespace('makeLazyLoadDB', 'tools')(knit_global(), path, variables = keys)
```

yihui (approved):
Great. Thanks!

PR #2260, opened 2023-05-25
Add `fig.asp = NULL` as default in `opts_chunks`

cderv:
option was added a while back in 464d4d0d0fe605dcc82bc3c5a8feb277e5f416ad

Related to https://github.com/quarto-dev/quarto-cli/pull/5670 by @mcanouil as Quarto should have probably known about this option from `opts_chunk`

@yihui any reason not to add it ? 

Thanks!

### DIFF
--- a/R/defaults.R
+++ b/R/defaults.R
@@ -83,7 +83,7 @@ opts_chunk = new_defaults(list(
   fig.keep = 'high', fig.show = 'asis', fig.align = 'default', fig.path = 'figure/',
   dev = NULL, dev.args = NULL, dpi = 72, fig.ext = NULL, fig.width = 7, fig.height = 7,
   fig.env = 'figure', fig.cap = NULL, fig.scap = NULL, fig.lp = 'fig:', fig.subcap = NULL,
-  fig.pos = '', out.width = NULL, out.height = NULL, out.extra = NULL, fig.retina = 1,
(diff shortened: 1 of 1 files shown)


yihui (commented):
Yes, the reason not to add it into `opts_chunk` is that it will invalidate existing cache, because cache depends on chunk options. This is one thing that I regret very much in **knitr** since it makes it hard to add new chunk options into `opts_chunk`.

RStudio uses `opts_chunk_attr` to recognize chunk options, e.g.,
https://github.com/yihui/knitr/blob/8403ca57946713bb532d340eed766a7e731125a3/R/defaults.R#L123

but even `opts_chunk_attr` doesn't contain all available chunk options. Changing this object is risk-free, though.

cderv:
> Yes, the reason not to add it into opts_chunk is that it will invalidate existing cache, because cache depends on chunk options. This is one thing that I regret very much in knitr since it makes it hard to add new chunk options into opts_chunk.

@mcanouil I knew there was a reason not every option was in `opts_chunk` ! I couldn't remember why exactly. 

The initial trigger of this is that Quarto uses `names(knitr::opts_chunk$get())` to know which options are knitr options. And then add others manually. 

We could probably do the same as RStudio IDE object but for Quarto, or use the same by providing all options into `opts_chunk_attr` and use this

PR #2263, opened 2023-06-19
Use named extract for code content

cderv:
This is to avoid partial matching with option like `code_folding = FALSE`

Found in **distill** while it was warning for non empty chunk content while using
````markdown
```{r, child="mychild.Rmd"}
```
````

Try this without this PR to get the warning

````markdown
---
title: "reprex"
---

```{r}
knitr::opts_chunk$set(code_folding = FALSE)
```

```{r, child="test2.Rmd"}
```

````

````
Warning message:
In call_block(block) :
  The chunk 'unnamed-chunk-2' has the 'child' option, and this code chunk must be empty. Its code will be ignored.
````

This PR makes sure that `params[['code']]` does not do partial matching with `params$code`

it creates an issue with distill because `code_fold = FALSE` is an option used in the distill format,.

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -55,7 +55,7 @@ call_block = function(block) {
 
   if (opts_knit$get('progress')) print(block)
 
-  params$code = parse_chunk(params$code) # parse sub-chunk references
+  params[['code']] = parse_chunk(params[['code']]) # parse sub-chunk references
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
Thanks!

PR #2264, opened 2023-06-20
Fix URL handling in write_bib():

dmurdoch:
- add `packageURL` argument.  If TRUE (the default), use the DESCRIPTION file URL in preference to what `citation()` would use.
  - if multiple URLs are present, drop all but the first


yihui (approved):
Thank you!

PR #2282, opened 2023-08-25
Convert dashes in chunk option names to dots

yihui:
Fix quarto-dev/quarto-cli#5852.

@cderv Instead of enumerating specific options, I'm just substituting all dashes to dots. The advantage is that we don't need to maintain a list of hard-coded option names (e.g., if we add new chunk options to **knitr** in future, we won't need to update this list). The downside is that users might really want to use dashes in option names, but I think this should be rare.

Also fix quarto-dev/quarto-cli#6576.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -252,6 +252,13 @@ fix_options = function(options) {
   options[gsub('-', '.', dashes)] = options[dashes]
   options[dashes] = NULL
 
+  # normalize aliases
+  aliases = c(fig.dev = 'dev', fig.dpi = 'dpi')
+  for (j in intersect(names(options), names(aliases))) {
+    options[[aliases[j]]] = options[[j]]
+    options[[j]] = NULL
+  }
+
   # if you want to use subfloats, fig.show must be 'hold'
   if (length(options$fig.subcap)) options$fig.show = 'hold'
   # if the animation hook has been set, fig.show must be 'animate'
--- a/R/utils.R
+++ b/R/utils.R
@@ -253,7 +253,7 @@ fix_options = function(options) {
   options[dashes] = NULL
 
   # normalize aliases
-  aliases = c(fig.dev = 'dev', fig.dpi = 'dpi')
+  aliases = c(fig.format = 'dev', fig.dpi = 'dpi')
   for (j in intersect(names(options), names(aliases))) {
     options[[aliases[j]]] = options[[j]]
     options[[j]] = NULL
--- a/R/utils.R
+++ b/R/utils.R
@@ -247,6 +247,10 @@ tikz_dict = function(path) {
 # but now also place to tweak default options
 fix_options = function(options) {
   options = as.strict_list(options)
+  # convert dashes in option names with dots
+  dashes = grep('-', names(options), value = TRUE)
+  options[gsub('-', '.', dashes)] = options[dashes]
+  options[dashes] = NULL
 
   # if you want to use subfloats, fig.show must be 'hold'
   if (length(options$fig.subcap)) options$fig.show = 'hold'
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
-Version: 1.43.6
+Version: 1.43.7
 Authors@R: c(
     person("Yihui", "Xie", role = c("aut", "cre"), email = "xie@yihui.name", comment = c(ORCID = "0000-0003-0645-5666")),
     person("Abhraneel", "Sarma", role = "ctb"),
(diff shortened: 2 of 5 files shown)


cderv (approved):
Oh so you do that in general `fix_options` 🤔 

I was thinking to keep that YAML parsing specific, in the partition function. 

Like 
* Parsing YAML options - those could be dashed version probably 
* Substitute - by . 
* Use `knitr::merge_list()` to merge so options parsed from YAML with any other existing, including the knitr default one. 

Doing in `fix_options()` is more general. Could be ok if you think so. 
Only thoughs are: 
* Indeed, if someone is already using knitr option with `-`, it will break (probably rare as you think)
* I did the equivalent of that in my branch on Quarto side 
````
  dashes = grep('-', names(options), value = TRUE)
  options[gsub('-', '.', dashes)] = options[dashes]
  options[dashes] = NULL
```` 
and I though this could may have side effect considering order if some `.` options are after their `-` version in `options` - Something related to matter of order when duplicated and which should be taken first. 

But I guess the situation may not happen. 

Other than those thoughts, it looks good to me.

**Comment by yihui** [comment]
> I was thinking to keep that YAML parsing specific, in the partition function.

The problem with that is users might have set global chunk options (which contain dashes) in the YAML metadata of the Quarto document. By doing the conversion in `fix_options()`, we don't need to worry where the options come from (global or local, comma-separated or YAML, etc.).

> and I though this could may have side effect considering order if some `.` options are after their `-` version in `options` - Something related to matter of order when duplicated and which should be taken first.

You fixed this problem in https://github.com/quarto-dev/quarto-cli/commit/0af9c8a1a1ba1be8e2bb02546e25d5ab44946d4e. I'm using the same fix. Previous in Quarto the list of chunk options could contain duplicated options like `list(tidy.opts = ..., tidy.opts = ...)` (with the latter `tidy.opts` converted from `tidy-opts`). With your fix, this problem shouldn't occur again.

I'll let you merge this PR since I don't have the privilege to close issues in the quarto-cli repo. If you merge it, it may automatically close the two issues. Thanks!

**Comment by eitsupi** [comment]
Maybe closing #2214?

File: R/utils.R [hunk 1]
```diff
@@ -247,6 +247,17 @@ tikz_dict = function(path) {
 # but now also place to tweak default options
 fix_options = function(options) {
   options = as.strict_list(options)
+  # convert dashes in option names with dots
+  dashes = grep('-', names(options), value = TRUE)
+  options[gsub('-', '.', dashes)] = options[dashes]
+  options[dashes] = NULL
+
+  # normalize aliases
+  aliases = c(fig.dev = 'dev', fig.dpi = 'dpi')
```
cderv, line 256: In quarto the alias is in fact `fig-format`

```suggestion
  aliases = c(fig.format = 'dev', fig.dpi = 'dpi')
```

cderv:
@yihui this PR is breaking Quarto right now. 

I think this all lies in your original comment 
> **Instead of enumerating specific options**, I'm just substituting all dashes to dots. The advantage is that we don't need to maintain a list of hard-coded option names (e.g., if we add new chunk options to knitr in future, we won't need to update this list). The downside is that users might really want to use dashes in option names, but I think this should be rare.

Quarto is only normalizing known **knitr** options currently. It means it assumes that unknown knitr option will still have dash (`-`) in their name. 

https://github.com/quarto-dev/quarto-cli/blob/33c9f8557c0afd35085980cc6560eddd69ca967c/src/resources/rmd/hooks.R#L270-L284

I am trying to see if this is only something of hard coded dashed version in R code to be used with dot version, but I believe this is more generic. 

I think Quarto was assuming that unknown **knitr** options would still have their dash in `options` within **knitr** scope, and only known knitr options would be normalized to their dot version. 

We should discuss this IMO.

cderv:
And also `fix_options()` happens in **knitr** after `opts_hooks` so we need to adapt Quarto code also for this PR to work because previous normalization happened in parser for `fig-` and `out-` and this is later now. This creates issue with `fig-cap` for example. 

We may need to revert this PR and synchronize it with next Quarto version to not create breakage - we have the conflict Quarto version + knitr version to deal with 🤔

yihui:
Okay, I'm glad that we started testing Quarto against the dev version of **knitr**. First, I think we should definitely normalize the dashes *before* `opts_hooks` runs. Second, we need to limit the option names to be normalized: probably only those in `unique(c(names(knitr:::opts_chunk_attr), names(knitr::opts_chunk$get())))` (if any other options need to be normalized, we can add them to `opts_chunk_attr` since this object can be freely changed but [not `opts_chunk`](https://github.com/yihui/knitr/pull/2260)).

yihui:
I just pushed another commit, and I think the problems should be mitigated. Let's see if there is anything else we missed.

cderv:
It seems to be all good on Quarto side now - at least for everything that is testing in CI. 

Two things to do with this new commit: 

* Remove / modify the unit test I added in this initial PR
* Adapt or fix the knitr-example one

I guess you are already onto this.

PR #2292, opened 2023-09-21
Restore `opts_current` after each code chunk, and also after `knit()` exits

### LINKED ISSUE #1988: Inline code chunks inheriting labels
It seems as though the label behavior for inline code chunks is inconsistent / problematic - specifically it seems that they inherit the label of the most recent proceeding code chunk, and in the case where an inline chunk occurs before any other it will have a blank label. See the sample Rmd below which demonstrates these behaviors:

``````
---
output: html_document
---

Inline - "`r knitr::opts_current$get("label")`"

# Unlabeled chunk

```{r}
knitr::opts_current$get("label")
```

Inline - "`r knitr::opts_current$get("label")`"

Inline - "`r knitr::opts_current$get("label")`"

# Label chunk

```{r label}
knitr::opts_current$get("label")
```

Inline - "`r knitr::opts_current$get("label")`"


# Different engines

```{python pychunk}
1+1
```

Inline - "`r knitr::opts_current$get("label")`"
``````

I've come across a specific [issue](https://github.com/ropensci/magick/issues/310) with this when using an inline code chunk to generate an image (via magick) as the `plot_counter()` is reset between the regular and inline chunk but since they share a label the plots from earlier regular chunk are overwritten.

I would think that the more desireable behavior would be for the chunk label to increment regardless of the chunk type, the labels could continue to use `unnamed-chunk-*` or could have a specific inline chunk label variant e.g. `inline-chunk-1`, `inline-chunk-2`, etc.

---

```r
xfun::session_info('knitr')
## R version 4.0.5 (2021-03-31)
## Platform: x86_64-apple-darwin20.3.0 (64-bit)
## Running under: macOS Big Sur 10.16, RStudio 1.4.1114
## 
## Locale: en_US.UTF-8 / en_US.UTF-8 / en_US.UTF-8 / C / en_US.UTF-8 / ## en_US.UTF-8
## 
## Package version:
##   evaluate_0.14   glue_1.4.2      graphics_4.0.5  grDevices_4.0.5
##   highr_0.9       knitr_1.32      magrittr_2.0.1  markdown_1.1   
##   methods_4.0.5   mime_0.10       stats_4.0.5     stringi_1.5.3  
##   stringr_1.4.0   tools_4.0.5     utils_4.0.5     xfun_0.22      
##   yaml_2.2.1  
```

yihui:
Will this fix #1988? (except for the support of inline labels)


cderv (commented):
Yes I believe this is what is needed. This means that Inline code won't have any label. 
We can think of something later; but this will probably make some existing code fails. 

See related magick issue: https://github.com/ropensci/magick/issues/310
It is using inline code to insert an image and rely on the fact that **knitr** will provide a name to a file when using `plot_counter()`. Previously it was using previous chunk name and overriding the previous image, but now it will have no name. 

When running the example there, I see for the inline code 
* `options$fig.path` being `figure/` instead of `test_files/figure-gfm/`. Not sure why this options is different.. 🤔 
* No label so `valid_path(options$fig.path, options$label)` is only `figure/`. Then suffix `-1` will apply

It seems `opts_chunk$get()` and `opts_current$get()` are not the same. 

Should `opts_current` in inline code be set to `opts_chunk` maybe ? Like in chunk ? 
Inline have no params, but they should probably default to the global options that could have been modified ? 
```diff
diff --git a/R/block.R b/R/block.R
index 35bbed87..12c486dc 100644
--- a/R/block.R
+++ b/R/block.R
@@ -547,6 +547,8 @@ merge_character = function(res) {
 }

 call_inline = function(block) {
+  optc = opts_current$get(); on.exit(opts_current$restore(optc), add = TRUE)
+  opts_current$restore(opts_chunk$get())
   if (opts_knit$get('progress')) print(block)
   in_input_dir(inline_exec(block))
 }
```

We could probably also deal with adding a label this way IMO 

```diff
diff --git a/R/block.R b/R/block.R
index 35bbed87..4e1c5c80 100644
--- a/R/block.R
+++ b/R/block.R
@@ -547,6 +547,9 @@ merge_character = function(res) {
 }

 call_inline = function(block) {
+  optc = opts_current$get(); on.exit(opts_current$restore(optc), add = TRUE)
+  params = opts_chunk$merge(list(label = unnamed_chunk("inline", inline_counter())))
+  opts_current$restore(params)
   if (opts_knit$get('progress')) print(block)
   in_input_dir(inline_exec(block))
 }
diff --git a/R/utils.R b/R/utils.R
index 404dd768..271dae05 100644
--- a/R/utils.R
+++ b/R/utils.R
@@ -17,6 +17,7 @@ knit_counter = function(init = 0L) {
 plot_counter = knit_counter(1L)
 shot_counter = knit_counter(1L)
 chunk_counter = knit_counter(1L)
+inline_counter = knit_counter(1L)

 # a vectorized and better version than evaluate:::line_prompt
 line_prompt = function(x, prompt = getOption('prompt'), continue = getOption('continue')) {
````

I know using Images in inline code is not the best usage - problaby **magick** could be fixed to correctly handle inlines.  

However, just ideas that you can tweak if you are interested,a s it seems setting label and global options as default for inline seems what should be to me. 







@rundel this will partially solve #1988 as discussed. 

@gadenbuie this should avoid the need to do this yourself (https://github.com/yihui/knitr/issues/1798#issuecomment-1715949630)

yihui:
I've applied your patch with a small change. Thanks!

Now inline code will have access to a unique label in `opts_current$get('label')`.

cderv:
Cool thanks ! 

I wonder how many side effect this unamed chunk scheme could cause 😅  it is possible that some user may rely on it too much. Would it invalidate existing cache too ? 

Not saying it is bad, but we could probably communicate ahead of time if this could help

yihui:
Yes, this could be a breaking change, so I've started the revdep check: https://github.com/yihui/crandalf/actions/runs/6644880330/job/18054982641 I'll see how bad this change could be, and write a blog post accordingly.

yihui:
I've reverted the change in the naming scheme, since the revdep check showed that a few packages would be broken.

Issue #2226, opened 2023-02-10
Improve YAML error thrown when parsing in-chunk options


cderv:
From discussion at https://github.com/quarto-dev/quarto-cli/issues/4306 - @yihui I agree with your comment 

````
---
title: "Example"
format: pdf
engine: knitr
editor: visual
---

```{r}
# install.packages(c("DBI", "RSQLite"))
db = DBI::dbConnect(RSQLite::SQLite(), ":memory:")
```


```{sql}
#| label: test multiqueries -5
#|  connection: db

select 1
```
````

Leading to this error 

````
processing file: my-homework.qmd
Error in yaml::yaml.load(meta, handlers = list(expr = parse_only)) : 
  Scanner error: mapping values are not allowed in this context at line 2, column 12
Calls: .main ... FUN -> parse_block -> partition_chunk -> <Anonymous>
Execution halted
````

which is not that helpful to know about the error. 

This is not specific to Qmd and would be useful to Rmd using YAML syntax too

**Comment by yihui** [comment]
Below is what I'd do and you can try to implement it:

First, we need to improve the error message from `yaml.load()`:

https://github.com/yihui/knitr/blob/77970b0717a9497b5ba275689a21c0af2e58ab07/R/parser.R#L303

This could be done in `tryCatch(..., error = function(e) ...)`. In the `error ` handler, we parse the line/column numbers from the original message (you can access from `e$message`):

```
  Scanner error: mapping values are not allowed in this context at line 2, column 12
```

Then print the original YAML text and insert `^~~~~~` beneath the problematic line, with `^` pointing to the error column. Users should see an error message like this:

```
Failed to parse YAML: 
  Scanner error: mapping values are not allowed in this context

label: test multiqueries -5
 connection: db
           ^~~~~~~~~~
```

I think this can greatly improve the error message. If we want to do better, we can point out the line numbers of the code chunk. To do that, first we need to apply this `lapply()` to `seq_along(groups)` instead of `groups`, and we will get the index of the current chunk:

https://github.com/yihui/knitr/blob/77970b0717a9497b5ba275689a21c0af2e58ab07/R/parser.R#L23

With the index, we can emit a message similar to:

https://github.com/yihui/knitr/blob/77970b0717a9497b5ba275689a21c0af2e58ab07/R/output.R#L311-L322

That means we need to factor out these lines of code so that it can be reused, i.e., we pass an index `i` to a function, and it will emit a message `Quitting from lines ...`.

[...]

PR #2294, opened 2023-09-28
Add a handler to improve YAML error message

pedropark99:
Hello! This PR seeks to fulfill the objective proposed at the issue #2226 

This PR adds a `tryCatch()` block to the step of parsing the YAML with the chunk options, and uses the `error` handler to build a better error message, as proposed in the issue #2226 . The new error message is in the following format:

```r
> knit("C:\\Users\\pedro\\Desktop\\test.Rmd")
Failed to parse YAML: 
Scanner error: mapping values are not allowed in this context at line 2, column 12


label: test multiqueries -5
 connection: db
             ^~~~~~

```

This error message was produced while knitting the following Rmd:

````md
---
title: "Example"
format: pdf
engine: knitr
editor: visual
---

```{r}
# install.packages(c("DBI", "RSQLite"))
db = DBI::dbConnect(RSQLite::SQLite(), ":memory:")
```


```{sql}
#| label: test multiqueries -5
#|  connection: db

select 1
```
````

However, this current format of the error message is still not complete. In other words, it only partially meets the full description given by @yihui in the mentioned issue. According to the full description given by @yihui , there are two groups of indeces (or indexes) that would help to improve the error message, which are:

- The line and column indexes that point to the exact position inside the YAML where the error occured;
- The line index (or label) of the code chunk that contains the YAML where the error occured;

The current format of the error message that I showed above only contains the first group of indexes (the line and column indexes inside the YAML). So I can still improve this version by including the line index (or label) of the code chunk. According to @yihui we could get the code chunk index by using a `groups` variable. I did quickly looked for a similar variable at the parent environment from the `tryCatch()` block, with `ls(envir = parent.env(environment()))`, but I did not found this variable, or, a similar global variable from which I could extract the code chunk index.

But I am more than happy to implement this second index if you can give more instructions on how to get this code chunk index/label from inside the `tryCatch()` block 😉.

Furthermore, since the `knitr` codebase tries to stick with base R as much as possible I ended up using `base::regexpr()` to extract/parse the column and line index from the original error message. But I fell that the parsing is still a bit clunk, so, I think that this part of the code can be improved.

Another thing that we can improve/change is how to print/return the improved error message. Since is an error message, I did try to print the message with a `stop()` call, however, the message was not being printed correctly. That is why, I ended up chosing to just returning the error message at the end of the error handler.

### CODE REVIEW

File: R/parser.R [hunk 1]
```diff
@@ -302,7 +302,37 @@ partition_chunk = function(engine, code) {
...
+        regex = "line (?<line>\\d+), column (?<column>\\d+)"
+        regex_result = regexpr(regex, message, perl = TRUE)
+        starts = attr(regex_result, "capture.start")
+        lengths = attr(regex_result, "capture.length")
+
+        line_index = substr(message, starts[,"line"], starts[,"line"] + lengths[,"line"] - 1)
+        column_index = substr(message, starts[,"column"], starts[,"column"] + lengths[,"column"] - 1)
+
+        line_index = as.integer(line_index)
+        column_index = as.integer(column_index)
```
yihui, line 318: You can use `regmatches()` with `regexec()`:

```suggestion
        x = e$message
        r = "line (?<row>\\d+), column (?<col>\\d+)"
        m = regmatches(x, regexec(r, x, perl = TRUE))[[1]][-1]
        row = as.integer(m['row'])
        col = as.integer(m['col'])
```

R/parser.R
```diff
@@ -302,7 +302,37 @@ partition_chunk = function(engine, code) {
...
+
+        spaces = paste(rep(" ", column_index), collapse = "")
+        cursor = paste(spaces, "^~~~~~", collapse = "")
```
yihui, line 321: Use `strrep()`:

```suggestion
        cursor = paste0(strrep(" ", col), "^~~~~~")
```

R/parser.R
```diff
@@ -302,7 +302,37 @@ partition_chunk = function(engine, code) {
...
+
+        split_indexes = seq_along(meta) <= line_index
+        before_cursor = meta[split_indexes]
+        after_cursor = meta[!split_indexes]
+        error_message = c(
+          "Failed to parse YAML: ", e$message, "\n",
+          before_cursor,
+          cursor,
+          after_cursor
+        )
```
yihui, line 331: Use `append()`:

```suggestion
        x = c("Failed to parse YAML: ", x, "\n", append(meta, cursor, row))
```

R/parser.R
```diff
@@ -302,7 +302,37 @@ partition_chunk = function(engine, code) {
...
+
+        return(paste(error_message, sep = "\n"))
```
yihui, line 333: Use `collapse` instead of `sep`:

```suggestion
        return(paste(x, collapse = "\n"))
```

yihui (changes requested):
Good job!

> According to @yihui we could get the code chunk index by using a `groups` variable. I did quickly looked for a similar variable at the parent environment from the `tryCatch()` block, with `ls(envir = parent.env(environment()))`, but I did not found this variable, or, a similar global variable from which I could extract the code chunk index.

I meant to change https://github.com/yihui/knitr/blob/77970b0717a9497b5ba275689a21c0af2e58ab07/R/parser.R#L23 to

```r
  lapply(seq_along(groups), function(i) {
    g = groups[[i]]
```

Then you got the index `i`. Pass it to https://github.com/yihui/knitr/blob/77970b0717a9497b5ba275689a21c0af2e58ab07/R/parser.R#L37

Of course you need to add a new argument to `parse_block()`. Then further pass `i` to https://github.com/yihui/knitr/blob/77970b0717a9497b5ba275689a21c0af2e58ab07/R/parser.R#L107

After you get the index in `partition_chunk()`, you can call `current_lines(i)` to get the line numbers.

If that's not clear enough, feel free to leave this task behind. I can tweak the PR by myself. Thanks!

pedropark99:
Hey @yihui ! First of all, thank you for the review!

**OBS:** There is one task in the CI pipeline that is failling because of an outdated R version which is not supported by one the `knitr` dependencies. Since I understand that the fix to this error is out of the scope of this PR, I will leave this error with you 😉.

I made the changes you requested in your review.

After that, I did a quick "debug walking" through the parser code, to get a better understanding on how it works. Then, I followed your instructions on how to get the chunk lines. With my latest changes, the error message is now on the following format:

```r
> knit("C:\\Users\\pedro\\Desktop\\test.Rmd")
In partition_chunk(engine, code, block_index) :
  Invalid YAML option format in chunk: 
Failed to parse YAML inside code chunk at lines 15-19. Scanner error: mapping values are not allowed in this context at line 2, column 7


connection: db
 label: test multiqueries -5
       ^~~~~~
```

I think that maybe the above message can still be improved if we change the last bit of the message like this:

```r
Failed to parse YAML inside code chunk at lines 15-19. Mapping values are not allowed in a code chunk context. Error found at line 2, column 7 inside the code chunk.

connection: db
 label: test multiqueries -5
       ^~~~~~
```

What is your opinion on it? If you think that this is a good addition, I can make the changes for us.

Now, in addition to `parser.R`, since I basically changed the function signatures for `partition_chunk()` and `parse_block()`, I also had to make changes in two more places:

- change all the calls to `parse_block()` inside the unit tests to include the new block index parameter.
- change the `partition_chunk()` documentation to include the new block index parameter.

The changes above led me to also recompile all functions documentation with `devtools::document()`.

yihui (approved):
Thank you! I'll finish the rest of work by myself.

> I think that maybe the above message can still be improved if we change the last bit of the message like this:

I agree, but I feel that we are hacking the message from the **yaml** package too much. Let's leave it unchanged for now, even though it's a little obscure (I think the `^~~~~` mark can offer enough guidance).

Issue #2295, opened 2023-09-28
opts_current locking in 1.44 is breaking JuliaCall


cderv:
@yihui Just to track this in order to help them solve this as for now CRAN version of Julia does not work with **knitr** 1.44 when plots. 

it came up in 
* https://github.com/quarto-dev/quarto-cli/issues/7055
after the impact of 
* https://github.com/yihui/knitr/issues/1798

**JuliaCall** is using `opts_current$set`, so they need to update but it seems they did not see problem yet. 
https://github.com/Non-Contradiction/JuliaCall/blob/c05473bea78a0197c639f7e82ab1c6f2e943e1cc/R/RMarkdown.R#L56C1-L60

It seems there is a stale maintenance question though there...
* https://github.com/Non-Contradiction/JuliaCall/issues/208

So hopefully we can help. 

- [ ] I plan to at least do a PR
- [ ] Should we consider a workaround on our side until PR is merged and new JuliaCall is on CRAN ?  Maybe follow more usual progressive change by warning and then error ? So that all package maintainer can update... 
    We could go through a basic search list to see which tool can be impacted (https://github.com/search?q=%22opts_current%24set%22+language%3AR&type=code)

yihui:
I tend to change the error to warning for now.

bschneidr:
Perhaps knitr could rely on JuliaConnectoR instead of JuliaCall?

https://cran.r-project.org/web/packages/JuliaConnectoR/index.html

JuliaCall [doesn't seem to be actively maintained](https://github.com/Non-Contradiction/JuliaCall/issues/208), whereas JuliaConnectoR is actively maintained and seems much better documented (for example, it now has a detailed [JSS paper](https://www.jstatsoft.org/article/view/v101i06)).

yihui:
@bschneidr The status of JuliaCall is indeed a little concerning, but the nice thing is this: https://github.com/Non-Contradiction/JuliaCall/blob/master/R/RMarkdown.R i.e., it has provided an engine function that can be set via `knitr::knit_engine$set()`. If JuliaConnectoR also provides such a function, I'll be happy to support it on knitr's side. Thank you!

cderv:
@bschneidr I would advise opening an issue at https://github.com/stefan-m-lenz/JuliaConnectoR for a julia engine for **knitr**. 

We can then chime in if help is needed for that. Thank you !

cderv:
Upstream has been fixed :
[...]

PR #2296, opened 2023-09-29
Use a warning instead of an error for opts_current locking

cderv:
This solves #2295 on our side by making for now the locking optional and only warn user. 

Not sure if the `immediate.` is necessary but I figure it would be better that this is seen and not hidden in all aggregated warning. 

I also added a message about the warning for now to become an error. maybe not the best wording... feel free to tweak !


yihui (approved):
Thanks!

PR #2300, opened 2023-10-20
Improve convert_chunk_header for YAML options

cderv:
* Use `-` and not `.` for options name
* Use the new aliases for `dev` and `dpi` 

This also fix an issue with numeric conversion because `yaml::as_yaml` to consider a precision that we don't want here I believe.


yihui (approved):
Thanks! I'll merge `dash_names()` with `dot_names()` later.

R/parser.R
```diff
@@ -862,6 +864,14 @@ convert_chunk_header = function(
             x = tolower(x)
             class(x) = 'verbatim'
             x
+          },
+          # use character with verbatim for no quotes
+          # so that numeric as kept unchanged (without changing precision)
+          # fig.width = 10, should not be fig-width: 10.0
+          numeric = function(x) {
+            x = as.character(x)
```
yihui, line 872: The number might be represented in scientific notation, e.g., 1e10, which won't work for YAML.

yihui (approved):
No worries! Code abstraction takes time. I just did it.

.github/workflows/R-CMD-check.yaml
```diff
@@ -57,8 +57,8 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           extra-packages: |
-            any::XML
             any::rcmdcheck
+            XML=?ignore-before-r=4.0.0
```
yihui, line 61: Not sure why, but this doesn't seem to work: https://github.com/yihui/knitr/actions/runs/6591830453/job/17911237979

yihui (approved):

.github/workflows/R-CMD-check.yaml
cderv: I think it probably requires to indicate the highest deps that requires it, which in our case is `gridSVG`. I tried a commit with this

cderv:
Thanks a lot for the fixes !

PR #2320, opened 2024-01-29
Improve support for knitr::spin(format = 'qmd')

kylebutts:
The approach taken here had the goal of not touching the original spin code, so that this only impacts `format = 'qmd'` and did so as simply as possible.

To do this, I find the any code chunk that *starts* with `#|` and prepend a `# %%` for the previous line. This prevents converting each option passed with `#|` from creating a new chunk.

Perhaps this could be integrated more deeply with how knitr::spin works, but this is the simplest solution.

@cderv

### DIFF
--- a/R/spin.R
+++ b/R/spin.R
@@ -3,8 +3,7 @@
 #' This function takes a specially formatted R script and converts it to a
 #' literate programming document. By default normal text (documentation) should
 #' be written after the roxygen comment (\code{#'}) and code chunk options are
-#' written after \code{#+} or \code{# \%\%} or \code{#-} or \code{# ----} or
-#' any of these combinations replacing \code{#} with \code{--}.
+#' written after \code{#|} or \code{#+} or \code{# \%\%} or \code{# ----}.
 #'
 #' Obviously the goat's hair is the original R script, and the wool is the
 #' literate programming document (ready to be knitted).
@@ -90,7 +89,7 @@ spin = function(
       block = strip_white(block) # rm white lines in beginning and end
       if (!length(block)) next
 
-      rc <- '^(#|--)+(\\+|-|\\s+%%| ----+| @knitr)'
+      rc = '^(#|--)+(\\+|-|\\s+%%| ----+| @knitr)'
       opt = grep(rc, block)
       # pipe comments (#|) should start a code chunk if they are not preceded by
       # chunk opening tokens
--- a/R/spin.R
+++ b/R/spin.R
@@ -159,42 +159,41 @@ spin = function(
   c(paste0(b, '{r'), '}', b, paste0(i, 'r \\1 ', i))
 }
 
-
+# find the position of the starting `#|` in a consecutive block of `#|` comments
+comment_start = function(x) {
+  i = startsWith(x, '#| ')
+  r = rle(i)
+  l = r$lengths
+  j = cumsum(l) - l + 1
+  j[r$values]
+}
 process_block_for_qmd <- function(block) {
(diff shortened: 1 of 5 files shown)


kylebutts:
Example:
``` r
library(devtools)
load_all()

spin_w_tempfile = function(..., format = "Rmd") {
  tmp = tempfile(fileext = ".R")
  writeLines(c(...), tmp)
  spinned = spin(tmp, knit = FALSE, format = format)
  result = readLines(spinned)
  file.remove(c(tmp, spinned))
  result
}

print_result = function(s) {
  cat(paste0(s, collapse = "\n"))
}

block = c(
  "# %% ",
  "#| echo: false",
  "#| message: false",
  "#| include: false",
  "1+1 ",
  "#| eval: false",
  "1+1",
  "",
  "#' # Header",
  "#' Text",
  "#| include: false",
  "1+1"
)

block |> process_block_for_qmd() |> print_result()
#> # %% 
#> #| echo: false
#> #| message: false
#> #| include: false
#> 1+1 
#> # %%
#> #| eval: false
#> 1+1
#> 
#> #' # Header
#> #' Text
#> # %%
#> #| include: false
#> 1+1
```

R/spin.R
```diff
@@ -155,6 +159,48 @@ spin = function(
...
+    } else {
+      chunk_start <- opt
+    }
+  }
+  need_chunk_start = intersect(exe_opt, chunk_start)
+  if (length(need_chunk_start)) {
+    for (i in seq_along(need_chunk_start)) { 
+      # insert position (keeping track of the additional elements we are adding)
+      j = need_chunk_start[i] + (i - 1)
+      block <- c(block[1:j-1], "# %%", block[j:length(block)])
```
yihui, line 197: You can probably [use `append()` here](https://yihui.org/en/2023/10/three-functions/#append-insert-elements-to-a-vector).

yihui (commented):
I don't have time to review this PR at the moment, but I feel the `for` loops might be unnecessary if you take advantage of `rle()`. Here is a short function (ninja code) that I just wrote for your reference:

```r
# find the position of the starting `#|` in a consecutive block of `#|` comments
comment_start = function(x) {
  i = startsWith(x, '#| ')
  r = rle(i)
  l = r$lengths
  j = cumsum(l) - l + 1
  j[r$values]
}
```

Test:

```r
block = c(
    "# %% ",
    "#| echo: false",
    "#| message: false",
    "#| include: false",
    "1+1 ",
    "#| eval: false",
    "1+1",
    "",
    "#' # Header",
    "#' Text",
    "#| include: false",
    "1+1"
)
comment_start(block)
#> [1]  2  6 11
```

Then you look back one line and see if that line starts with the `rc` pattern (such as `# %%`). If not, add `# %%`.

kylebutts:
Correct on both points @yihui! The `rle` solution is nice, thank you.

yihui (approved):
I just finished revising this PR. Thanks!

yihui:
I forgot to mention that I remove the limitation to `qmd`, i.e., this idea should work for any formats (`Rnw`, `Rhtml`, etc.). That's because the `#|` comments are not special to `qmd`. They work for all **knitr** document formats.

t-kalinowski:
Would it be possible to make the `#|` prefix customizable, or perhaps changed to `#'|` in the context of `knitr::spin()`? 

It conflicts with the way users can pass options to guildai (the experiment tracker). Guildai is no longer being developed by us, but there is other work ongoing where I'm hoping to use similar syntax in R scripts.

PR #2330, opened 2024-03-04
add ALTER as sql statment, that is not returning anything

maxschmi:
I added the SQL `ALTER` statement to commands that should only run with dbExecute and not dbFetch as there won't be any returns from an `ALTER` statement

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,7 @@
 ## MINOR CHANGES
 
 - Fixed broken vignettes, improved CSS for HTML vignettes, and reduced the file sizes.
+- SQL code chunks that run `ALTER` statements are only executed and not tried to fecth a result (thanks, @maxschmi, #2330).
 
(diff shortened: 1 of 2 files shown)


yihui (approved):
Sounds good! Could you add a news item to `NEWS.md`? Then I'll merge this PR. Thank you!

NEWS.md
```diff
@@ -31,6 +31,7 @@
 ## MINOR CHANGES
 
 - Fixed broken vignettes, improved CSS for HTML vignettes, and reduced the file sizes.
+- SQL code chunks that run `ALTER` statements are only executed and not tried to fecth a result (thanks, @maxschmi, #2330).
```
yihui, line 34:
```suggestion

- SQL code chunks that run `ALTER` statements are only executed and not tried to fecth a result (thanks, @maxschmi, #2330).
```

yihui (approved):
Great! Thanks again!

PR #2340, opened 2024-04-26
Support custom cache pack/unpack via S3 generics for non-serializable objects

### LINKED ISSUE #2176: Caching reference objects
Is there a way, or could support be added, to tell knitr how to serialize objects of a particular class before caching, and how to restore them after caching?  See [this](https://stackoverflow.com/questions/72648083/can-knitrs-cache-work-with-pointer-objects) and [this](https://stackoverflow.com/questions/73867522/use-of-terra-in-rmd-with-cache-t) questions on stackoverflow for more context 

The "terra" package uses S4 objects with one slot that has a reference to a C++ object (via a Rcpp module). These S4 objects can be serialized and unserialized with very little effort. For example, one can do 

```
library(terra)
f <- system.file("ex/elev.tif", package="terra")
r <- terra::rast(f)  # S4 object that has a reference class
x <- serialize(r, NULL) 
# recreate the object 
s <- terra::rast( unserialize(x) )  
s
#class       : SpatRaster 
#dimensions  : 90, 95, 1  (nrow, ncol, nlyr)
#resolution  : 0.008333333, 0.008333333  (x, y)
#extent      : 5.741667, 6.533333, 49.44167, 50.19167  (xmin, xmax, ymin, ymax)
#coord. ref. : lon/lat WGS 84 (EPSG:4326) 
#source      : memory 
#name        : elevation 
#min value   :       141 
#max value   :       547 
```

But can one use that mechanism with knitr caching?  I suppose that it could be possible, in principle, to specify that objects of class "SpatRaster" should be passed to `serialize` before caching, and that `terra::rast` should be used when restoring the object. Whether this is easy to do in practice, I do not know.

---

### LINKED ISSUE #2339: Problem caching instances of torch modules and datasets
Caching chunks that create an instance of torch module or of a torch dataset yields an `external pointer is not valid` error when the instance is used in another chunk. 

Example with torch module:
        
        ```{r, cache=TRUE}
        lin <- nn_linear(2, 3)
        # torch_save(lin, "lin.pt")
        ``` 
        
        ```{r}
        # lin <- torch_load("lin.pt")
        x <- torch_randn(2)
        lin$forward(x)
        ```

Example with torch dataset:

        ```{r, cache=TRUE}
        ds_gen <- dataset(
          initialize = function() {
            self$x <- torch_tensor(1:10, dtype=torch_long())
            }, 
            .getitem = function(index) {
            self$x[index]
          },
          .length = function() {
             length(self$x)
          }
        )
        
        ds <- ds_gen()
        ``` 

        ```{r}
        ds[1:3]
        ```
If there is no cache, the chunks are executed without problems. However, when a cache exists, an error is created when trying to access the cached instance of the module or of the dataset:

     Error in cpp_tensor_dim(x$ptr) : external pointer is not valid

This might be due to the fact that R torch package relies on reference classes (R6 and/or R7) and could be related to issue #2176.  In any case, caching would be useful to cache trained instance of a module or instances of datasets which involve a lot processing during initialization. 

At the moment, the only alternative is to save the torch model in the cached chunk with `torch_save`  and load it in the uncached chunk with `torch_load` (see comments in the chunk above).  However, afaik, there is no method to save and load torch datasets.

atusy:
This PR allows implementing `knit_cache_hook` methods which may preprocess objects (e.g., save to an external file) and define custom loaders.

I will add a NEWS item after we agree with the design.

- refactor(cache): use saveRDS/readRDS instead of makeLazyLoadDB/lazyload
    - For migration, `cache_save` replaces rdb/rdx files with rds file
    - For backward compatibility, `cache_load()` attempts `lazyload()` if rdb/rdx files are available
- ~~feat(cache): allow pre/postprocessing cache objects)~~
    - `knit_cache_preprocess` preprocesses objects being saved
    - `knit_cache_postprocess` postprocesses objects being loaded 
- feat!(cache): implement knit_cache_hook instead of pre/post-processors
    - Call `knit_cache_hook` methods on saving cache
        - Methods may save extra files under `${cache_path(h)}__extra` directory
        - Methods may return custom loader functions which is saved to `${cache_path(h).rds}`

With this PR, we can add some hooks on objects to be cached.
For example, we can use writeLines to save character objects.

````
```{r}
library(knitr)
registerS3method(
  "knit_cache_hook",
  "character",
  function(x, nm, path) {
    # Cache x as is if it extends character class
    if (!identical(class(x), "character")) {
      return(x)
    }

    # Preprocess data (e.g., save data to an external file)
    # Create external files under the directory of `paste0(path, "__extra")`
    # if knitr should cleanup them on refreshing/cleaning cache
    d <- paste0(path, "__extra")
    dir.create(d, showWarnings = FALSE, recursive = TRUE)
    f <- file.path(d, paste0(nm, '.txt'))
    writeLines(x, f)

    # Return loader function
    # which receives ellipsis for future extentions and has knit_cache_loader class
    structure(function(...) readLines(f), class = 'knit_cache_loader')
  },
  envir = asNamespace("knitr")
)
```

```{r, cache=TRUE}
x <- 'foo bar'
print(x)
```

```{r}
print(x)
```
````

### CODE REVIEW

**Comment by atusy** [comment]
maybe preprocess and postprocess are not good names...

**Comment by atusy** [comment]
I got to fix tests

File: R/cache.R [hunk 1]
```diff
@@ -128,10 +142,16 @@ cache_meta_name = function(hash) sprintf('.%s_meta', hash)
 # a variable name to store the text output of code chunks
 cache_output_name = function(hash) sprintf('.%s', hash)
 
+# process cached objects before save and after read
+knit_cache_preprocess = function(x, ...) UseMethod('knit_cache_preprocess')
+knit_cache_preprocess.default = function(x, ...) x
+knit_cache_postprocess = function(x, ...) UseMethod('knit_cache_postprocess')
+knit_cache_postprocess.default = function(x, ...) x
```
atusy, line 149: Postprocess is skipped if a package is not loaded.

atusy:
To solve the above problems, I implemented the `knit_cache_hook` generic function in place of `knit_cache_preprocess` and `knit_cache_postprocess`. See updated description for the details.

yihui (commented):
I feel this implementation is too complicated, and I'd like to propose a different way: in the latest version of **xfun**, I added two functions `lazy_save()` and `lazy_load()` as a different and simple implementation of base R's `lazyLoad()` and `tools:::makeLazyLoadDB()`. By default, `xfun::lazy_save()/lazy_load()` use the rds format, ~~which should solve the problems #2176 and #2339 (I've only briefly tested #2176).~~

For backward compatibility, we can first test if `*.rdb`/`*.rdx` exist. If they do, we use the old approach (base R), otherwise we switch to **xfun**'s lazy loading.

What do you think?

---

Correction: I don't remember how I tested #2176 now. I was expecting that this would just work:

````md
```{r}
library(terra)
```

```{r, cache=TRUE}
r = rast(matrix(1:12,3,4))
```

```{r}
r
```
````

but it doesn't (even if we save `r` to `*.rds`). The object `r` will fail to load in a new R session, and we still have to do `wrap()`/`unwrap()`:

````md
```{r}
library(terra)
```

```{r, cache=TRUE}
r = rast(matrix(1:12,3,4))
p = wrap(r)
```

```{r}
unwrap(p)
```
````

That said, I had this issue in mind when designing the new cache system for **litedown**. With **litedown**, it's possible to customize the read/write methods for cache via the chunk option `cache.rw`, e.g.,

````md
---
title: "Caching terra objects with litedown"
knit: litedown:::knit
---

```{r}
library(terra)
rw_terra = list(
  name = 'terra',
  save = function(x, file) {
    if (inherits(x, 'SpatRaster')) x = wrap(x)
    saveRDS(x, file)
  },
  load = function(...) {
    x = readRDS(...)
    if (inherits(x, 'PackedSpatRaster')) x = unwrap(x)
    x
  }
)
```

```{r, cache=TRUE, cache.rw=rw_terra}
r = rast(matrix(1:12,3,4))
```

```{r}
r
```
````

**Comment by atusy** [comment]
Thanks for the comment. 
I do not have a strong opinion, but let me leave some comments below.

I accepted the complexity for following reasons:

- the feature is mainly for package developers and not for end users
- usage is limited (I guess)

With my implementation, user's do not have to care about what is going on under saving/loading caches.

For developers, I agree chunk option is a good idea.
The implementation becomes simple.
However, this imposes end-users to understand tricks for edge-cases.
Can we expect end-users read documents carefully before facing troubles on cache behavior?

**Comment by yihui** [comment]
Good points, and I agree. Let me think more about it. Thanks!

**Comment by yihui** [comment]
Reworked. The cache backend now uses `xfun::lazy_save()` / `xfun::lazy_load()` (rds) instead of `tools::makeLazyLoadDB()` / `lazyLoad()`; existing `.rdb`/`.rdx` caches still load, so it's backward-compatible.

Instead of the `knit_cache_hook` generic (which had to dispatch on the object when saving but fall back to a stored loader closure when loading), customization is now done with **two** S3 generics that both dispatch on a real object:
- `knit_cache_pack(x)` — called on each object before caching, returns a serializable version;
- `knit_cache_unpack(x)` — called after an object is read back, restores the original.

Both default to identity. The pack/unpack happen at xfun's per-object read/write layer, so **lazy loading is preserved** — an object is only unpacked when it's actually accessed. This closes #2176; packages register methods for their classes in `.onLoad()` and users configure nothing, e.g. terra:

```r
registerS3method("knit_cache_pack", "SpatRaster", function(x, ...) terra::wrap(x), envir = asNamespace("knitr"))
registerS3method("knit_cache_unpack", "PackedSpatRaster", function(x, ...) terra::unwrap(x), envir = asNamespace("knitr"))
```

@atusy What do you think of this approach?

yihui:
I'll merge this PR first. The next CRAN release won't take place anytime soon (typically it will be several months later), so you still have enough time to review and discuss. I'm open to alternative approaches. Thanks!

atusy:
Sounds nice! thanks a lot!!

PR #2341, opened 2024-04-28
Update themes to be consistent with A.S. highlight.

dcser123:
Related to this [PR](https://github.com/yihui/highr/pull/11) and this [SO](https://stackoverflow.com/questions/71280893/knitr-rnw-highr-error-in-def-sng-color-not-being-created) post.


yihui:
Thanks!

PR #2349, opened 2024-06-06
[gha] Allowing to trigger workflow with extra packages to test

cderv:
let's try adding evaluate dev version as a job in the matrix


cderv:
This is using workflow dispatch event, which allows to easily test a knitr dependency dev version using API call to trigger the run . 

````bash
gh workflow run knitr-examples.yaml -f extra-packages=github::r-lib/evaluate
gh workflow run R-CMD-check.yaml  -f extra-packages=github::r-lib/evaluate
````

This will install dev **evaluate** and run usual knitr tests with this installed.

PR #2353, opened 2024-07-08
Avoid using evaluate internals

hadley:
In a few years, once this code is widespread in installed knitr, this will allow us to make some simplifications to evaluate


yihui (approved):
Thanks!

PR #2356, opened 2024-07-12
Explicitly state implicit R version dependency

MichaelChirico:
{evaluate} on CRAN now depends on R>=4.0.0. That means we can't install {knitr} on an older version of R (through normal means). I think it would be useful to state this up front rather than encounter a failure due to dependency {evaluate} being "missing".

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -114,7 +114,7 @@ Authors@R: c(
 Description: Provides a general-purpose tool for dynamic report generation in R
     using Literate Programming techniques.
 Depends:
-    R (>= 3.3.0)
+    R (>= 4.0.0)
 Imports:
     evaluate (>= 0.15),
     highr (>= 0.11),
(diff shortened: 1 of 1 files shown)


DESCRIPTION
```diff
@@ -114,7 +114,7 @@ Authors@R: c(
 Description: Provides a general-purpose tool for dynamic report generation in R
     using Literate Programming techniques.
 Depends:
-    R (>= 3.3.0)
+    R (>= 4.0.0)
```
yihui, line 117:
```suggestion
    R (>= 3.6.0)
```

yihui (commented):
**evaluate** will restore support for R 3.6.0 (https://github.com/r-lib/evaluate/pull/200 https://github.com/r-lib/evaluate/issues/173), and we need to wait for the next CRAN release of **evaluate**. Thanks!

PR #2358, opened 2024-07-31
Passing environment as an argument to `eng_r`

abhsarma:
I am trying to add a feature to my [R package](https://github.com/MUCollective/multiverse) where each code chunk implements multiple analysis (and each analysis is executed in a separate environment). This presents some complications during knitting as every analysis is executed in the `global_env()`: https://github.com/yihui/knitr/blob/fb1f4231d2df7b156e314686f39fc040ce807513/R/block.R#L189

I am wondering if it is possible to allow env be passed as an optional argument to `eng_r()` which defaults to `global_env()`. I've made a PR (#) with the relevant changes, but please let me know if this creates other issues or if there is a better way of doing this.

Thanks!


yihui (commented):
`knit_global()` is used in many places in this package. I guess this change is not enough.

I can provide an API to change the global environment (e.g., `knit_global(envir)` to change to a custom `envir`). Would that meet your need?

abhsarma:
I think so? This change was sufficient for what I was trying to achieve, but I am obviously not as familiar with the `knitr` internals as you are and there may be scenarios which I have not thought of, so I defer to what you think is best :)

yihui (approved):
Now you can change the environment via `knit_global(envir)`. I guess the good practice is to record the old envir, set the new envir, and restore the envir later, e.g.,

```r
old_env = knit_global(new_env)
on.exit(knit_global(old_env), add = TRUE)
# do your stuff in new_env
```

PR #2363, opened 2024-09-18
Use evaluate 1.0.0

hadley:
Including new `trim_intermediates_plot()` function.

Although now that I look closely at this I notice that rgl provides a `is_low_change()` method so that this might be a breaking change, and I need to consider this in evaluate.


yihui:
Confirmed this is a breaking change for **rgl**, as you suspected.

rgl registers `is_low_change.rglRecordedplot` (and imports `is_low_change` from knitr):
```r
is_low_change.rglRecordedplot <- function(p1, p2) {
  inherits(p2, "rglRecordedplot") && p1$plotnum == p2$plotnum
}
```
It discriminates 3D scenes by `plotnum`, not by display list. The current `merge_low_plot` dispatches through the `is_low_change` generic, so rgl's method is honored.

`evaluate::trim_intermediate_plots` (checked against evaluate 1.0.5) bypasses that path entirely:

1. It is not a generic and never calls `is_low_change`, so `is_low_change.rglRecordedplot` becomes dead code.
2. It gates on `is.recordedplot()`, i.e. `inherits(x, "recordedplot")`. rgl objects are class `rglRecordedplot`, **not** `recordedplot`, so they aren't even considered — rgl plots would silently lose intermediate-plot trimming.

Before merging, `evaluate` would need an extension point equivalent to `is_low_change` (so rgl can hook in), or knitr should retain the generic as a shim for non-base-graphics plots. Otherwise this needs a coordinated rgl release. Holding until the rgl path is resolved.

yihui:
Closing — as noted above, this would drop the `is_low_change` generic that rgl extends. Keeping it in knitr for now. Thanks @hadley! Happy to take a new PR down the line if `evaluate` grows an extension point for this (or the rgl side gets sorted).

PR #2364, opened 2024-09-18
Drop `fix_recordedPlot()`

hadley:
I'm reasonably certain that this is only required in R 3.3.0 and earlier as it was fixed in https://www.stat.auckland.ac.nz/~paul/Reports/DisplayList/dl-record.html. knitr now requires R 3.6, so this code is no longer needed.


yihui (approved):
Thanks!

PR #2365, opened 2024-09-19
fix #2332: pass knitr::asis_output() to the `output` hook

### LINKED ISSUE #2332: Make `asis_output` works for Quarto the same as `results: asis`
Related to 
* https://github.com/quarto-dev/quarto-cli/issues/9185

Example 
````markdown
---
title: "Test"
format: 
  typst:
    keep-typ: true
    keep-md: true
---

```{r}
#| results: asis
knitr::asis_output("`#set page(flipped: true)`{=typst}")
```

````

Using `asis_output()` will not apply raw output here, because `options$results = 'asis'` us onlyfor LaTeX output 

https://github.com/yihui/knitr/blob/031e08c74c501e0e1a5d625e0c053977b7a230ec/R/output.R#L514-L518

I think we need it for Quarto as the wrapping used will check for this value 
https://github.com/quarto-dev/quarto-cli/blob/f43107222f70ca3579a520d496597b75b5bd0097/src/resources/rmd/patch.R#L116-L117

yihui:
and every `output` hook should check if the chunk option `results == 'asis'`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 - In-chunk references of the form `<<label>>` should not be resolved if `label` is not found in the document (thanks, @jennybc @gadenbuie, #2360).
 
+- `asis_output()` was not passed to the `output` hook (thanks, @cderv, #2332).
+
 ## MAJOR CHANGES
 
 - Unbalanced chunk delimiters (fences) in R Markdown documents are strictly prohibited now.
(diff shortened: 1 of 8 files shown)


cderv (approved):
FWIW our current tests are passing with this PR. 

I need to think more about the potential problem, but I think we do already overwrite knitr's output hook in Quarto to do just this. 

So I think this is all good.

yihui:
Okay, I'll merge it for now.

PR #2369, opened 2024-10-10
feat: provide a stack trace for fatal errors in chunk

eternal-flame-AD:
This is the idea which will provide a backtrace of the error instead of just the message: https://github.com/r-lib/evaluate/pull/226#issuecomment-2403428797.

Before:

```r
Error in select(., nothing) : Can't select columns that don't exist.
✖ Column `nothing` doesn't exist.

Quitting from lines 3-11 [unnamed-chunk-1] (bad.Rmd)
```
After: 

```r
Error in `eng_r()`:
! Error in chunk unnamed-chunk-1
Caused by error in `select()`:
! Can't select columns that don't exist.
✖ Column `nothing` doesn't exist.
Backtrace:
 1. global myerror()
 4. dplyr:::select.data.frame(., nothing)

Quitting from lines 3-11 [unnamed-chunk-1] (bad.Rmd)
```

### DIFF
--- a/R/block.R
+++ b/R/block.R
@@ -259,14 +259,24 @@ eng_r = function(options) {
   } else if (cache.exists && isFALSE(options$cache.rebuild)) {
     cache$output(options$hash, 'list')
   } else in_input_dir(
-    evaluate(
-      code, envir = env, new_device = FALSE,
-      keep_warning = if (is.numeric(options$warning)) TRUE else options$warning,
-      keep_message = if (is.numeric(options$message)) TRUE else options$message,
-      stop_on_error = if (is.numeric(options$error)) options$error else {
-        if (options$error && options$include) 0L else 2L
-      },
-      output_handler = knit_handlers(options$render, options)
+    withCallingHandlers(
+      evaluate(
+        code, envir = env, new_device = FALSE,
+        keep_warning = if (is.numeric(options$warning)) TRUE else options$warning,
+        keep_message = if (is.numeric(options$message)) TRUE else options$message,
+        stop_on_error = if (is.numeric(options$error)) options$error else {
+          if (options$error && options$include) 0L else 2L
+        },
+        output_handler = knit_handlers(options$render, options)
+      ),
+      error = function (cnd) {
+        if (xfun::pkg_available('rlang', '1.0.0')) {
+          rlang::abort(paste0("Error in chunk ", options$label), parent = cnd)
+        } else {
+          signalCondition(cnd)
+          stop(cnd)
+        }
+      }
     )
   )
   if (options$cache %in% 1:2 && (!cache.exists || isTRUE(options$cache.rebuild))) {


yihui (commented):
To be honest, I don't fully understand the condition handling stuff in R, but I wonder if this will be a better place to make the change: https://github.com/yihui/knitr/blob/0f59bbabb98a49bb76f2b222cd103d8998454acc/R/output.R#L312-L315

Thank you!

eternal-flame-AD:
I will think about this more later today, but my initial thoughts for this part is:

Rlang documentation says

> entrace() is meant to be used as a global handler.

It seems to be a UX improvement tool to make stop() emit a trace rather than a wrapper that magically adds stack trace. I tried a toy example:

```r
myerr <- \() { entrace(cnd = errorCondition("Test")) } # no stack trace when called
trace <- \() myerr() # stack trace only contains trace(), not myerr()
```

xfun::handle_error seems to be undocumented API and, is this meant to catch errors that are even supposed to crash R so it is written as an exit handler instead of just a catch?

yihui:
`xfun:::handle_error()` doesn't try to catch errors at all (it doesn't use `tryCatch()` or `withCallingHandlers()`). It is intended to run a "handler" function when errors occur in the expression. You can basically treat `xfun:::handle_error(expr)` as `expr`.

eternal-flame-AD:
My original thought about writing it there was I think an R backtrace is only useful for R engine.  Seems only R engine output is improved (presumably because the call to evaluate() here erased the call stack before your chunk-level handler took it. 

We probably don't really have a case of where we intentionally want to suppress traces for one type of engine. It is fine and probably clearer to lift it up.

Found another issue when testing this that I can reproduce on master ... I will look at it this weekend to see what condition triggered this and if it is my environment is bad or another bug

````py

```{python}
print("Hello Python!")
```

```
  Error in loadable(pkg) : node stack overflow
Error during wrapup: node stack overflow
Error: no more error handlers available (recursive errors?); invoking 'abort' restart

Quitting from lines 90-91 [hello] (error.Rmd)
```
```

**Comment by eternal-flame-AD** [comment]
Just a bump in case this got lost :)

**Comment by cderv** [comment]
No it does not get lost. I have a current workload on Quarto work because we're are working on v1.6 release shortly. So my mind is not right now in R work - sorry to keep you waiting. I'll come back to this very quickly !

**Comment by cderv** [comment]
So change in **evaluate** have been released to CRAN. This is now what we get when **rlang** is available

````markdown
---
title: test
---

```{r}
library(tidyverse)

mtcars %>%
  ggplot(aes(x = wt, y = mpg)) +
  geom_qq() +
  geom_smooth(method = "lm")
```

````

````r
> knitr::opts_chunk$set(error = FALSE)
> knitr::knit("test.Rmd")


processing file: test.Rmd
  |...................................................................................................................................................| 100% [unnamed-chunk-1]

Error in `geom_qq()`:
! Problem while computing stat.
ℹ Error occurred in the 1st layer.
Caused by error in `compute_layer()`:
! `stat_qq()` requires the following missing aesthetics: sample.
Backtrace:
  1. knitr::knit("test.Rmd")
  2. knitr:::process_file(text, output)
       at knitr/R/output.R:253:3
  5. knitr:::process_group(group)
       at knitr/R/output.R:311:5
     ...
 44. ggplot2 (local) f(l = layers[[i]], d = data[[i]])
 45. l$compute_statistic(d, layout)
 46. ggplot2 (local) compute_statistic(..., self = self)
 47. self$stat$compute_layer(data, self$computed_stat_params, layout)
 48. ggplot2 (local) compute_layer(..., self = self)

Quitting from lines 5-12 [unnamed-chunk-1] (test.Rmd)
````

I was thinking a more reduced backtrace, but somehow with simpler document this is better

````markdown
---
title: document with error
---

```{r}
f <- function() g()
g <- function() h()
h <- function() rlang::abort("!")
f()
```
````

````r
> knitr::knit("test.Rmd")


processing file: test.Rmd
  |...................................................................................................................................................| 100% [unnamed-chunk-1]

Error in `h()`:
! !
Backtrace:
 1. global f()
 2. global g()
 3. global h()

Quitting from lines 5-10 [unnamed-chunk-1] (test.Rmd)
````

So I don't think this PR is needed anymore. I'll keep trying improving but I need more **rlang** knowledge

Issue #1967, opened 2021-02-11
support fig.alt for non HTML output


cderv:
This would be nice to have `fig.alt` work also for other outputs than Pandoc supports. 

See discussion and notes : 

* About support in DOCX: https://github.com/yihui/knitr/pull/1900#issuecomment-717101270
* About support in Pandoc in general: https://github.com/yihui/knitr/issues/1965#issuecomment-777584905

What we know is 

* Office supports alt text and Pandoc supports it. It would requires a special handling though as the syntax work well for docx only (not HTML for example)
* For latex and PDF, it does not seem to exist a generic way but some Journals seems to have solutions in there templates (For example : https://authors.acm.org/journals/how-to-write-alt-text-and-why). So maybe it could be possible to come up with a solution. 

Related question on community: 
* https://community.rstudio.com/t/how-to-include-both-alt-text-and-caption-in-a-figure-to-knit-to-word-with-rmarkdown/95662/2

cderv:
See a user's workaround in https://github.com/rstudio/rmarkdown/issues/1867#issuecomment-801227405

apreshill:
Hi @cderv (cc @hadley)

Saving this link here for reference on how to add alt text for MS outputs: https://support.microsoft.com/en-us/topic/add-alternative-text-to-a-shape-picture-chart-smartart-graphic-or-other-object-44989b2a-903c-4d9a-b742-6a75b451c669

cderv:
See discussion in https://github.com/rstudio/rmarkdown/issues/2106#issuecomment-854486564 about PPTX format

cderv:
For later, this is what we got from Pandoc when parsing the syntax

````powershell
❯ pandoc -t native
![look a puffin](https://upload.wikimedia.org/wikipedia/commons/9/94/Puffin_Mrkoww.jpg "Title: A proper puffin")
^Z
[Para [Image ("",[],[]) [Str "look",Space,Str "a",Space,Str "puffin"] ("https://upload.wikimedia.org/wikipedia/commons/9/94/Puffin_Mrkoww.jpg","fig:Title: A proper puffin")]]
````

**Comment by zachmayer** [comment]
I'd love to have alt text for markdown outputs too

**Comment by capnrefsmmat** [comment]
FYI, LaTeX is beginning to support alt text: the graphicx package now accepts an `alt` option to `\includegraphics`, as in

```latex
\includegraphics[width=3in,alt={University logo}]{logo.pdf}
```

[...]

PR #2378, opened 2024-10-27
Support alt text for images in LaTeX output

capnrefsmmat:
Since 2021, graphicx supports an alt option containing alt text. Pass `fig.alt` through unless `knitr.include_graphics.alt` is FALSE (which can be set if e.g. you're using an older TeX Live whose graphicx does not support this.)

Ongoing LaTeX work to support accessible (tagged) PDFs will eventually use this option to embed alt text in PDFs. See <https://www.latex-project.org/publications/indexbytopic/pdf/>

See also #1967.

I haven't contributed to knitr before, so not sure of a few things:

1. I didn't see tests that checked `\includegraphics` output. Are there any tests I should change/add?
2. I added the option because graphicx versions before 2021 will throw an error when they get alt text, so some users may want to turn this off. But I defaulted the option to TRUE, so the alt text is included in LaTeX output by default. Should it default to FALSE instead? The new behavior will be surprising to people with old TeX installs, but on the other hand `fig.alt` is doing nothing for them right now. At the least I could add a FAQ entry for the error they'd get.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 - In-chunk references of the form `<<label>>` can be disabled via the chunk option `ref.chunk = FALSE` now (thanks, @jennybc @gadenbuie, #2360).
 
+- Added support for `fig.alt` for LaTeX output, i.e., using `\includegraphics[alt={alt text}]` (thanks, @capnrefsmmat, #2378).
+
 ## BUG FIXES
 
 - In-chunk references of the form `<<label>>` should not be resolved if `label` is not found in the document (thanks, @jennybc @gadenbuie, #2360).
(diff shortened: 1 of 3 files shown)


yihui (commented):
1. You can either add tests to `tests/testit/` (you can create a new script there) or add an example .Rnw document in the repo https://github.com/yihui/knitr-examples.
2. I tend to keep it simple for now by not introducing a global option. If users have to use an old version of LaTeX, they still have the option of not using `fig.alt`.

Thanks!

capnrefsmmat:
Okay, I've removed the global option and added a test script.

yihui (approved):
Thanks!

PR #2388, opened 2025-01-15
Improve error traceback

cderv:
This PR follows up on **evaluate** 1.0.3 work. 

- **knitr** does entrace all error with **rlang** when available. This is done now in the calling handler of evaluate, closest to the error being thrown

- **evaluate** does handle already trimming the backtrace for error thrown in the code evaluated. **knitr** need also to do it for error happening in its `value` handler passed to evaluate, and calling `knit_print()`
	- This is done at the `knit_print()` level directly; 
		But `options$render` could be use to change the value handler, and in that case this would not apply. So it could be done also at the value handler 
		https://github.com/yihui/knitr/blob/8a37857140a0c1a656d5ff76b9742f6a5c01933c/R/utils.R#L882-L887
	
	- Or **evaluate** could also do it so that it happens for any value handler, including knitr's one. 

## Tests 

They are added to companion PR in knitr-example 
- https://github.com/yihui/knitr-examples/pull/93

The CI check below are ran using this https://github.com/yihui/knitr-examples/pull/93 PR.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1131,8 +1131,6 @@ txt_pb = function(total, labels) {
 is_quarto = function() isTRUE(.knitEnv$is_quarto)
 
 with_options = function(expr, opts_list) {
-  old <- options(opts_list)
-  defer(options(old))
   local_options(opts_list)
   expr
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -845,6 +845,8 @@ current_input = function(dir = FALSE) {
 
 # cache output handlers from evaluate; see .onLoad
 default_handlers = NULL
+# use rlang handler for error only if available; see .onLoad
+rlang_entrace_handler = NULL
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -1127,3 +1129,20 @@ txt_pb = function(total, labels) {
 }
 
 is_quarto = function() isTRUE(.knitEnv$is_quarto)
+
+with_options = function(expr, opts_list) {
+  old <- options(opts_list)
+  defer(options(old))
```
hadley, line 1135: Is this necessary with the definition of `local_options()` below?

cderv: definitely not. Thanks for the review - this is bad rebase of the PR. 

Thank you!

yihui (approved):
Thanks!

PR #2390, opened 2025-01-22
Display error traceback when vignettes fail in R CMD build

hadley:
Recording some work so that I can come back to it later.

### DIFF
--- a/R/output.R
+++ b/R/output.R
@@ -318,7 +318,7 @@ process_file = function(text, output) {
           if (progress && is.function(pb$interrupt)) pb$interrupt()
 
           if (xfun::pkg_available('rlang', '1.0.0')) {
-            if (is_R_CMD_build()) {
+            if (is_R_CMD_build() || is_R_CMD_check()) {
               cnd <- tryCatch(rlang::entrace(e), error = identity)
               error <<- format(cnd)
             } else {
@@ -353,7 +353,7 @@ process_file = function(text, output) {
 }
 
 rule <- function() {
-  # Used by pkgbuild
+  # Used by pkgbuild; please don't change without letting us know
   paste0(strrep("~", getOption("width")), "\n")
 }
 
--- a/R/output.R
+++ b/R/output.R
@@ -308,20 +308,36 @@ process_file = function(text, output) {
     if (progress && is.function(pb$update)) pb$update(i)
     group = groups[[i]]
     knit_concord$set(block = i)
+
+    extra <- NULL
+
     res[i] = xfun:::handle_error(
       withCallingHandlers(
         if (tangle) process_tangle(group) else process_group(group),
         error = function(e) {
           if (progress && is.function(pb$interrupt)) pb$interrupt()
-          if (xfun::pkg_available('rlang', '1.0.0')) rlang::entrace(e)
+
+          if (xfun::pkg_available('rlang', '1.0.0')) {
+            if (Sys.getenv("_R_CHECK_LICENSE_") == "") {
+              cnd <- tryCatch(rlang::entrace(e), error = identity)
+              extra <<- paste(capture.output(print(cnd)), "\n", collapse = "")
+            } else {
+              rlang::entrace(e)
+            }
+          }
         }
       ),
       function(loc) {
         setwd(wd)
         write_utf8(res, output %n% stdout())
-        paste0('\nQuitting from lines ', loc)
+
+        message <- paste0('\nQuitting from lines ', loc)
+        if (!is.null(extra)) {
+          message <- paste0(message, "\n", rule(), extra, rule())
+        }
       },
-      if (labels[i] != '') sprintf(' [%s]', labels[i]), get_loc
+      if (labels[i] != '') sprintf(' [%s]', labels[i]),
+      get_loc
     )
     knit_concord$set(offset = NULL)
   }
@@ -335,9 +351,13 @@ process_file = function(text, output) {
   res
 }
 
+rule <- function() {
+  paste0(strrep("~", getOption("width")), "\n")
+}
+
 # return a string to point out the current location in the doc
 get_loc = function(label = '') {
-  paste0(current_lines(), label, sprintf(' (%s)', knit_concord$get('infile')))
+  paste0(knit_concord$get('infile'), ":", current_lines(), label)
 }
 
 auto_out_name = function(input, ext = tolower(file_ext(input))) {
--- a/R/output.R
+++ b/R/output.R
@@ -308,18 +308,15 @@ process_file = function(text, output) {
     if (progress && is.function(pb$update)) pb$update(i)
     group = groups[[i]]
     knit_concord$set(block = i)
-
-    error <- NULL
-
+    error = NULL
     res[i] = xfun:::handle_error(
       withCallingHandlers(
         if (tangle) process_tangle(group) else process_group(group),
         error = function(e) {
           if (progress && is.function(pb$interrupt)) pb$interrupt()
-
           if (xfun::pkg_available('rlang', '1.0.0')) {
             if (is_R_CMD_build() || is_R_CMD_check()) {
-              cnd <- tryCatch(rlang::entrace(e), error = identity)
+              cnd = tryCatch(rlang::entrace(e), error = identity)
               error <<- format(cnd)
             } else {
               rlang::entrace(e)
@@ -330,15 +327,13 @@ process_file = function(text, output) {
       function(loc) {
         setwd(wd)
         write_utf8(res, output %n% stdout())
-
-        message <- paste0('\nQuitting from ', loc)
+        message = paste0('\nQuitting from ', loc)
         if (!is.null(error)) {
-          message <- paste0(message, "\n", rule(), error, "\n", rule())
+          message <- paste0(message, '\n', rule(), error, '\n', rule())
         }
         message
       },
-      if (labels[i] != '') sprintf(' [%s]', labels[i]),
-      get_loc
+      if (labels[i] != '') sprintf(' [%s]', labels[i]), get_loc
     )
     knit_concord$set(offset = NULL)
   }
@@ -352,14 +347,14 @@ process_file = function(text, output) {
   res
 }
 
-rule <- function() {
+rule = function() {
   # Used by pkgbuild; please don't change without letting us know
-  paste0(strrep("~", getOption("width")), "\n")
+  paste0(strrep('~', getOption('width')), '\n')
 }
 
 # return a string to point out the current location in the doc
 get_loc = function(label = '') {
-  paste0(knit_concord$get('infile'), ":", current_lines(), label)
+  paste0(knit_concord$get('infile'), ':', current_lines(), label)
 }
 
 auto_out_name = function(input, ext = tolower(file_ext(input))) {
--- a/R/output.R
+++ b/R/output.R
@@ -318,9 +318,9 @@ process_file = function(text, output) {
           if (progress && is.function(pb$interrupt)) pb$interrupt()
 
           if (xfun::pkg_available('rlang', '1.0.0')) {
-            if (Sys.getenv("_R_CHECK_LICENSE_") == "") {
+            if (Sys.getenv("R_BUILD_TEMPLIB") != "") {
               cnd <- tryCatch(rlang::entrace(e), error = identity)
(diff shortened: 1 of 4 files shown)


R/output.R
```diff
@@ -308,20 +308,36 @@ process_file = function(text, output) {
...
+
     res[i] = xfun:::handle_error(
       withCallingHandlers(
         if (tangle) process_tangle(group) else process_group(group),
         error = function(e) {
           if (progress && is.function(pb$interrupt)) pb$interrupt()
-          if (xfun::pkg_available('rlang', '1.0.0')) rlang::entrace(e)
+
+          if (xfun::pkg_available('rlang', '1.0.0')) {
+            if (Sys.getenv("_R_CHECK_LICENSE_") == "") {
```
hadley, line 321: `is_R_CMD_check()` doesn't work correctly for me. Need to investigate why.

R/output.R
```diff
@@ -308,20 +308,36 @@ process_file = function(text, output) {
...
+    extra <- NULL
+
     res[i] = xfun:::handle_error(
       withCallingHandlers(
         if (tangle) process_tangle(group) else process_group(group),
         error = function(e) {
           if (progress && is.function(pb$interrupt)) pb$interrupt()
-          if (xfun::pkg_available('rlang', '1.0.0')) rlang::entrace(e)
+
+          if (xfun::pkg_available('rlang', '1.0.0')) {
```
hadley, line 320: For vignettes, this implies that your package needs to depend on rlang. I think that's ok, but should be documented somewhere.

R/output.R
```diff
@@ -335,9 +351,13 @@ process_file = function(text, output) {
...
 }
 
+rule <- function() {
+  paste0(strrep("~", getOption("width")), "\n")
+}
+
 # return a string to point out the current location in the doc
 get_loc = function(label = '') {
-  paste0(current_lines(), label, sprintf(' (%s)', knit_concord$get('infile')))
+  paste0(knit_concord$get('infile'), ":", current_lines(), label)
```
hadley, line 360: This uses a more standard format for file + line numbers that VS Code/Positron will autolink.  I think it's still worthwhile to use cli here, so I'll explore adding that in the future.

hadley: And cli hyperlinks don't work in R CMD check anyway, because if it did, it would open in the file in the check directory, not the current project.

R/output.R
cderv: So here we want to do the extra capture when not in R CMD CHECK ? 

My understanding is `_R_CHECK_LICENSE_` is `TRUE` when  checking with `--as-cran` which `is_R_CMD_check()` is testing. 🤔 

I have a doubt now.

hadley: Right, we only need to add the extra traceback inside `R CMD check`. But you're right: now that I look at it the condition appears to be the wrong way around.  

...

Oh yeah, it's not actually doing anything 😆 

OTOH it's not actually that bad to do this unconditionally. Now when you just knit it with a button (so it runs in the terminal), you get:

```
4/5 [unnamed-chunk-1]
Error in h() : This is a problem
Calls: .main ... withCallingHandlers -> withVisible -> eval -> eval -> f -> g -> h

Quitting from lines fail.Rmd:16-18 [unnamed-chunk-1]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<error/rlang_error> 
Error in `h()`: 
! This is a problem 
--- 
Backtrace: 
    ▆ 
 1. └─global f() 
 2.   └─global g() 
 3.     └─global h() 
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

OTOOH you end up with two tracebacks if you run `rmarkdown::render()`.

hadley: Oooooh, the problem is that error occurs during R CMD **buld** not check.

hadley:
I think this is ready for review. When I install this version of knitr and build https://github.com/hadley/testvignette, I see this:

```
E  creating vignettes (1.1s)
   --- re-building ‘fail.Rmd’ using rmarkdown
   
   Quitting from fail.Rmd:16-18 [unnamed-chunk-1]
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   <error/rlang_error>
   Error in `h()`:
   ! This is a problem
   ---
   Backtrace:
       ▆
    1. └─global f()
    2.   └─global g()
    3.     └─global h()
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   
   Error: processing vignette 'fail.Rmd' failed with diagnostics:
   This is a problem
   --- failed re-building ‘fail.Rmd’
   
   SUMMARY: processing the following file failed:
     ‘fail.Rmd’
   
   Error: Vignette re-building failed.
   Execution halted
```

There's quite a lot of duplication in the output, but I don't see an obvious way to resolve this, because it's generated by R CMD build, not knitr or rmarkdown.

bastistician:
Thanks for looking into this. Inspecting knitr-based vignette errors when mass-checking R packages has been a bit cumbersome in the past, mostly because the call was missing (so I have sometimes used a patched R to get more output). So I was happy to find this PR and hope that diagnostics will be provided for both vignette build and check errors in the future.

Both R CMD build *and* check use `tools::buildVignettes()` to build ("weave") each package vignette using the engine it declares. When the engine's `weave()` function errors, this is reported with
```
Error: processing vignette '<file>' failed with diagnostics:
<conditionMessage>
```
For Sweave vignettes, the message includes the call (from evaluating chunk expressions using `try()`) as well as the chunk number/label (from `RweaveTryStop()`). Given that this PR is specifically about building package vignettes, I wonder if it would make more sense to amend the `weave()` functions in knitr's vignette engines rather than `process_file()`? Could `knitr:::vweave()` etc. catch errors and collect all useful information in the message?

Alternatively, I have experimented with a minor change of the `buildVignettes` output from `<conditionMessage>` to `<as.character(e)>` (the patch mentioned above), which would include both the call and the message from the error condition, also for "rlang_error"s (but not their backtrace it seems), so something like:
```
Error: processing vignette 'fail.Rmd' failed with diagnostics:
Error in `h()`:
! This is a problem
```
Printing the condition object might be another option.

gaborcsardi:
AFAICT you already get a traceback for `R CMD build`, with line numbers and chunk names:
```
❯ R CMD build testvignette
* checking for file ‘testvignette/DESCRIPTION’ ... OK
* preparing ‘testvignette’:
* checking DESCRIPTION meta-information ... OK
* installing the package to build vignettes
* creating vignettes ... ERROR
--- re-building ‘fail.Rmd’ using rmarkdown

Quitting from fail.Rmd:16-18 [unnamed-chunk-1]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<error/rlang_error>
Error in `h()`:
! This is a problem
---
Backtrace:
    ▆
 1. └─global f()
 2.   └─global g()
 3.     └─global h()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Error: processing vignette 'fail.Rmd' failed with diagnostics:
This is a problem
--- failed re-building ‘fail.Rmd’

SUMMARY: processing the following file failed:
  ‘fail.Rmd’

Error: Vignette re-building failed.
Execution halted
```

hadley:
@bastistician those all sounds like reasonable alternative implementations, but I can't quite tell if you think there's something missing in this PR?

bastistician:
Yes, I was especially hoping for improved error reporting for knitr-based vignettes also during `R CMD check` (e.g., when an updated dependency breaks a vignette or downstream vignettes are broken in a revdep check). AFAICS this PR adds special instrumentation only for `R CMD build` so wouldn't improve the check output:
```
* checking re-building of vignette outputs ... ERROR
Error(s) in re-building vignettes:
  ...
--- re-building ‘fail.Rmd’ using rmarkdown

Quitting from fail.Rmd:16-18 [unnamed-chunk-1]
Error: processing vignette 'fail.Rmd' failed with diagnostics:
This is a problem
--- failed re-building ‘fail.Rmd’
```
In my comment above I was reflecting whether the error handling should instead go to knitr's vignette engines so that it affects `tools::buildVignettes()` regardless of being called during build or check.

OTOH, I can also see some value in a more general change of what `tools::buildVignettes()` outputs in case it catches a "weave" error, not just the `conditionMessage()`. I think that would need a report on R's Bugzilla.

cderv (approved):
Looks good. 

> could easily add a || is_R_CMD_check() to make this PR to work then too.

Doing this seems to work also 
```
* using log directory 'C:/Users/chris/AppData/Local/Temp/testvignette/testvignette.Rcheck'
* using R version 4.4.2 (2024-10-31 ucrt)
* using platform: x86_64-w64-mingw32
* R was compiled by
    gcc.exe (GCC) 13.3.0
    GNU Fortran (GCC) 13.3.0
* running under: Windows 11 x64 (build 26100)
* using session charset: UTF-8
* checking for file 'testvignette/DESCRIPTION' ... OK
* this is package 'testvignette' version '0.0.0.9000'
* package encoding: UTF-8
* checking package namespace information ... OK
* checking package dependencies ... OK
* checking if this is a source package ... OK
* checking if there is a namespace ... OK
* checking for executable files ... OK
* checking for hidden files and directories ... OK
* checking for portable file names ... OK
* checking whether package 'testvignette' can be installed ... OK
* checking installed package size ... OK
* checking package directory ... OK
* checking DESCRIPTION meta-information ... OK
* checking top-level files ... OK
* checking for left-over files ... OK
* checking index information ... OK
* checking package subdirectories ... OK
* checking whether the package can be loaded ... OK
* checking whether the package can be loaded with stated dependencies ... OK
* checking whether the package can be unloaded cleanly ... OK
* checking whether the namespace can be loaded with stated dependencies ... OK
* checking whether the namespace can be unloaded cleanly ... OK
* checking loading without being on the library search path ... OK
* checking files in 'vignettes' ... WARNING
Files in the 'vignettes' directory but no files in 'inst/doc':
  'fail.Rmd'
* checking examples ... NONE
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes ... WARNING
Directory 'inst/doc' does not exist.
Package vignette without corresponding single PDF/HTML:
  'fail.Rmd'
* checking re-building of vignette outputs ... ERROR
Error(s) in re-building vignettes:
  ...
--- re-building 'fail.Rmd' using rmarkdown

Quitting from fail.Rmd:16-18 [unnamed-chunk-1]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
<error/rlang_error>
Error in `h()`:
! This is a problem
---
Backtrace:
    ▆
 1. └─global f()
 2.   └─global g()
 3.     └─global h()
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Error: processing vignette 'fail.Rmd' failed with diagnostics:
This is a problem
--- failed re-building 'fail.Rmd'

SUMMARY: processing the following file failed:
  'fail.Rmd'

Error: Vignette re-building failed.
Execution halted

* checking PDF version of manual ... OK
* DONE
Status: 1 ERROR, 2 WARNINGs

````

R/output.R
hadley: Yeah, sorry, I meant depend in the loose sense.

R/output.R
hadley: Yeah, although using cli would be even more useful there (although not helpful for vignettes)

yihui (approved):
> we could easily add a `|| is_R_CMD_check()` to make this PR to work

I'll merge the PR after that is added. Thanks!

hadley:
@yihui done!

hadley:
Thanks @yihui!

PR #2391, opened 2025-01-29
Allow for column alignment for org table export

mclements:
For table export to Org Mode, (i) the column alignment is currently lost and (ii) the output table includes colons which are not part of the Org Mode table specification. To address these issues, I have added some code to keep the column alignment and cleaned up the output table format.

I hope that this is helpful.

Sincerely, Mark.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 - Display error traceback when vignettes fail in `R CMD build` (thanks, @hadley, #2390).
 
+- `kable()` properly supports column alignment for Org Mode tables now (thanks, @mclements, #2391).
+
 ## MINOR CHANGES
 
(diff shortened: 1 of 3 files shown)


yihui (commented):
Thanks! Can this be achieved by the `align.fun` argument of `kable_mark()`? Manipulating the strings via regex feels like an ad hoc hack. Perhaps a custom `align.fun` function will be much more straightforward.

mclements:
I have now updated this to use `kable_mark()`, with one call to `gsub()`.

What are the semantics when there are no column names? Specifically, should we include a top line? The current implementation removes both the blank column names (index=1) and the line (index=3):

```
if (is_null_colnames) res = res[-c(1,3)]
```

mclements:
If you are okay with my assumptions for a table with no header, then can you review the pull request as it is, please?

yihui (approved):
> If you are okay with my assumptions for a table with no header, then can you review the pull request as it is, please?

That sounds good to me. I've tweaked the PR and will merge soon. Thanks!

PR #2396, opened 2025-03-31
`entrace()` can return `NULL`

hadley:
Unfortunately we somehow missed this in our testing of #2390, but `entrace()` returns `NULL` if the error has already been entraced. This changes fixes that, preserving the original error object.


cderv:
I reworked on #2388 and this will cover the problem from pkgdown. 

`rlang::entrace()` always happen a evaluate handler now, so we just need to catch the error message and duplicate it when R CMD check context so that it shows in log output. 

This should work in vignette too (from what I tested with failrmd fork).

hadley:
Great, thanks!

PR #2403, opened 2025-05-23
Remove plyr usage from example

MichaelChirico:
I'm going around removing old {plyr} code with regex and this came up. I'm not sure if {plyr} is even _technically_ a dependency in this case. I have no idea what `merge_recurse()` is, it's not on CRAN:

https://github.com/search?q=repo%3Acran%2Freshape%20path%3A%2FNAMESPACE%24%2F%20%2Fmerge_recurse%2F&type=code

Feel free to reject this PR, just sharing since it's done already & short.


yihui (approved):
Thanks!

PR #2422, opened 2025-10-28
Implement OpenTelemetry Instrumentation

### LINKED ISSUE #2421: Add OpenTelemetry
Implement a basic layer of OpenTelemtery so that users can get tracing 'for free' if they have this enabled.

The goal is to implement across packages which may be used 'in production'.

Allows rich observability e.g. if packages such as mirai are run within knitr chunks.

shikokuchuo:
Closes #2421.

This PR implements basic OpenTelemetry instrumentation for knitr.

Abides by general otel [semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/) in terms of what span names / attributes should look like. As knitr does not fit into any of the existing categories, there are no compulsory attributes to include.

For now, I've included 'engine', 'echo' and 'eval' options as attributes (metadata). Feel free to suggest others to add.

The following is an example screenshot of the spans created (from knitting purrr vignettes). You can see the mirai spans nested within the knitr chunk span.

Screenshot 2025-10-28 at 15 35 08

As per our video call, I've verified that this works fine for python chunks running reticulate.

We cache the otel tracer on package load - if it's not enabled then there is minimal overhead as `otel_local_active_span()` has an early return if the cached `otel_is_tracing` variable is `FALSE`.

Todo:

- [x] Consider other attributes
- [x] Add docs and tests

Feel free to ask any questions / point to other alternatives for implementation that I may not have thought of.

Thanks!


yihui (approved):
Looks good to me! Just let me know when you think it's ready and I'll merge it. Thanks!

cderv (approved):
Indeed really clean integration. Thank you!

atusy:
How nice! I didn't know there is otel in R.

yihui (approved):
I'll merge this PR in a second. If you come up with anything else, please feel free to open subsequent PRs. Thanks!

PR #2423, opened 2025-11-05
Update internal `otel_refresh_tracer()` helper

shikokuchuo:
A small follow-up to #2422.

Just so you have the latest version of this helper, which contains much cleaner code. No resulting change in behaviour.


yihui (approved):
Great. Thanks!

PR #2424, opened 2025-11-10
Align otel tracer caching implementation

shikokuchuo:
A final follow-up to #2422. As we're now merging implementations for httr2, ellmer and DBI, this PR simply aligns the implementation for ease of future maintenance. No outward changes.


yihui (approved):
Thanks!

PR #2443, opened 2026-06-30
fix: render knit2html output next to input when output dir differs (#2408)

### LINKED ISSUE #2408: Including images when output dir is elsewhere
I have a simple.Rmd file with no more than: 
````
test

``` {r Graph2,  fig.width=7.7,fig.height=6.8}
plot(1,1)

``` 
````

If I use 

```{r}
knitr::knit2html("simple.rmd", output= "OUT.html") 
```  

everything is fine. 
If I use 

```{r}
some_where_else <- "C:\\Temp\\"
knitr::knit2html("simple.rmd", output=paste0(some_where_else, "OUT.html") )
```  

the image is not included. 
``` 
Warnmeldungen:
1: In replace(z) :
 File 'figure/Graph2-1.png' not found (hence cannot be embedded).
``` 

Is this intended? 

Versions: R 4.5.0  knitr 1.50
It works under R 4.3.3 and older knitr version 1.23 or something

LeonidasZhak:
Summary

Stage local Markdown resources before `knit2html()` renders HTML to a different output directory, so generated figure files can still be embedded.

Thanks to maintainers

Thanks for the issue discussion and for narrowing the problem to the `knit2html()` / `mark_html()` resource handoff.

Issue or motivation

`knit2html()` can warn that files such as `figure/Graph2-1.png` cannot be embedded when the HTML output is written outside the knitted Markdown directory. This leaves broken figure references in the final HTML. The issue is reported in #2408.

Root cause

`knit()` writes the intermediate Markdown and figure files next to the input document, but `mark_html()` resolves embeddable local resources relative to the output location. When the output directory is different, the figure files are no longer available where the HTML renderer looks for them.

Change

- Detect local relative resource paths referenced by the knitted Markdown.
- Stage those files into the HTML output directory before calling `mark_html()`.
- Remove staged files and any empty staging directories after rendering.
- Add a regression test that reproduces the moved-output case and checks that the PNG is embedded instead of left as a broken relative path.

Tests

- `Rscript -e 'pkgload::load_all(".", quiet = TRUE); source("tests/testit/test-utils-conversion.R")'`
- `Rscript -e 'pkgload::load_all(".", quiet = TRUE); source("tests/testit/test-hooks-md.R")'`
- `Rscript -e 'testit::test_pkg("knitr")'`
- `R CMD build .`
- `_R_CHECK_FORCE_SUGGESTS_=false R CMD check --no-manual knitr_1.51.7.tar.gz`

Scope

This stays inside `knit2html()` resource staging for moved HTML outputs. It does not change `rmarkdown::render()`, does not modify `litedown` itself, and does not alter unrelated Markdown rendering behavior.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -68,6 +68,7 @@ Authors@R: c(
     person(c("Kevin", "K."), "Smith", role = "ctb"),
     person("Kirill", "Mueller", role = "ctb"),
     person("Kohske", "Takahashi", role = "ctb"),
+    person("Leonidas", "Zhak", role = "ctb"),
     person("Lorenz", "Walthert", role = "ctb"),
     person("Lucas", "Gallindo", role = "ctb"),
     person("Marius", "Hofert", role = "ctb"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 - The `alt` attribute of figure images now has HTML tags stripped (via `xfun::strip_html()`) and is properly escaped for use in HTML attributes (via `xfun::html_escape(attr = TRUE)`). Previously, HTML in `fig.cap` (e.g., a link) could appear verbatim in `alt`, and double quotes in captions could break the `alt` attribute value (thanks, @cderv, #2004).
 
+- `knit2html()` now stages local figure resources when the HTML output is written to a different directory, so embedded images no longer break with warnings like `File 'figure/...png' not found` (thanks, @LeonidasZhak, #2408).
+
 - Due to a change in Asymptote, the output file may not be recognized or included correctly (thanks, @DeliciousRoastPotato @shangeconnew, #2025).
 
 - `has_crop_tools()` should not throw an error when `tlmgr` exists but cannot be executed (thanks, @lsandig, rstudio/rmarkdown#2612). It returns `FALSE` now.
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 - The `alt` attribute of figure images now has HTML tags stripped (via `xfun::strip_html()`) and is properly escaped for use in HTML attributes (via `xfun::html_escape(attr = TRUE)`). Previously, HTML in `fig.cap` (e.g., a link) could appear verbatim in `alt`, and double quotes in captions could break the `alt` attribute value (thanks, @cderv, #2004).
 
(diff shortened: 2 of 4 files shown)


LeonidasZhak:
Signed. Thank you. 🤩

yihui:
I pushed a commit that simplifies this fix (`maintainerCanModify` was on) — please take a look and let me know if it works for you.

## Root cause

The intermediate figures aren't really a "resource staging" problem. `knit()` writes `figure/*.png` next to the input, but the renderer resolves relative resource paths against the **output** directory. In current knitr, `knit2html()` uses litedown's `mark()`, and litedown deliberately resolves resources against the output dir:

```r
# litedown/R/mark.R
out_dir = dirname(output_path(input, output) %||% '.')
...
ret = in_dir(out_dir, embed_resources(ret, options))
```

So when the output goes elsewhere, the relative `figure/...png` paths point at the wrong directory and can't be embedded. (This is why it worked in knitr 1.23, which used the old `markdown` package that resolved against the input dir.)

## Simpler approach

Because the embedded HTML is self-contained, we can just render next to the knitted Markdown (where the figures already are) and then move the finished HTML to the output location. That avoids parsing Markdown for resource URLs (which misses reference-style links, CSS `url()`, `srcset`, etc.) and avoids copying/cleaning files in the output tree:

```r
if (xfun::same_path(dirname(out), dirname(output))) {
  mark(out, output, ...)
} else {
  html = with_ext(out, 'html')
  on.exit(if (!xfun::same_path(html, output)) file.remove(html), add = TRUE)
  mark(out, html, ...)
  file.copy(html, output, overwrite = TRUE)
}
```

Your regression test still passes unchanged.

Caveat: this is correct when resources are embedded (the default for `mark_html`/`self_contained`). If someone disables embedding, figures stay as relative links and moving the file alone won't fix them — but the original copy-based approach was also only meaningful for the embedding case.

## Recommendation

For robust R Markdown → HTML rendering where the output lives in a different directory, `litedown::fuse()` handles figure paths correctly out of the box (it sets `fig.path` relative to the output with absolute paths). `knit2html()` is a v1 convenience wrapper; the commit above makes the moved-output case work, and I've noted `fuse()` in NEWS as the more robust option.

yihui (approved):

LeonidasZhak:
Thanks for the detailed follow-up. I reviewed the simplified approach and agree that it addresses the root cause more directly: render beside the knitted Markdown, embed resources there, and then move the completed HTML. The unchanged regression test and the caveat about non-embedded resources make sense. #2443 looks good to me.

PR #2451, opened 2026-08-20
Support customization for dev='gridSVG' using dev.args

### LINKED ISSUE #2450: Feature request: Customizing dev="gridSVG" (possibly via dev.args)
I am trying to use `dev='gridSVG'` to produce SVG plots via `gridSVG`. This works well, but I want to pass some additional customization arguments: 

(a) When the backend `svglite` device is called, I want to specify `pointsize = 14`.
(b) When `grid.export()` is called, I want to specify `strict=FALSE`.

I don't think it's currently possible to do either, but it would be nice to have this possibility.

I don't know what the best interface do this would be. For other devices, the `dev.args` is the obvious place, but here there are two different functions involved, so that may be misleading. Still, I did some experiments to get `dev.args = list(pointsize = 14, strict = FALSE)` to work. This required some argument filtering in a few places.

If this is a reasonable approach, my changes can be seen [here](https://github.com/yihui/knitr/compare/master...deepayan:knitr:master). I'm happy to submit a PR if you want.

For now, I'm bypassing this by defining a custom `knit_print()` method.

---

- [X] I have fully read the issue guide at https://yihui.org/issue/.
- [X] I have provided the necessary information about my issue.
- [X] I have learned the Github Markdown syntax, and formatted my issue correctly.

deepayan:
PR for <https://github.com/yihui/knitr/issues/2450>


yihui:
Thanks! Your approach of filtering `dev.args` per target function is exactly right for gridSVG (passing `pointsize` to `svg()` and `strict` to `grid.export()`). I kept that.

I did drop one change from the PR: filtering `dargs` against `formals(device)` in `plot2dev()`. Many devices returned by `dev_get()` are wrappers of the form `function(...) ...` (`png`, `jpeg`, `tiff`, `ragg_*`, `quartz_*`, `Cairo*`, `tikz`), so `names(formals(device))` is just `"..."` and the intersection would silently discard *all* `dev.args` for those devices (and break the tikz `engine` arg). Instead the filtering is applied only on the `grid.export()` path, where the extra args actually matter.

Also added the equivalent `formals()` filtering for `svglite`.

deepayan:
Thanks, this looks good.

My goal in filtering the arguments sent to `do.call(device, ...)` was to ensure that things like `strict=TRUE` would be dropped when calling `pdf()` (say), but I didn't realize the drawback. This means I can set `strict=FALSE` only when `dev='gridSVG'`, but I can live with that.

[I also realized later that `pointsize` is effectively ignored by grid, and dev='gridSVG' doesn't support (base) graphics, so support for `pointsize` is currently pointless.]

PR #2452, opened 2026-08-26
feat: let `animation.hook` generate LaTeX code

jolars:
`hook_plot_tex()` hardcoded `\animategraphics{}` for chunks with `fig.show='animate'`, so the chunk option `animation.hook` and the package option `animation.fun` were silently ignored for LaTeX output, even though `hook_plot_html()` has always honoured them.

Call the hook when either option is set to a user-provided function, and keep generating `\animategraphics{}` otherwise. The built-in hooks generate HTML, so they stay ignored for LaTeX output and existing documents are unaffected. This lets LaTeX packages other than `animate` be used for animations, e.g. `xmpmulti` for beamer overlays, which is something I need for my course slides that I write in quarto.

I used Claude Code Opus 5 to help out here.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: knitr
 Type: Package
 Title: A General-Purpose Package for Dynamic Report Generation in R
-Version: 1.51.13
+Version: 1.51.14
 Authors@R: c(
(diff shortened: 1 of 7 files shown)


yihui (approved):
Thanks!

Issue #2171, opened 2022-09-17
default relative path processing in `include_graphics()` can create broken path


naikymen:
Hi!

I'm having difficulties including images during knitting.

Here is a reproducible example: [test.zip](https://github.com/yihui/knitr/files/9592044/test.zip)

The notebook is in `R/Rmd1/notebook1.Rmd`. If I run the notebook in RStudio everything works.

Knit the document to get the following error:

![image](https://user-images.githubusercontent.com/3259326/190864969-ba8eeeb5-878a-43a3-968a-c3e6e6bfa82d.png)

Relevant bits about the example:

* I usually keep notebooks in subdirectories of the project, tipically under `R/`.
* I like to set `knitr::opts_knit$set(root.dir = here::here())` at the setup chunk, so paths in chunks are always relative to the project while I'm working.
* I like to save knitted documents to `output/renders` (again relative to the project's root) by setting `knit` in the YAML header to the following function: `rmarkdown::render(inputFile, encoding = encoding, output_dir = "output/renders")`.

If i omit the `include_graphics` call, Knit works and R is able to find the file with `file.exists`:  [notebook1.pdf](https://github.com/yihui/knitr/files/9592072/notebook1.pdf)

If instead of normalizing the path I pass a path relative to the project's root, using `knitr::include_graphics("results/image_1.png")`. I get a different error:

```
! Unable to load picture or PDF file 'results/image_1.png'.
<to be read again> 
                   }
l.174 \includegraphics{results/image_1.png}
```

I would appreciate if someone could point me in the right direction. What path should I pass to `include_graphics` in this case?

I'm submitting this as a bug report because I have been unable to find out why this happens, everything else seems to work, and it is unexpected behavior from my perspective.

Thank you for all the work you've done :)

Best!

N.

---

By filing an issue to this repo, I promise that

- [x] I have fully read the issue guide at https://yihui.org/issue/.
- [x] I have provided the necessary information about my issue.
    - If I'm asking a question, I have already asked it on Stack Overflow or RStudio Community, waited for at least 24 hours, and included a link to my question there.
    - If I'm filing a bug report, I have included a minimal, self-contained, and reproducible example, and have also included `xfun::session_info('knitr')`. I have upgraded all my packages to their latest versions (e.g., R, RStudio, and R packages), and also tried the development version: `remotes::install_github('yihui/knitr')`.
    - If I have posted the same issue elsewhere, I have also mentioned it in this issue.
[...]

PR #2459, opened 2026-09-03
fix: don't error in include_graphics() when only the converted path exists

yihui:
## Problem

The **msprog** vignette (a knitr reverse dependency) newly fails to build:

```
Error in `knitr::include_graphics()`:
! Cannot find the file(s): "unscheduled.png"
```

It built fine before, so this is a regression, not a package bug.

## Root cause

The vignette contains:

```r
knitr::include_graphics(paste0(getwd(), "unscheduled.png"))
```

Note the missing path separator: `paste0(getwd(), "unscheduled.png")` yields an absolute path like `/path/to/vignettesunscheduled.png` that does **not** exist literally.

`#2171` (commit a695994) changed the existence check in `include_graphics()` to test the **original** path (`path0`) instead of the **converted** (relative) path, to avoid false negatives when an absolute path is made relative to an output dir that differs from knitr's working directory.

But `xfun::relative_path("/path/to/vignettesunscheduled.png", "/path/to/vignettes")` strips the common prefix and returns `"unscheduled.png"` — which **does** exist next to the vignette. So:

- **Before #2171:** existence check ran on the converted path (`unscheduled.png`) → found → no error. (Worked, somewhat by luck.)
- **After #2171:** existence check ran on the original absolute path → not found → error.

## Fix

Consider a file found if it exists under **either** the original path (`path0`) **or** the converted path (`path`):

```r
if (error && length(p <- path[
  !xfun::is_web_path(path0) & !file.exists(path0) & !file.exists(path)
])) stop('Cannot find the file(s): ', quote_vec(p))
```

This keeps #2171's fix (original absolute path exists when `root.dir` differs) while restoring the pre-#2171 behavior when only the converted path exists. A genuinely missing file still errors.

## Tests

Added a regression test in `tests/testit/test-plot.R` for the "converted path exists" case. The existing #2171 test (original path exists, converted path resolves against a different base) still passes, as does the genuinely-missing-file case. Verified end to end by rendering an msprog-style chunk with `rmarkdown::render()`.


yihui:
Closing since https://github.com/noemimontobbio/msprog/pull/1 has been merged and the new version of msprog is on CRAN.

PR #2460, opened 2026-09-04
Cite import() base packages in DESCRIPTION as Imports

MichaelChirico:
https://github.com/yihui/knitr/blob/60e1c3b1bca5e1b57b7deb0f81071bb558d52596/NAMESPACE#L155-L158

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -119,9 +119,13 @@ Depends:
     R (>= 3.6.0)
 Imports:
     evaluate (>= 0.15),
+    grDevices,
+    graphics,
     highr (>= 0.11),
     methods,
+    stats,
     tools,
+    utils,
     xfun (>= 0.52),
     yaml (>= 2.1.19)
 Suggests:


yihui (commented):
Thanks! I'm curious why this is necessary.

MichaelChirico:
Depends what we mean by "necessary", there's no CRAN requirement for this now.

I wrote a little bit more here: https://github.com/yrosseel/lavaan/pull/634

yihui:
Okay, thanks for the explanation! Unless R may take any of these base packages out of base R (which I feel should be extremely unlikely), I won't be too worried about it, since these dependencies are guaranteed by base R.

MichaelChirico:
Yea, I don't think there's any urgency here. It does feel "correct" to me, anyway.

For example, here's cases where the analogous change makes even clearer sense: DESCRIPTION is more obviously wrong to declare {stats} as `Suggests` when it's in NAMESPACE `import*`s.

https://github.com/r-lib/rlang/pull/1918
https://github.com/tidymodels/rsample/pull/612
https://github.com/ikosmidis/enrichwith/pull/3

But, is it actually any more correct to have stats omitted entirely?

Anyway, until R itself enforces this, it's mostly theoretical.

yihui:
Having {stats} in `Suggests` is definitely wrong. For omitting it, I can't say if it's wrong or correct. Theoretically, yes, it's more correct to list these base packages in `Imports`.

MichaelChirico:
My read of Kurt's comment is this won't be pursued as something `R CMD check` enforces anytime soon:

https://bugs.r-project.org/show_bug.cgi?id=19160

yihui:
Got it. Thanks for the follow-up!

PR #2461, opened 2026-09-10
Convert test-table.R to a testit snapshot test

yihui:
## Summary

Rewrites the `kable()` tests (`tests/testit/test-table.R`) as a **testit snapshot test** (`tests/testit/test-table.md`).

Instead of asserting equality against long inline LaTeX/HTML strings and multiline character vectors, each `{r}` code block now prints the rendered table and the following block records the expected console output. This is easier to read and maintain than escaped string literals, and testit discovers `test-*.md` files automatically via `test_snap()`.

## Details

- `%==%` equality between two `kable()` calls → `identical(as.character(...), ...)` printing `[1] TRUE`.
- `grepl()` checks → recorded as printed `[1] TRUE`.
- `has_rownames()` group → four printed logicals.
- Edge-case empty-matrix loop kept as a side-effect block (empty output).

## Verification

Populated outputs with `update = TRUE`, then reran with `update = FALSE`: snapshot matches. All generated outputs are identical to the original expected literals.


cderv:
Is testit snapshot a new feature of testit ? Really cool !

yihui:
Yes, I added it in February this year: https://github.com/yihui/testit/releases/tag/v0.16 The `.md` snapshots are easy to read to both human and machine. I don't know why other people didn't use `.md` in other testing frameworks, as it appears to be the obvious choice to me :)

PR #2464, opened 2026-09-15
fix: apply the ditaa engine's default arguments

youdie006:
`R/engine.R:423` supplies ditaa's default arguments behind `if (cmd == 'ditta')`. `cmd` is `options$engine` and the registered engine is `ditaa`, so the branch can never be taken. Three lines below, the same `cmd` in the same list is spelled correctly.

The miss is silent: the `if` has no `else` and returns `NULL` invisibly, `list_fun()` wraps it as `function() NULL`, and `c(opts$args1(), ...)` at `:210` concatenates nothing. So every ditaa chunk has run with ditaa's stock settings since the arguments were added in #2092, and the comment on the line above has promised otherwise since 2022.

```
before   running: ditaa                   in.ditaa in.ditaa
after    running: ditaa -s 2 -T -S -E     in.ditaa in.ditaa
```

**This changes rendered output for existing ditaa users**: `-s 2` scales 2x, `-T` makes the background transparent, `-S` drops shadows, `-E` disables edge separation. That is your stated intent on the line above, but it is visible, so if you would rather keep the current effective behaviour and delete the dead line instead, say so and I will close this.

<details>
<summary>Verification</summary>

R 4.6.1, `testit`, full `tests/run-all.R` with `R CMD INSTALL` before every row.

| row | `R/engine.R` md5 | exit | result |
|---|---|---|---|
| master + the new test | `9ca8d174f2b0d41c17f4fa4fb677b6e9` | 1 | fails assertion 1: `grepl("^ditaa -s 2 -T -S -E ", ...)` is FALSE |
| this change | `4fb08ef30025b29a7d56f66cbbed431f` | **0** | whole suite OK |
| drop the condition entirely | `4261d31af7d2f06f48e48d120a182c70` | 1 | fails assertion 2, on **both** `dot` and `asy` |

The two mutants kill different assertions, so neither test case is redundant: the first proves the ditaa row, the second proves that `dot` and `asy` must not inherit the arguments.

The test captures the `running: ...` message `eng_exec` emits before `system2`, so **it needs no `ditaa`, `dot` or `asy` binary** and is safe on every CI leg.

I checked the other key lookups in `R/` for the same shape: `R/engine.R:427-428` covers all three registered engines, `R/block.R:531` is restricted by `match.arg`, and both `R/table.R` alignment lookups are guarded at `:220`. This is the only one with no producer.

No roxygen comment is touched, so no `man/` regeneration.

</details>

Disclosure: written with AI assistance (Claude). Every number above is measured.


yihui (approved):
Thank you very much!

PR #2470, opened 2026-09-17
Warn and drop out.width/out.height for pptx output (#2003)

### LINKED ISSUE #2003: `out.width` and `out.height` warning as ignored for pptx output format
Among docx, pptx, rtf, and odt, pptx is the only format that does not support width and height.

Do we need warning when output is pptx and out.width and out.height are specified?

_Originally posted by @atusy in https://github.com/yihui/knitr/issues/1746#issuecomment-523700310_

Following https://github.com/yihui/knitr/pull/1746 we never added the warning in knitr. Pandoc still does not support `width` and `height` for PPTX output (https://github.com/jgm/pandoc/issues/4586) 

It would be useful to add a warning about this when output format is detected to be pptx. 

cc @apreshill

yihui:
Closes #2003.

Among the Office output formats (`docx`, `pptx`, `rtf`, `odt`), `pptx` is the only one for which Pandoc does not support the image width/height attributes ([jgm/pandoc#4586](https://github.com/jgm/pandoc/issues/4586)). Previously, `out.width`/`out.height` were silently passed through to Pandoc and ignored for `pptx`.

Now, for `pptx` output, these options are dropped with a warning (consistent with how `fig.align`/`fig.alt` are already handled for Office formats). Other Office formats such as `docx` continue to honor `out.width`/`out.height`.

Added a regression test in `test-hooks-md.R`.


yihui:
Closing without merging. On reflection, adding a warning for `pptx` isn't the right move: the limitation is upstream in Pandoc (https://github.com/jgm/pandoc/issues/4586), and if/when Pandoc gains support for image width/height in `pptx`, we'd have to remember to remove the warning. Since we can't predict when that will happen, it's better not to add knitr-side code that we'd need to track and revert.

PR #2473, opened 2026-09-17
Use a plot object's own alt text (e.g. ggplot2::labs(alt=)) as the default fig.alt (#2001)

### LINKED ISSUE #2001: Allow plot methods to register custom alt text
It would be useful if plot methods could register default alt text to use when `fig.alt` is not supplied. I'm not sure how exactly this would work, but maybe knitr could provide some function that plot methods could call. Then (e.g.) `print.ggplot2` could do:

```R
print.ggplot2 <- function(x, ...) {
  if (isTRUE(getOption("knitr.in.progress")) && is_installed("knitr", 2.0.0)) {
    knitr::set_default_alt(x$labels$title)
  }
  ...
}
```

This would be a default, so would be overridden by explicitly supplied `fig.alt`. (And ggplot2 would also gain some richer ways of setting the alt text, probably via `labs()`).
 

---

yihui:
Fixes #2001.

## Problem

Plot objects can carry their own alt text (e.g. `ggplot2::labs(alt = ...)`), but knitr had no way to use it as the default `fig.alt`. The only workaround was per-chunk boilerplate with `eval.after` + `!expr`:

````
```{r}
#| fig.alt: !expr last_alt
p <- ggplot(...) + labs(alt = "...")
last_alt <- ggplot2::get_alt_text(p)
p
```
````

## This PR

When `fig.alt` is not set, knitr reads the alt text stored in the plot object and uses it as the default. So this "just works":

````
```{r}
ggplot(mtcars, aes(wt, mpg)) + geom_point() + labs(alt = "Scatterplot of mpg vs weight")
```
````
→ `Scatterplot of mpg vs weight`

## Design (revised — no public generic)

An earlier revision of this PR introduced an exported S3 generic `fig_alt()` so any package could register a method. That was more machinery than the problem needs: **ggplot2** is the only plotting package today with a way to store alt text in the object, and knitr already recognizes special plot classes (`htmlwidget`, `shiny.tag`) by class in the same `evaluate` value handler. So this now:

- recognizes alt-carrying plots by class (`.alt.classes`, currently just `ggplot`) in the value handler, right next to the existing `htmlwidget`/`shiny.tag` checks;
- reads the text via `get_alt_text()` (fetched from the **ggplot2** namespace, available since ggplot2 3.4.0), collected in plot-printing order;
- merges into `fig.alt` after chunk evaluation (alongside the `eval.after` pass), so nothing mutates a consumed option or unlocks `opts_current` mid-evaluation (#1798).

No exported API, no `NAMESPACE`/man/`.onLoad` changes. A new plotting package can be supported later by extending `.alt.classes` and `plot_alt_text()`.

## Behavior

- **Multi-plot chunks:** `fig.alt` is already a per-plot recyclable vector, so `collected[i]` maps to plot `i`.
- **Precedence / mixing:** an explicit `fig.alt` wins element-wise; `NA` (or `""`) falls back to the plot's own alt text, so `fig.alt = c(NA, "custom")` works.
- **Alignment safety:** only recognized plot classes occupy a slot (ordinary printed values don't shift the mapping); a recognized plot with no alt text gets an `NA` placeholder.
- **Known limitation (unchanged):** base-graphics plots are side-effect recorded with no printed object, so in a chunk mixing base plots with ggplot objects the positional mapping could drift; homogeneous chunks (the common case) align exactly.

### CODE REVIEW

**Review by yihui — COMMENTED** [review]
This feels so complex that I tend to just let users use the workaround... How often do ggplot users set the `alt` text?

**Review by cderv — COMMENTED** [review]
Thanks. 

> How often do ggplot users set the alt text?

I will ask around for feedback on this part about usage of `labs(alt =)`

> This feels so complex that I tend to just let users use the workaround... 

The workaround implies a non-obvious behavior of some options being evaluated after cell code, allowing for reusing some values, and it also leverages `!expr` which is R only, and not really that advertise in Quarto side. 

So I think having this new method is useful - if a users sets `labs(alt=`)` and no `fig-alt`, then I think this is good to be able to use it as it exists. 

In addition, we could also try to come up with the same idea for Jupyter cells if we can so that it works the same in Quarto for **ggplot2** and **plotnine** if they ever support labs too. They don't yet https://plotnine.org/reference/labs.html

I understand your interrogation though as the current manual setting is efficient already now that `ggplot2::get_alt_text()` exists.

File: NEWS.md [hunk 1]
```diff
@@ -47,6 +47,7 @@
 
 - `kable()` gained support for the global option `knitr.kable.keep.whitespace`. By default, leading and trailing white spaces in table cells are trimmed, but setting `options(knitr.kable.keep.whitespace = TRUE)` preserves them, which is useful for aligning numbers with a monospace font in the Markdown source (thanks, @jmbarbone, #2066).
 
+- The alt text stored in a plot object is now used as the default value of the chunk option `fig.alt` when the option is not set explicitly. Currently this is supported for **ggplot2** plots, so alt text set via `ggplot2::labs(alt = ...)` is picked up automatically without having to also set `fig.alt` in the chunk. For chunks with multiple plots, `fig.alt` is matched to plots element-wise, and an explicitly supplied `fig.alt` still takes precedence (you can mix the two with `NA`, e.g., `fig.alt = c(NA, "custom alt text")`) (thanks, @hadley, #2001).
```
yihui, line 50:
```suggestion
- The alt text stored in a plot object is now used as the default value of the chunk option `fig.alt` when the option is not set explicitly. Currently this is supported for **ggplot2** plots, so alt text set via `ggplot2::labs(alt = ...)` is picked up automatically without having to also set `fig.alt` in the chunk. For chunks with multiple plots, `fig.alt` is matched to plots element-wise, and an explicitly supplied `fig.alt` still takes precedence (you can mix the two with `NA`, e.g., `fig.alt = c(NA, "custom alt text")`) (thanks, @hadley, #2001).

```

yihui:
I feel the initial S3 generic `fig_alt()` was overly complicated for this small task, so I narrowed it down to support ggplot2 only for now. If other plot package authors are interested in the similar support, I can reconsider a general approach (such as providing the S3 generic for them to register their methods or some other mechanism of registration).

I'll merge this relatively simple PR but am still open to alternative implementations.

PR #2475, opened 2026-09-17
read_chunk(): read chunks from a knitr document by label (#2041)

### LINKED ISSUE #2041: read_chunk() could support reading Rmd and use chunk label
I am opening this here to not forget. It is possibly opened elsewhere but I did not find it. 

This is something that could be useful and see asked already for example in the community 
https://community.rstudio.com/t/use-chunk-labels-from-rmarkdown-with-knitr-read-chunk/106243

`read_chunk()` could read an R file and parse special comment.  (https://bookdown.org/yihui/rmarkdown-cookbook/read-chunk.html)

It could be interesting to be able to read an Rmd file and do the same but with chunk label and chunk content. 

Not sure how difficult it would be. Just throwing the idea.

yihui:
Closes #2041.

## Motivation

`read_chunk()` currently only understands **R scripts**, where chunks are delimited by `## ---- label ----` markers. #2041 (from @cderv) asks for it to also read a **knitr source document** (e.g. `.Rmd`, `.Rnw`) and pick up its code chunks by chunk label, so they can be reused elsewhere — mirroring the community request it links to.

## What this does

When `read_chunk()` is given a document rather than an R script, it extracts the code chunks and registers them keyed by chunk label, so a later chunk referencing that label reuses the code (the same mechanism as chunks read from an R script — code is stored unevaluated).

```r
# ext-src.Rmd contains:
#   ```{r ext-chunk}
#   z <- 99
#   ```

knitr::read_chunk("ext-src.Rmd")
```

````
```{r ext-chunk}
```
value is `r z`   # -> value is 99
````

## Design

- **Detection** — the document type is resolved from the file extension via `detect_pattern()`, or from the content when reading from `lines=` without a path. If the input isn't a recognized knitr document, `read_chunk()` falls back to the existing R-script parsing. R-script behavior is unchanged, and is also always used when `labels=` is supplied.
- **Extraction** — a lightweight `chunks_from_doc()` reuses the existing `divide_chunks()` / pattern machinery to split chunks and reads the label from the chunk header (`xfun::csv_options()`), rather than going through the much heavier `parse_block()` (which merges options, tracks dependencies, errors on duplicate labels, and has side effects). Works across markdown-family and Sweave/LaTeX/HTML patterns.
- Unlabeled chunks get an automatic label (`unnamed-chunk-N`); text and inline code between chunks are ignored.

## Tests

Added to `tests/testit/test-parser.R`:
- reads chunks from an `.Rmd` (label + multi-line body, options stripped)
- reads chunks from an `.Rnw`
- unlabeled document chunk gets an automatic label
- a `.R` script (even with a fence-like string in a comment) is still parsed as a script, not misdetected as a document

Regression suites (`parser`, `patterns`, `block`, `output`, `chunk`, `utils`, `tangle`) all pass.

## Note

As @atusy noted in the thread, reusing chunks across documents has caveats (hidden dependencies on variables/packages; harder to trace which file uses a chunk). This PR only adds the capability that was requested; it doesn't encourage it over factoring shared code into an R script/function.


cderv:
Great ! 

So much new things to update in the R Markdown Cookbook 😅

yihui:
True :)

PR #2477, opened 2026-09-17
Pass engine-modified chunk options to the chunk hook (#2333)

### LINKED ISSUE #2333: output hooks and current chunk options
I don't know if this is expected behavior, or an issue to fix, but this would be some improvement. 

It seems `output` hook is using options sets by engine or function in the chunk (like `options$results = 'asis'`), but others like `chunk` hook will only use the current chunk option as the one set in the source code 

Example 

````markdown
```{r, echo=FALSE}
knitr::knit_engines$set(demo = function(options) {

    options[["results"]] <- "asis"

    knitr::engine_output(
      options,
      options$code,
      out = "Should be `asis` content."
    )
  })

local({
  ohook <- knitr::knit_hooks$get("chunk")
  knitr::knit_hooks$set(chunk = function(x, options) {
    if (options$results != "asis" && nzchar(x)) {
      x <- paste(c("::: {.wrapped}", x, ":::"), collapse = "\n")
    }
    ohook(x, options)
  })
  ohook2 <- knitr::knit_hooks$get("output")
  knitr::knit_hooks$set(output = function(x, options) {
    x <- ohook2(x, options)
    if (options$results != "asis" && nzchar(x)) {
      x <- paste(c("::: {.internal}", x, ":::"), collapse = "\n")
    }
    x
  })
})
```

```{r}
1 + 1
```

```{demo}
Do no matter
```

````

````r
> knitr::knit("test.Rmd"); xfun::file_string("test.md")


processing file: test.Rmd
                                                                                                                                                                              
output file: test.md

[1] "test.md"


::: {.wrapped}

```r
1 + 1
```

::: {.internal}

```
## [1] 2
```


:::
:::


::: {.wrapped}

```demo
Do no matter
```


Should be `asis` content.

:::
````

The `.wrapped` div is never removed, while the `.internal` div is removed if the `results == "asis"` as set by the engine in this example. 

This is related to 
* https://github.com/quarto-dev/quarto-cli/issues/9184

Currently in Quarto only setting `output: asis` explicitly on the cell will allow to remove the external wrapping of output. 

Quarto redifines the chunk hook for this wrapping and checks the `options$results`

I wonder if there should be a mechanism for **knitr** hooks to consider the options as redefined possibly by the cell behavior and not just the explicit definition on cell options.

yihui:
Fixes #2333.

A language engine can modify the chunk options before producing output — e.g. set `options$results = 'asis'` and then call `engine_output()`. Those modified options were passed to the `output` hook (through `sew()`), but the `chunk` hook still received the options as **declared in the chunk header**. The two hooks could therefore disagree, which is what @cderv reported: Quarto redefines the `chunk` hook to wrap output and keys the wrapping on `options$results`, so an engine that sets `results = 'asis'` internally still got its output wrapped, even though the `output` hook already treated it as `asis`.

## Reproducer (from the issue)

An engine sets `results = 'asis'`; a user `chunk` hook wraps non-asis output in `::: {.wrapped}` and an `output` hook wraps non-asis output in `::: {.internal}`. Before this change, the demo chunk's output is still wrapped in `.wrapped` (chunk hook sees the header default) while `.internal` is correctly omitted (output hook sees the engine's change). After the change, both hooks agree and neither wraps the asis output.

## Change

- `engine_output()` attaches the (possibly engine-modified) `options` to its returned value as a `chunk_opts` attribute.
- `block_exec()` recovers those options after calling the engine, before running the `after` hooks and the `chunk` hook, so every hook sees the same options.

The recovery is guarded (`is.list(attr(output, 'chunk_opts'))`), so engines that don't go through `engine_output()` (e.g. plot engines returning image paths) are unaffected.

## Tests

Added a regression test in `tests/testit/test-engine.R`: an engine sets `results = 'asis'`, and the test asserts the `chunk` hook observes `options$results == 'asis'`. Verified it fails without the fix and passes with it.


yihui:
For context: this is the knitr-side fix that @cderv referenced in quarto-dev/quarto-cli#12024. That PR currently works around the gap with a hack — the Quarto `output` hook stashes `asis_output = TRUE` into a side environment when `results == "asis"`, and the `chunk` hook reads it back to restore `options$results`. From that PR's description:

> to make the change in knitr solves the problem at quarto level, modified options needs to be saved for the remaining cell context handling (until chunk hook). (This could be something done too #2333)

With this change, knitr propagates the engine-modified options to the `chunk` hook itself, so Quarto could drop that workaround.

cderv:
Thank you ! 

I updated issue in quarto-cli so I can adapt. 

Glad to see all those issues closed in knitr - awesome work with 🤖 !! 👏

cderv:
I am currently checking, but this seems to be a cause of breakage in Quarto. 

Investigation will tell us how this can be fixed, but I need to assess the impact for a new knitr and old quarto. Hopefully I can patch, or we can find a way to handle backward compatibility.

I'll report back.

PR #2481, opened 2026-09-18
include_graphics(): honor dpi for non-HTML/LaTeX Pandoc output (#2385)

### LINKED ISSUE #2385: `dpi` in `include_graphics` is not used for docx output
https://github.com/yihui/knitr/blob/4510c09120c9f137859c9c56f497f9de6d92961f/R/plot.R#L502-L506

It seems we don't use `dpi` calculation for `width` for other than latex or html. 

This is used here to set the `out.width`
https://github.com/yihui/knitr/blob/4510c09120c9f137859c9c56f497f9de6d92961f/R/output.R#L610-L611

````markdown
---
title: Test
output: word_document
---

```{r}
knitr::include_graphics("pyramide_age_qdl_2.jpeg", dpi = 192)
```

````
With this file [pyramide_age_qdl_2.jpeg](https://github.com/user-attachments/assets/0e37fc41-aa1a-44cf-89ca-c4864d1cc313)

It gives this markdown with no width 
````
![](pyramide_age_qdl_2.jpeg)
````

setting `out.width=500` for example will lead to 
````
![](pyramide_age_qdl_2.jpeg){width=500}
````

which works in Pandoc thanks to [`link_attributes`](https://pandoc.org/MANUAL.html#extension-link_attributes) extension 

For office document we use `hook_plot_md_pandoc`
https://github.com/yihui/knitr/blob/4510c09120c9f137859c9c56f497f9de6d92961f/R/hooks-md.R#L16-L26
Which support width 
https://github.com/yihui/knitr/blob/4510c09120c9f137859c9c56f497f9de6d92961f/R/hooks-md.R#L118-L135

Maybe we should support this ?

### PR DESCRIPTION
**Pull request by yihui**
Fixes #2385.

`raster_dpi_width()` computes an image width from the file's pixel dimensions and the `dpi` argument of `include_graphics()`. It only did so for LaTeX (inches) and HTML (pixels) output — for other Pandoc formats such as Word (`docx`) and ODT it returned `NULL`, so `dpi` had no effect.

This emits a physical width in inches for all non-HTML formats (LaTeX and the rest), which Pandoc understands in its image attribute syntax `{width=...in}`. HTML output is unchanged (bare pixel value).

Example (`output: word_document`):

````
```{r}
knitr::include_graphics("img.jpeg", dpi = 192)
```
````

now produces `{width=...in}` instead of `` with no width.

### CODE REVIEW

File: tests/testit/test-plot.R [hunk 1]
```diff
@@ -213,6 +213,27 @@ assert('include_graphics() converts absolute paths relative to the output dir',
...
+    # LaTeX uses a physical width in inches
+    opts_knit$set(rmarkdown.pandoc.to = 'latex', out.format = 'latex')
+    (raster_dpi_width(f, 192) %==% '2in')
+    # other Pandoc formats (e.g. Word) also get inches, not NULL (#2385)
+    opts_knit$set(rmarkdown.pandoc.to = 'docx', out.format = 'markdown')
+    (raster_dpi_width(f, 192) %==% '2in')
+
+    unlink(f)
+  }
+})
```
yihui:
```suggestion
if (loadable('png')) assert('raster_dpi_width() computes a width for non-HTML/LaTeX output too (#2385)', {
  f = tempfile(fileext = '.png')
  grDevices::png(f, width = 384, height = 288); plot(1); dev.off()
  op = opts_knit$get(c('rmarkdown.pandoc.to', 'out.format'))
  on.exit(opts_knit$set(op), add = TRUE)

  # 384px at 96 dpi -> 96px in HTML (bare pixel value)
  opts_knit$set(rmarkdown.pandoc.to = 'html', out.format = 'markdown')
  (raster_dpi_width(f, 96) %==% 384)
  # LaTeX uses a physical width in inches
  opts_knit$set(rmarkdown.pandoc.to = 'latex', out.format = 'latex')
  (raster_dpi_width(f, 192) %==% '2in')
  # other Pandoc formats (e.g. Word) also get inches, not NULL (#2385)
  opts_knit$set(rmarkdown.pandoc.to = 'docx', out.format = 'markdown')
  (raster_dpi_width(f, 192) %==% '2in')

  unlink(f)
})
```

PR #2489, opened 2026-09-18
Let opts_hooks options take precedence over engine-modified options at the chunk hook (#2488)

### LINKED ISSUE #2488: eng_html_asset (css/js) and eng_sxss (sass/scss) still force results = 'asis' onto the chunk hook after #2477
Following up on #2333 and #2477, which I think introduced a side effect for a couple of other engines.

#2477 makes `engine_output()` attach the options it was called with as a `chunk_opts` attribute, so `block_exec()` can pass the engine's effective options on to the `chunk` hook instead of only the header-declared ones. For `highlight` the second commit in #2477 removed the internal `options$results <- 'asis'` since it was a no-op for the engine's own output and only mattered for the leak to the `chunk` hook.

It looks like the same thing is still happening for two other engines that call `engine_output()` right after forcing `results` to `'asis'` themselves:

https://github.com/yihui/knitr/blob/32b63a055472356df489cbc5b9a52b6353ea8347/R/engine.R#L558-L565

`eng_html_asset` backs the `css` and `js` engines, and does the same thing further down for `sass`/`scss`:

https://github.com/yihui/knitr/blob/32b63a055472356df489cbc5b9a52b6353ea8347/R/engine.R#L920-L924

Before #2477 this `results <- 'asis'` only affected the `output` hook (through `sew()`), same as before for `highlight`. Now it also reaches the `chunk` hook unconditionally, for every `css`/`js`/`sass`/`scss` chunk, no matter what an `opts_hooks` callback set `results` to earlier in the pipeline.

Minimal reproducer, pure knitr, no rmarkdown. `repro.Rmd`:

````markdown
---
title: repro
---

```{r setup, include=FALSE}
knitr::opts_hooks$set(results = function(options) {
  if (grepl("<1>", options$code[1], fixed = TRUE)) {
    options$results <- "hold"
  }
  options
})

local({
  ohook <- knitr::knit_hooks$get("chunk")
  knitr::knit_hooks$set(chunk = function(x, options) {
    if (!identical(options$results, "asis") && nzchar(x)) {
      x <- paste(c("::: {.wrapped}", x, ":::"), collapse = "\n")
    }
    ohook(x, options)
  })
})
```

```{css, echo=FALSE}
/* <1> */
body { color: red; }
```
````

Then `knitr::knit("repro.Rmd")`.

With knitr at 793a020c (the parent of #2477) the `css` chunk's output is wrapped in `.wrapped`, matching what the `opts_hooks` callback asked for. With current master the wrapping silently disappears, the `chunk` hook always sees `results == "asis"` for this chunk regardless of the `opts_hooks` override.

This is how I found it: it's the same mechanism Quarto's own `chunk` hook relies on (forcing `results` to something other than `asis` whenever it detects an inline annotation marker like the `<1>` above, so it can wrap the output in its own markup), and it silently stopped working for `css`/`scss`/`sass` chunks after this landed on the dev channel.

I don't know if extending the `eng_highlight` fix to `eng_html_asset` and `eng_sxss` the same way is the right call here, since for those engines `results <- 'asis'` still does mean something for the `output` hook path too, not just a vestigial no-op like for `highlight`. Is there meant to be a way for something set via `opts_hooks` to stay authoritative for the `chunk` hook even when an engine also sets `results` internally, or is the `chunk` hook expected to always defer to whatever the engine ends up using?

### LINKED ISSUE #14910: css-code-annotation smoke test breaks against knitr dev/master, not yet in a knitr release
The smoke test `tests/docs/smoke-all/2023/01/31/css-code-annotation.qmd` started failing on CI (`.code-annotation-container-grid` missing) on both `ubuntu-latest` and `windows-latest`, starting with the push run at commit 572b946c6 on 2026-09-18. The last known-good push run was at a79a39a9 on 2026-09-11.

For an annotated `{css}` chunk (annotation = trailing `/* <1> */` comment), the entire annotation markup is missing from the rendered HTML, not just a class. No `<dl>`, no `code-annotation-anchor`, no `code-annotation-target` span, plain Pandoc ids (`cb1`, `cb2`) instead of our annotation-aware ids (`annotated-cell-N`).

This isn't caused by a quarto-cli change. `tests/renv.lock` is byte-identical between the last-green and first-red runs. Our CI (`test-smokes.yml`) and `tests/configure-test-env.ps1`/`.sh` deliberately overwrite `knitr`/`rmarkdown` from `yihui.r-universe.dev`'s dev channel after `renv::restore()`, so CI's knitr version is unpinned and rolling rather than the one pinned in `renv.lock`. yihui/knitr@dd0f913f2 (merged 2026-09-17, not in a knitr release yet, current release is 1.52 from 2026-09-06) landed on that dev channel and got pulled in by the next CI run.

I confirmed this by a direct local A/B render, same quarto-cli checkout, only the installed knitr ref changed:

- yihui/knitr@793a020c1 (parent of dd0f913f2): annotation markup present, `.code-annotation-container-grid` present.
- yihui/knitr@dd0f913f2: annotation markup entirely absent.

I've opened yihui/knitr#2488 upstream with a minimal pure-knitr reproducer and the mechanism: `eng_html_asset` (backs the `css`/`js` engines) and `eng_sxss` (`sass`/`scss`) force `options$results <- 'asis'` internally before calling `engine_output()`, and since dd0f913f2 that now reaches the `chunk` hook unconditionally, overriding whatever an `opts_hooks` callback set earlier. That's exactly what our own `chunk` hook relies on:

https://github.com/quarto-dev/quarto-cli/blob/c912ed669e5486cc0999060cb5a875dfae5c926f/src/resources/rmd/hooks.R#L811-L822

`knitr_options_hook` forces `results = 'hold'` whenever it detects an inline annotation marker, specifically so our `chunk` hook (not the engine) decides how to wrap the output. For `css`/`scss`/`sass` chunks that override no longer takes effect once knitr picks up dd0f913f2.

Since the current knitr release doesn't have this yet, nobody on a released knitr/quarto-cli combination is affected today. This only shows up because our own CI intentionally runs against knitr's dev channel. We'll want to:

1. See how yihui/knitr#2488 gets resolved upstream, and confirm the fix (whenever it ships) actually restores our annotation grids before it reaches a knitr release.
2. Separately decide whether `hooks.R`'s override should be made more defensive against an engine overriding `results` after the `opts_hooks` callback runs, independent of the upstream fix.

I'll leave a note that #14909 was opened automatically against the CI failure and diagnoses an unrelated pre-existing issue in `format-html.ts`, it doesn't fix this.

### PR DESCRIPTION
**Pull request by yihui**
Closes #2488. Also unblocks quarto-dev/quarto-cli#14910.

## Problem

Follow-up to #2333/#2477. Since #2333, options that a language engine changes internally (e.g. the `css`/`js`/`sass`/`scss` engines set `results = 'asis'` so their raw HTML is written verbatim) are attached to the engine output and handed to the `chunk` hook, so it agrees with the `output` hook.

But that engine change reaches the `chunk` hook **unconditionally**, silently overriding any value an `opts_hooks` callback set for the same option earlier in the pipeline. @cderv found this breaks Quarto: `knitr_options_hook` forces `results = 'hold'` whenever it detects an inline annotation marker (e.g. a trailing `/* <1> */` in a `css` chunk), specifically so Quarto's own `chunk` hook decides how to wrap the output. Once knitr picked up #2333 on the dev channel, that override stopped taking effect for `css`/`scss`/`sass` chunks and the annotation markup silently disappeared.

## Approach

The first commit special-cased the four built-in engines. That only patched the built-ins and left the same hole open for any third-party engine, so this PR replaces it with a single, general rule (see the diff — the earlier commit's `engine_output()` `chunk_opts` argument and per-engine changes are reverted):

- In `block.R`, record which options an `opts_hooks` callback actually changed.
- After the engine runs, when recovering the engine-modified options for the `chunk` hook (#2333), re-apply the `opts_hooks`-changed options on top.

So an option set by `opts_hooks` is the user's final word and wins over an engine's later internal change, regardless of which engine is used. The #2333 behavior is otherwise unchanged: an engine's change still flows to the `chunk` hook for any option `opts_hooks` did not touch.

## Verification

- @cderv's reproducer: on master the `.wrapped` div silently disappears; with this change it is restored, and the `<style>` block is still written verbatim.
- New regression test in `tests/testit/test-engine.R` using a plain `demo` engine (not one of the four built-ins), asserting the `opts_hooks` value reaches the `chunk` hook even though the engine forces `results = 'asis'`. The existing #2333 test still passes.

No `NEWS.md` entry: the regression was introduced by #2333 on the current dev version and never shipped in a CRAN release.

### CODE REVIEW

**Review by cderv — COMMENTED** [review]
Thanks for picking this up so quickly, and for going further than I managed to. I had convinced myself this was mine to work around downstream, so I am glad I asked rather than patching around it.

I tested the branch against Quarto three ways: the reproducer from the report, a real document, and our own smoke test for this feature. That test fails against current master and passes against your branch, with no change on the Quarto side, so this fixes it for us.

I also checked that a user-defined engine setting `results = 'asis'` still propagates it to the `chunk` hook, since that was the case from #2333 I originally cared about, and it does.

I have a small self-contained document that pins a knitr ref and reports pass or fail for this, if that is useful to you for checking other engines. 

<details><summary>Details</summary>
<p>

````markdown
---
title: "knitr #2488 / #2489: code annotations in css chunks"
format: html
ir:
  packages:
    - yihui/knitr@fix-2488-engine-asis-chunk-hook
---

Render with `ir render knitr-2489-reprex.qmd` and read the page. Two things to look at: the annotated `css` chunk below either gets its annotation gutter and numbered list or renders as a plain code block, and the last section prints a verdict line for the same contract checked in process, without Quarto involved.

To see the failure, change the pinned ref in the frontmatter to `yihui/knitr@32b63a055472356df489cbc5b9a52b6353ea8347` (master) or `yihui/knitr@dd0f913f2a1ffcd2eee3df01f4c8e870a400392b` (the commit that introduced it). `yihui/knitr@793a020c1b50b9c49e00dd5f8b1eb76c914d4619` is the commit before it and also renders correctly. The ref has to be changed in the frontmatter: passing a different `--with yihui/knitr@...` on the command line conflicts with the pin rather than overriding it, and dependency resolution fails.

The `css` chunk and its annotation list are the same ones used by Quarto's own smoke test for this feature.

```{r}
#| echo: false
d <- utils::packageDescription("knitr")
cat("knitr", d$Version, "\n")
cat("RemoteSha:", d$RemoteSha %||% "not a git install", "\n")
```

## A plain css chunk, no annotation marker

```{css}
h1 {
  color: blue;
}
```

## A css chunk with an annotation marker

```{css}
h2 {
  color: red; /* <1> */
}
```
1. Ensure that the H2 elements are red

## In-process check of the chunk hook contract

This does not depend on Quarto at all. It reproduces what Quarto's option hook and chunk hook do: force `results` away from `asis` when the chunk source carries an annotation marker, then wrap anything that is not `asis`. The engine must not overwrite that decision.

```{r}
#| echo: false
out <- local({
  engines <- knitr::knit_engines$get()
  hooks <- knitr::knit_hooks$get()
  ohooks <- knitr::opts_hooks$get()
  on.exit(
    {
      knitr::knit_engines$restore(engines)
      knitr::knit_hooks$restore(hooks)
      knitr::opts_hooks$restore(ohooks)
    },
    add = TRUE
  )

  knitr::opts_hooks$set(results = function(options) {
    if (any(grepl("<1>", options$code, fixed = TRUE))) {
      options$results <- "hold"
    }
    options
  })
  knitr::knit_hooks$set(chunk = function(x, options) {
    if (!identical(options$results, "asis") && nzchar(x)) {
      x <- paste(c("::: {.wrapped}", x, ":::"), collapse = "\n")
    }
    x
  })

  knitr::knit(
    text = c(
      "```{css, echo=FALSE}",
      "/* <1> */",
      "body { color: red; }",
      "```"
    ),
    quiet = TRUE
  )
})

wrapped <- grepl("{.wrapped}", out, fixed = TRUE)
verbatim <- grepl('<style type="text/css">', out, fixed = TRUE)
cat("chunk hook saw the opts_hooks value:", wrapped, "\n")
cat("raw css still written verbatim:      ", verbatim, "\n")
cat("VERDICT:", if (wrapped && verbatim) "PASS" else "FAIL", "\n")
```

````

</p>
</details>

yihui:
Thanks for testing it three ways and for the reprex document — that's useful.

Heads-up: after you reviewed I pushed c1ff4337, which replaces the per-engine patch (660d2ec3, the commit your review points at) with a single general rule in `block.R`. Rather than special-casing `css`/`js`/`sass`/`scss`, it records which options an `opts_hooks` callback changed and re-applies them over the engine-recovered options before the `chunk` hook runs. So the same protection now covers any engine, including third-party ones, not just the four built-ins — and it directly answers your "is `opts_hooks` meant to stay authoritative" question with "yes".

I ran your in-process contract check against the new commit and it still reports `VERDICT: PASS` (chunk hook sees the `opts_hooks` value, raw CSS still written verbatim), and the #2333 case you originally cared about is unchanged. If you get a chance, a quick re-test of your Quarto smoke test against c1ff4337 rather than 660d2ec3 would be a good confirmation, since the mechanism changed.

cderv:
Re-tested against c1ff43374. The in-process contract check still comes back `VERDICT: PASS` (chunk hook sees the `opts_hooks` value, raw css written verbatim), and the Quarto smoke test now passes directly against this commit too, no workaround needed on our side. Ran the full annotation-related Quarto test suite as well (22 tests) and nothing else regressed. Looks like the general rule answers the case just as well as the per-engine patch did.

yihui:
Perfect. I'll merge then. Thanks for confirming!

PR #2490, opened 2026-09-18
purl(): let comment='' keep eval=FALSE code uncommented (#2425)

### LINKED ISSUE #2425: Please allow uncommented code output to R script with chunk options eval=FALSE purl=TRUE
It seems to be knitr::purl feature to comment out the code chunks with options `eval=FALSE, purl=TRUE` in the extracted R script, and I could not find a workaround (option `coment=""` doesn't help). Let me show some use cases where this feature is a show stopper:

- The source document explains how to "play with data" on a specific example to achieve the desired result, but the extracted script should be usable with any data set. The script may thus ask the user about input/output file names or parameters relevant to its purpose. However, evaluation of something like `readLines("stdin", n = 1)` hangs the knitting process, at least in Rstudio.
- The source document explains how to use the generated script, for example describing its command line options. Of course there are no CLI options when knitting so eval=FALSE makes perfect sense then, but the code should be working in the extracted script.

yihui:
Closes #2425.

When `purl()` tangles a chunk with `eval = FALSE`, its code is commented out in the R script (with `#`). This makes it impossible to produce a script in which such code is meant to be *run* — e.g. code that reads interactive input (which would hang during knitting, hence `eval = FALSE` in the source document) but should still execute when the script is run directly. `comment = ""` was tried by the reporter as a workaround but had no effect.

`tangle_mask()` now honors the chunk option `comment`: `comment = ""` (or `NA`) leaves the `eval = FALSE` code uncommented. The default behavior is unchanged (code is commented out with `#`).

Given a chunk `{r, eval=FALSE, comment=""}` containing `x <- 1`, `purl()` now emits the line uncommented (`x <- 1`); without `comment=""` it stays `# x <- 1` as before.

Added a test in `tests/testit/test-tangle.R`; full `testit` suite passes.


R/block.R
```diff
@@ -654,7 +654,7 @@ tangle_block = function(x) {
     eval_read_chunk(code)
   }
   code = parse_chunk(code)
-  code = tangle_mask(code, ev, x$params$error)
+  code = tangle_mask(code, ev, x$params$error, params$comment)
```
yihui:
```suggestion
  code = tangle_mask(code, ev, x$params$error, params$comment %n% '#')
```

R/block.R
```diff
@@ -686,8 +686,14 @@ eval_read_chunk = function(code) {
...
 
-tangle_mask = function(code, eval, error) {
-  if (isFALSE(eval)) code = comment_out(code, '#', newline = FALSE)
+tangle_mask = function(code, eval, error, comment = '##') {
+  # code from a chunk with eval=FALSE is commented out with '#' in the R script,
+  # unless the chunk option comment='' (or NA) is set, in which case the code is
+  # left uncommented so that it can still be run from the script (#2425)
+  if (isFALSE(eval)) {
+    prefix = if (is.null(comment) || is.na(comment) || !nzchar(comment)) comment else '#'
+    code = comment_out(code, prefix, newline = FALSE)
```
yihui:
```suggestion
    code = comment_out(code, comment, newline = FALSE)
```

PR #2498, opened 2026-09-18
include_graphics(): add copy argument to copy images into the figure dir (#1513)

### LINKED ISSUE #1513: Add 'copy' parameter to include_graphics
Is it possible to add a 'copy' parameter to `knitr::include_graphics()` such that the image file does gets copied to local figure directory with the proper numbering, e.g. `figure7.png`?

Some context: the magick package automatically embeds figures when they are printed. 

```r
#' @export
#' @method knit_print magick-image
#' @importFrom knitr knit_print
"knit_print.magick-image" <- function(x, ...){
  if(!length(x))
    return(invisible())
  ext <- ifelse(all(tolower(image_info(x)$format) == "gif"), "gif", "png")
  tmp <- tempfile(fileext = paste0(".", ext))
  image_write(x, path = tmp, format = ext)
  knitr::include_graphics(tmp)
}
```

The issue is that the temporary png or gif file gets stored in tempdir which only works if pandoc is set to embed images. It would be nice if there were a way to let `knitr::include_graphics` automatically copy the figure to it's local figure directory and insert it from there.

yihui:
Closes #1513.

Adds a `copy` argument to `include_graphics()`. When `copy = TRUE`, each image is copied into the current chunk's figure directory (named like the chunk's plots, e.g. `figure/label-1.png`) and the copy is included instead of the original.

This solves the case raised in the issue: a `magick`-image `knit_print` method (or any code) produces an image in `tempdir()`, which only works if pandoc embeds images. With `copy = TRUE` the file lands in the figure dir and survives to the rendering step of non-self-contained formats.

Default is off; the global default can be set via `options(knitr.graphics.copy = TRUE)`.


R/plot.R
```diff
@@ -459,9 +465,23 @@ par2 = function(x) {
...
+  # copy external images into the chunk's figure directory (named like the plots
+  # this chunk would generate) and include the copies instead; useful for images
+  # in a temporary directory (e.g. from magick) that would not survive to the
+  # rendering step of non-self-contained output formats (#1513)
+  if (copy && length(path)) {
+    dest = mapply(
+      function(p, i) fig_path(paste0('.', file_ext(p)), number = i),
+      path, seq_along(path), USE.NAMES = FALSE
+    )
+    dir.create(dirname(dest[1]), showWarnings = FALSE, recursive = TRUE)
```
yihui:
```suggestion
    xfun::dir_create(dirname(dest[1]))
```

PR #2499, opened 2026-09-19
Add pkg_bib_nocite() for a nocite package bibliography (#1733)

### LINKED ISSUE #1733: write_bib could capture all citations for a nocite bibliography
Per the documentation, `write_bib()` provides only the package citation, not all the authors' preferred citations. As discussed in [this thread](https://twitter.com/tjmahr/status/1095767445514739713), the issue is with auto-generating citekeys. This is an understandable problem when the objective is to have a `.bib` file that can be used for in-text citations. 

But for many use cases, R and R packages are not often being cited in the text of a document, rather only included in the References section. In such a use case, citekeys could be auto-generated arbitrarily, and then passed into a `nocite` block. Example syntax: 

```
---
nocite: |
  `r write_bib()`
...
```

where `write_bib()` could return a concatenated string of the citekeys, `"@citation1, @citation2, ..."`, in addition to generating the relevant `.bib` file. 

In such a situation, it doesn't matter if citekeys are not preserved when citations change, because the citekeys are updated immediately, and not included in the text. While this solution does not solve the entire problem, it does make the `write_bib()` function substantially more useful.

This could even be baked into an RMarkdown feature that could be enabled with a YAML flag: `citepackages: yes` could create the bibliography file, add the citekeys as nocite references, and, combined with any other included `.bib` files, generate a References list for the document. 

I've been toying with making this kind of function, so I might take a swing at modifying `write_bib()` to work this way. But just wanted to throw it out there if you could make this kind of change quick(er than I could). 

---

- [x] I have fully read the issue guide at https://yihui.name/issue/.
- [X] I have provided the necessary information about my issue.
- [X] I have learned the Github Markdown syntax, and formatted my issue correctly.

yihui:
Closes #1733

`write_bib()` generates BibTeX entries for R packages but does not surface their citation keys. To list packages in the References section via the YAML `nocite` field (without citing them inline), users had to track the auto-generated keys by hand.

`pkg_bib_nocite()` returns those keys formatted for `nocite`, e.g. `@R-knitr, @R-xfun`. It captures **all** entries a package provides — not only the auto-generated `R-<pkg>` `Manual` entry, but also any extra preferred citations shipped in the package's `CITATION` (which carry their own keys).

Typical use in the YAML metadata:

```yaml
---
bibliography: packages.bib
nocite: '`r knitr::pkg_bib_nocite(c("knitr", "rmarkdown"), file = "packages.bib")`'
---
```

When `file = ''` (the default), the entries are discarded and only the keys are returned; the console echo that `xfun::pkg_bib()` produces for `file = ''` is suppressed so nothing leaks into the rendered output.


yihui:
Closing — better solved without a new function. Pandoc's `nocite: '@*'` already means "include every entry from the bibliography in the References," so `write_bib()` writing a `.bib` plus `nocite: '@*'` covers the "cite all packages" case with no key bookkeeping. Details on #1733.

PR #2505, opened 2026-09-19
Rcpp engine: compile from files so a source can #include sibling headers (#2367)

### LINKED ISSUE #2367: Feature request: rcpp code chunk compilation of {.h, .cpp} pairs or .cpp files that include other user-defined header files
Currently, an Rcpp code chunk expects a single standalone `.cpp` file.

## Compilation of {.h, .cpp} pairs

If you have a habit of separating function declaration in a header file with
`.h` extension and function implementation in the eponymous `.cpp` file, then
including the `cpp` file or the `.h` file in a Rcpp code chunk will result in
failure to compile. This is because Rcpp code chunks make use of
`Rcpp::sourceCpp()` function, which only accepts a single `.cpp` file by design.

## Compilation of .cpp files that include other user-defined header files

If you have multiple files to include in a single Rcpp code chunk, then you
cannot include them directly in the Rcpp code chunk. This is because, behind the
scene, `Rcpp::sourceCpp()` creates a -- possibly temporary -- cache directory
into which it copies only the single input file provided in the `file` argument.

## Workarounds

I have found workarounds for the two above mentioned problems which I made
a repository of to experiment: <https://github.com/astamm/rcpp-code-cell-processing>.

In short:

- Workaround to 1st issue: You should convert every pair of file (`myfile.h`, `myfile.cpp`) 
into a single file (`myfile.hpp`) and then include the `.hpp` file in the Rcpp code chunk.
- Workaround to 2nd issue: You can manually copy all the files you want to include in the
Rcpp code chunk to the cache directory created by `Rcpp::sourceCpp()`. This can
be done by setting the `cacheDir` argument of the `Rcpp::sourceCpp()` function
to the desired cache directory via the `cacheDir` optional argument. Behind the
scene, `Rcpp::sourceCpp()` subsequently creates a sub-directory in that
directory whose name is platform-dependent and retrieved via
`Rcpp:::.sourceCppPlatformCacheDir()`. We can therefore decide beforehand of the
location of that cache directory, create it and copy the files we want to
include in the Rcpp code chunk to that directory.

Could it be a nice feature to have this directly handled by **knitr** upon knitting?

Thanks!

---

yihui:
Closes #2367.

An `Rcpp` chunk that reads code from files via the `file` option previously compiled the *concatenated* content of those files. A source file containing `#include "foo.h"` then failed to compile, because `Rcpp::sourceCpp()` builds in a temporary directory where the sibling header does not exist. This is also why the `file = c("foo.h", "foo.cpp")` suggestion did not work.

## Fix

When the `file` option is set, `eng_Rcpp()` now compiles the file(s) directly instead of the concatenated code:

- all listed files are copied into one build directory under their original basenames;
- the last `.cpp`/`.cc`/`.cxx` file is compiled with `sourceCpp(file = ...)`.

Because `sourceCpp()` adds the source file's directory to the include path, a source file can now `#include` its sibling headers. This covers both cases from the issue:

````md
```{Rcpp, file=c("foo.h", "foo.cpp")}
```
````

Inline Rcpp chunks (no `file` option) are unchanged — the hot path still calls `sourceCpp(code = ...)`.

## Tests

Added a test in `test-engine.R`. It first probes that a trivial inline `Rcpp` chunk compiles (Rcpp can load without a C++ toolchain, e.g. on some CRAN machines), and skips only if that probe fails. The actual assertion — a chunk with `file = c("inc.h", "src.cpp")` compiles and its exported function works — then runs unguarded, so a regression in the multi-file path surfaces as a failure rather than being masked as a skip.

### CODE REVIEW

File: tests/testit/test-engine.R [hunk 1]
```diff
@@ -40,6 +40,31 @@ local({
   })
 })
 
+# Rcpp can be loaded on a machine without a C++ compiler (it ships as a binary),
+# so first probe that a trivial chunk actually compiles; skip only if it doesn't
+# (no toolchain). The real assertion then runs unguarded, so a regression in the
+# multi-file path surfaces as a failure instead of being masked as a skip (#2367)
```
yihui:
```suggestion
# multi-file path surfaces as a failure instead of being masked as a skip
```

tests/testit/test-engine.R
```diff
@@ -40,6 +40,31 @@ local({
...
+  writeLines(c('#include <Rcpp.h>', 'int addone(int x);'), h <- file.path(d, 'inc.h'))
+  writeLines(c(
+    '#include "inc.h"', 'int addone(int x) { return x + 1; }',
+    '// [[Rcpp::export]]', 'int add_one(int y) { return addone(y); }'
+  ), cpp <- file.path(d, 'src.cpp'))
+  # a source file that #includes a sibling header must be compiled from files,
+  # not from the concatenated code (the header would be missing at build time)
+  assert('an Rcpp chunk compiles a source file with its sibling header (#2367)', {
+    knit(text = sprintf('```{Rcpp, file=c("%s", "%s")}\n```', h, cpp), quiet = TRUE)
+    (exists('add_one')); (add_one(41L) == 42L)
```
yihui:
```suggestion
    (add_one(41L) == 42L)
```

PR #2510, opened 2026-09-20
Add optional internal debug logging via debugme (#2203)

### LINKED ISSUE #2203: A logging solution for debugging `knitr` and dependent packages
Hi, @yihui. I've been working on solutions for Python in `knitr` with `reticulate`.  I'm also exploring and extending (for me) features of `rmarkdown` and `bookdown`, including some templating with `jinjar`. It's been all fun and games, but the interdependence and complexity of these separate packages is starting to slow down my adventures... 

Beyond that, I hit a nasty file location bug recently and I'm unable to use some standard R debugging tools like `browser()` or `recover()` to inspect the environment at the failure point because it happens many levels deep (an R function called from a Python hook, called by a `knitr` block, possibly in a child process). I thought of using some "cat/print" calls, but there's another problem as `stdout` and `stderr` are both used in my setup for one thing or another and I'm never sure if the message could be swallowed somehow by `knitr` itself...

---

**I'd like you to consider adding some logging facility to `knitr`.** It may be either a standalone package like [`logging`](https://github.com/WLOGSolutions/r-logging) or a streamlined version of it included in the source. I'm thinking of something that could:
- log to a file, configurable somehow
- include and format context information like function scope, working dir, etc.
- be called from dependent packages like `rmarkdown` and `reticulate`

For example, `knitr` could expose a _logger object_, like implemented in Python's and R's `logging`, that would be used from other interacting packages by calling `knitr:::logger$debug("sample message")`. The code base(s) could then have some basic messages like "entering R/Python block", "running after hooks", "reading results from cache", etc. predefined for localizing the stage of code execution, and when debugging something we would be able to add a couple of temporary extra logging messages without needing to care about the whole logging setup.

**I'm motivated to write a draft PR, but I wanna know your opinion first.**

---

yihui:
Closes #2203.

Wires up [debugme](https://cran.r-project.org/package=debugme) for debugging knitr and packages that build on it, following the design agreed in the issue thread.

**Strictly opt-in.** `.onLoad()` only activates it when the `DEBUGME` environment variable is set to include `knitr` and debugme is installed:

```r
if (Sys.getenv('DEBUGME') != '' && requireNamespace('debugme', quietly = TRUE))
  debugme::debugme()
```

So the feature has zero effect — and debugme is never even loaded — unless a developer explicitly turns it on (e.g. `Sys.setenv(DEBUGME = "knitr")` before `library(knitr)`). debugme is added to `Suggests`.

**Markers.** A few `!DEBUG` string-literal markers are seeded at natural points in the chunk pipeline (chunk execution entry, cache load, cache save). debugme rewrites these literals into log calls at load time; when logging is disabled they are inert no-op statements. More can be added over time.

Verified end-to-end: with `DEBUGME=knitr`, knitting a cached chunk logs `block_exec()` + `saving cache` on the first run and `loading cache` on a re-run; with `DEBUGME` unset, output is silent. Existing block/cache/engine tests pass.


yihui:
Closing without merging — see my summary on #2203. In short: debugme markers have to live in knitr's source to trace its internals, and anyone editing the source to add markers can add the two-line enabler at the same time, so there's no durable piece for knitr to ship. This PR stays visible as a copy-paste starting point.

PR #2518, opened 2026-09-20
Use yihui/actions for R package dependencies and checks

yihui:
Switch the `setup-r-dependencies` and `check-r-package` steps in R-CMD-check from r-lib/actions (pak-based) to yihui/actions (base-R installer, no pak).

The pak backend has been failing on the macOS runner. The base-R installer resolves and installs Suggests from r-universe/CRAN directly, which should fix it (same setup as xfun and litedown).

Also:
- drop redundant `extra-packages` — gridSVG/targets/JuliaCall/magick are Suggests and get installed automatically
- drop `rcmdcheck` — `check-r-package` runs `R CMD build`/`check` directly
- coverage now via `covr::to_cobertura` + `codecov-action`, matching xfun

This PR runs the swapped workflow against itself, so a green macOS job here confirms the fix before it lands on master.


remlapmot:
See <https://github.com/r-lib/pkgdepends/issues/485>

yihui:
@remlapmot Thanks a lot for the info!

